Issue #2, opened 2015-01-22
RSP Interpreter lle graphics issue(s) in SD Hiryuu No Ken Densetsu


LegendOfDragoon:
Here's a screenshot of the intro  with RSP interpreter, using HatCat's fork of Angrylion's plugin for graphics.
![sd hiryuu intro](https://cloud.githubusercontent.com/assets/10492645/5850658/8b0437bc-a1ae-11e4-89da-679432d8547d.png)

Here's what it looks like in a match, also using zilmar's RSP interpreter.
![sd hiryuu match](https://cloud.githubusercontent.com/assets/10492645/5850662/a0288dc8-a1ae-11e4-96cf-800358b998e2.png)

cxd4:
I have also file a RSP pj64 interpreter bug here:
https://dl.dropboxusercontent.com/u/16494013/n64/Quake%2064%20%28U%29.pj.zip

I will look at this issue and mine in the next few days to see if they might be coinciding, thanks.

Also feel free to edit your message to use GitHub's native image uploading feature so that they show directly in the post--do you mean that the second one "what it looks like in a match" is with PJ64's RSP, while the first one is with mine?

LegendOfDragoon:
Both previous screenshots were taken with zilmar's RSP interpreter. I had trouble with save states, so i didn't post a comparison picture. it turns out, zilmar's RSP was also responsible for the save state problem. I'll go ahead and edit 1st post to use native image uploading feature.
Here's a decent comparison picture. I didn't take the pictures at the exact same in-game time frame.
This is with zilmar's RSP
![sd hiryuu bad](https://cloud.githubusercontent.com/assets/10492645/5850632/121d7778-a1ae-11e4-8a0b-49b3d8f48ca2.png)

This is with HatCat's RSP
![sd hiryuu good](https://cloud.githubusercontent.com/assets/10492645/5850639/2681cba6-a1ae-11e4-98da-3267b32dc8f1.png)

cxd4:
Okay.
Now try with this commit:
https://github.com/cxd4/project64/commit/5b346077ef536a42f79ffb1b914f9dcf62f1bd6f

Your issue here looks pretty similar.
Tell me if my commit to RSP 1.7.0.9 has fixed it.

LegendOfDragoon:
Ya that fixed the gfx issues. Nice job!

cxd4:
The issue is fixed now in the master branch:
[...]

PR #6, opened 2015-01-22
Fix a couple games in RSP interpreter:  CFC2 sign-extends $vcr.

cxd4:
Actual hardware behavior on the RSP interprets the vector control register (specified by MIPS.R.rd in the instruction decode phase) such that it ends up treating the vector flags register RSP_Flags[0], RSP_Flags[1], or RSP_Flags[2] as a 16-bit input and *sign-extending* it, not *zero-extending* it the way that Project64 has been doing.  This little overlooked detail has been causing bugs in LLE gfx for at least _Quake 64_, if not also:  https://github.com/project64/project64/issues/2 .

Example:
```
ori     $at, $0, 0x8000
ctc2    $at, $vco # RSP_Flags[0] <-- 0x8000
cfc2    $at, $vco
```

In Project64, the final result of $at is 0x00008000.
On real h/w, the correct result is $at = (0x00008000 | ~0x0000FFFF), or 0xFFFF8000.

### DIFF
--- a/Source/RSP/Interpreter Ops.c
+++ b/Source/RSP/Interpreter Ops.c
@@ -501,12 +501,14 @@ void RSP_Cop2_MF (void) {
 }
 
 void RSP_Cop2_CF (void) {
-	switch ((RSPOpC.rd & 0x03)) {
-	case 0: RSP_GPR[RSPOpC.rt].W = RSP_Flags[0].UHW[0]; break;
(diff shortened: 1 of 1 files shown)


project64:
thanks for that, I added just the change above of the sign extended part.

cxd4:
Whichever works.

The switch() code was a hindrance to function size/portability maintenance reasons, but you did the necessary change.

PR #13, opened 2015-01-23
Logic and start emulation fix

death-droid:
Fix some behaviour with Project 64, first one is the setting of NTSC rather than making use of the actual system type.

Second is fixing the crash caused with the "Start Emulation" option that was cause by g_BaseSystem not being initialized

### DIFF
--- a/Source/Project64/N64 System/Mips/Eeprom.cpp
+++ b/Source/Project64/N64 System/Mips/Eeprom.cpp
@@ -63,7 +63,7 @@ void CEeprom::EepromCommand ( BYTE * Command) {
 		if (Command[1] != 8 && bHaveDebugger()) { g_Notify->DisplayError("What am I meant to do with this Eeprom Command"); }
 		ReadFrom(&Command[4],Command[3]);
 		break;
-	case 5:
+	case 5: //Write from Eeprom
 		if (Command[0] != 10 && bHaveDebugger()) { g_Notify->DisplayError("What am I meant to do with this Eeprom Command"); }
 		if (Command[1] != 1 && bHaveDebugger()) { g_Notify->DisplayError("What am I meant to do with this Eeprom Command"); }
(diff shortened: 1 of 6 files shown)


cxd4:
Looks simple enough to me I guess, would probably merge (then delete that extra spacing thrown at the end of one of the commit line changes :P).

Yeah it supports MPAL this time.  Incidentally while I was looking at Mupen64, I randomly observed that supports MPAL too ... it does (system == PAL) ? 50 : 60; just like death-droid did here.

No I don't know which games are MPAL.

project64:
I like the changes in general, small little changes I want to make to it tho .. so I might make a tweaked version of the pull request when I get some time

death-droid:
Out of learning interests, what changes would need to be done?

project64:
mostly I want g_Settings->LoadDword(Game_ScreenHertz) to return the correct value. 

so it will take in to account the system type if nothing it set in the rdb.

also:
g_BaseSystem = new CN64System(g_Plugins, false);

should delete g_BaseSystem if it is not null

death-droid:
Ahhh i see what you mean, as how I've done it ruins the ability to control it via RDB.


The line i put above it should delete g_BaseSystem if its not NULL?
CN64System::CloseSystem(); function checks if its not null, deletes it etc..

death-droid:
^Only way i could thing of handling that case, since we only find out game system type after the ROM has been loaded.

project64:
ok I think I have everything now in with the following commits:
https://github.com/project64/project64/commit/a9f8dd461f0d632f904a1dfe858f24f17f62e6ed
https://github.com/project64/project64/commit/ba6c04c6f072a12b96b3b8b2fe71d63ba3cbbbe8
https://github.com/project64/project64/commit/8c1852afa036a772ccbc9959871398203461b110

death-droid:
Yep you got it!

PR #16, opened 2015-01-23
Deprecate `Indx` functional LUT on loop count in RSP interpreter.

cxd4:
This old `Indx` array is no longer of any use whatsoever.
It is an impediment to accuracy, readability, and speed.

It was created when the RSP plugin was first being made in the late 90s.  Back then, there was a risk of prematurely overwriting the elements.  Take for example:
```asm
vand    $v0, $v1, $v0[3]
```
This would effectively do:
```c
for (i = 0; i < 8; i++)
    VR[0][i] = VR[1][i] & VR[0][3];
```
This loop had a fundamental problem:  On iteration i = 3, `VR[0][3] = VR[1][3] & VR[0][3];` prematurely updated what would be the source element for all future iterations.

To solve this problem, zilmar created the `Indx` array sometime before 2000.
He had since forgotten this was an element of preventing RSP bugs and not a piece of the algorithm of the RSP interpreter's opcodes.  (This extra array really has nothing to do with the hardware or any algorithms of any opcodes.)  It is a hack, and it has since already been superseded by zilmar's superior method of creating a new temporary vector:  `VECTOR result`, and then just before returning from each function, `RSP_Vect[sa] = result;`.

I have not yet done these readability/optimization changes to RSP recompiler.
If it is decided that this PR should be accepted, then I can work on analogizing these changes to the RSP recompiler as well.  It's trickier to do with the recompiler since it doesn't use the `VECTOR result` method yet, though for accuracy (and speed) I think it should in a future PR, with the eventual removal of the entire `Indx` array altogether from the DLL, unless it is decided to decline this PR.

Overall, the problem with the `Indx` array is that nobody can figure out the simplicity of the interpreter opcodes by looking at a function of a function, when it should really just be a function.  The loop to update the entire vector register should have one dimension of LUT, rather than a LUT going off of another LUT and scrambling the element definitions in pseudo-random order, just enough to evade the overwrite bug.  (It wasn't even readable enough for developer to remember what the Indx array was even existing for, causing him to think there was a bug with overwriting elements.)

### DIFF
--- a/Source/RSP/Interpreter Ops.c
+++ b/Source/RSP/Interpreter Ops.c
@@ -536,7 +536,7 @@ void RSP_Vector_VMULF (void) {
 	VECTOR result;
 
 	for (count = 0; count < 8; count ++ ) {
-		el = Indx[RSPOpC.rs].B[count];
+		el = count;
 		del = EleSpec[RSPOpC.rs].B[el];
 
 		if (RSP_Vect[RSPOpC.rd].UHW[el] != 0x8000 || RSP_Vect[RSPOpC.rt].UHW[del] != 0x8000) {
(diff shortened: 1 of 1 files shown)


project64:
was looking at this more, it probably can be removed by the interpter but it is also used in the recompiler so it can not be completely removed

cxd4:
I have not removed it from the recompiler yet.
If it was not used by the recompiler I would have attempted a second commit to delete `Indx` entirely.

I was planning to remove the recompiler's dependency of this tidbit in a future PR for some speed improvement.

project64:
ok updated code with this

PR #25, opened 2015-01-24
Change VI Refresh back to 1500

Fanatic-64:
Fixes speed on most games


cxd4:
I think it's safe to agree with this.

squall's usually had a knack for admittedly jumping the gun on a few things, and I think (can't perfectly remember) he himself mentioned changing his mind a few months later about thinking the default refresh rate should be 2200 anyway.

oddMLan:
Conker's Bad Fur Day definitely looks a lot less smooth (FPS-wise) with VI refresh rate of 1500. Which is a shame, because with a 2200 VI refresh rate it looks buttery-smooth. Other games as well get benefited from it as well. :/

So I don't know how to feel about switching it back to 1500. :/

Fanatic-64:
VI Refresh can be set on the RDB. So if a game benefits from a higher (or lower) value it can be set individually. 1500 works correctly for most games, so it should be set as default.

cxd4:
yeah I think Resident Evil and/or Conker's was the reason squall recommended 2200 as the new default

But, I never even pay this much attention to how the games play, or how things are timed exactly.  I just play them over to look at the LLE if nothing else.

oddMLan:
I've just seen this video, and it looks closer to the 1500 VI refresh rate.
https://www.youtube.com/watch?v=p5E2Q4ko62M

Whichever one is more accurate, I guess.

PR #37, opened 2015-01-29
Plugin code cleanup

death-droid:
Implement @Twostars cleanup of the plugin system, makes all plugin (RSP,AUDIO,GFX and CONTROLLER) derive functions from a singular class. Cleans up the code significantly and removes the need to have duplicate code in every seperate plugin file that does the same thing. Should also make it much easier to modify and add new specifications that apply to all plugins.

### DIFF
--- a/Source/Project64/N64 System/Mips/Audio.cpp
+++ b/Source/Project64/N64 System/Mips/Audio.cpp
@@ -78,9 +78,9 @@ void CAudio::LenChanged ( void )
 		m_Status = 0;
 	}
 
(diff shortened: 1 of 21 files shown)


project64:
ok created a branch that mast most of this functionality, have a look at it, if your happy with it I will merge that branch and close this pull request

death-droid:
Looks all good to me!

death-droid:
Just need to add plugin base.cpp and plugin base.h to Visual studio 2013 project file

project64:
Ok merged the code

PR #45, opened 2015-02-01
Fixed some problems with GetInstructionInfo

LegendOfDragoon:
I originally was going to submit my changes for MFC0, but I decided to keep it simple. I think what messed instruction reordering up, was the fact that when RSPOpC.rd == 4 or 7 for COP0_MF, it finishes sub block. I personally would rather not have it finish sub block. That can be changed some other time i guess. There are some other changes that need to be done in MFC0 as well.

Mia Hamm should work with Instruction Reordering enabled now.


project64:
looks fine, but only did a basic look over it, if it fixes a game great. I guess when we get to more testing if there are problems then we will see what is happening then.

PR #46, opened 2015-02-01
fixed almost all Glide64 warnings

cxd4:
All compiler warnings (tested in VS2013 this time) in the Project64 solution have now been fixed for Glide64 with 3 exceptions.

Two of them are in the GlideHQ module.  I am not qualified to fix them...they seem to be PNG decoder regressions with texturing and I am not really a texture enhancements person so leave this to anyone else to fix.  I was reluctant enough to fix all of these warnings I already have...I usually prefer to let the official maintainer of the code fix the warnings or just copy off their methods (since it's the same coding personality throughout the rest of the project anyway which is more consistent).  I don't really know anything about GlideHQ though, just basic RDP stuff.

The other remaining warning is in Glide64 ucode06.h.  It says `d` is uninitialized potentially, even though it clearly was being read and written before even that line of the code where the warning complained.  So if I have no idea why the warning is even there, it's safest not to fix it.

I was able to safely fix all Glitch64 warnings.

In the process of fixing the "potentially uninitialized" warnings, I saw that it was often because they were getting initialized within a `switch` statement, where the expression being `switch`'d on was not necessarily one encompassing all the possible case indices.  To fix this, I added a `default` case to guarantee initialization, initializing things like pointers, to NULL pointers, encouraging a crash in Project64 in case such buggy behavior within PJGlide64 should arise so that it can be debugged later.

### DIFF
--- a/Source/Glide64/Combine.cpp
+++ b/Source/Glide64/Combine.cpp
@@ -16007,12 +16007,15 @@ void InitCombine ()
 void ColorCombinerToExtension ()
 {
   wxUint32 ext_local, ext_local_a, ext_other, ext_other_a;
+
   switch (cmb.c_loc)
   {
   case GR_COMBINE_LOCAL_ITERATED:
     ext_local = GR_CMBX_ITRGB;
     ext_local_a = GR_CMBX_ITALPHA;
     break;
+  default:
+    FRDP("Invalid combiner locality %u.\n", cmb.c_loc);
   case GR_COMBINE_LOCAL_CONSTANT:
     ext_local = GR_CMBX_CONSTANT_COLOR;
     ext_local_a = GR_CMBX_CONSTANT_ALPHA;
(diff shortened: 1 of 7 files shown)


cxd4:
Please question and criticize any of the commits in this pull request that aren't certain before merging.  I tried to keep most of them 100% clear-cut readable, but not 100% of them are.  I'll answer any questions if anyone is confused on any commits.

I know that some of these warnings were already fixed earlier by zilmar, but [accidentally?] re-created in one of mudlord's latest commits:
https://github.com/project64/project64/commit/ba4610be618b5dc65f19b9ac8d80b062023eec16

That commit I guess was to staticize OpenGL extension loading without going through WGL/an extensions library and use fixed functions, but I don't know the inner details of Glide64's usage of OpenGL as an API, just how to use OpenGL, so I can trust that commit...just not the warnings introduced by it.

project64:
beyond getting glide to compile and use the setting system I have not really touched this code. Actually I really no nothing about the gfx side of things. So I am just assuming it is correct, nothing jumps out as completely wrong and as you have done more with gfx then me it should be ok

PR #53, opened 2015-02-04
removed invalid op-code from RSP commands stepper

cxd4:
I guess the RSP emulator treats LWC2:WV as an invalid op-code as it should, so the RSP debugger should probably do the same.  There is a SWV opcode:  It means "store transposed wrapped" and is based on STV.  There is even a LTWV opcode, just not on the RSP.  If it existed on the RSP it would have been named "LTWV", not "LWV", despite appearances.  Even so, the opcode is invalid and probably only existed in the commands stepper due to simple copy and paste, or it was just forgotten to remove it.

### DIFF
--- a/Source/RSP/RSP Command.c
+++ b/Source/RSP/RSP Command.c
@@ -1105,10 +1105,6 @@ char * RSPLc2Name ( DWORD OpCode, DWORD PC ) {
 		sprintf(CommandName,"LFV\t$v%d [%d], 0x%04X (%s)",command.rt, command.del, 
 			(command.voffset << 4), GPR_Name(command.base));
 		break;
-	case RSP_LSC2_WV:
-		sprintf(CommandName,"LWV\t$v%d [%d], 0x%04X (%s)",command.rt, command.del, 
-			(command.voffset << 4), GPR_Name(command.base));
-		break;
 	case RSP_LSC2_TV:
 		sprintf(CommandName,"LTV\t$v%d [%d], 0x%04X (%s)",command.rt, command.del, 
 			(command.voffset << 4), GPR_Name(command.base));


project64:
how do you know the opcode is invalid?

cxd4:
You ask how I **know** it is invalid?

You bring up a fine point.  Strictly speaking, I don't know it is invalid.

In fact the only blame I could think to see, is to not see the disrespectful appearance behind claiming information to be correct, when I have not had the physical resources to be able to reverse-engineer the RSP myself.  While I have studied some of others' reverse-engineering results, I never bothered to ask about LTWV, because I am just sure that it doesn't exist.  However, I'm not one who can prove it.

More importantly than whether I can be trusted on that, I would be happy if you could reject this pull request, so that I can move on to my next one, which is much more important.

project64:
you can have more then one pull request you know :P but Ok I will  close it.

it was in the original opcode matrix I had, so is it valid or not I am not sure. I only reversed ops that I had a hit against in used code. Does that mean it is not valid, or just not used, not sure.

I do not see any harm in leaving it in unless you got a more accurate opcode matrix or something else to say it is actually wrong.

oddMLan:
cxd4, I don't know if you already know this but I suggest you to push your commits in branches so then you can submit more than 1 PR

https://github.com/Kunena/Kunena-Forum/wiki/Create-a-new-branch-with-git-and-manage-branches

cxd4:
> it was in the original opcode matrix I had,

Which came from Project Unreality, which misinterpreted the documentation that was leaked to him.

> I do not see any harm in leaving it in

Me neither.  I don't really care.  This commit that I did was one of nitpicking.

>  unless you got a more accurate opcode matrix or something else to say it is actually wrong.

Giving you a more accurate opcode matrix does not mean you can see for yourself that said matrix is more accurate.  It's sort of like leaking documentation:  If you send it, the information is less likely to be reverse-engineered to validate the accuracy.  I don't think you would be interested in that kind of information since reverse-engineering is more reliable.

cxd4:
>cxd4, I don't know if you already knew this but I suggest you to push your commits in branches so then you can submit more than 1 PR at a time

Sorry, I tried to follow the steps in your link, but when I create a new branch and do a pull request to that it doesn't come out any better than trying to open multiple pull requests from my master branch.
https://github.com/cxd4/project64/tree/test

I can't seem to figure out doing multiple PRs.  Now if somebody who actually knows the process (or at least that it works) could explain it to me in detail (that page is partly missing a few steps I think?) that could help, but this is why I think there should be an IRC channel...too weird to talk about it here.

Issue #40, opened 2015-01-29
Divorce Glide64.ini from Project64.rdb


Fanatic-64:
Having Glide64 for PJ64 read (and write!) configuration settings from the RDB is a major hassle. First, because it means I have to maintain the settings for Project64, the Jabo plugins and Glide64. It is not possible to separately update the RDB and Glide64 settings. Plus, most people use the offical Glide64 Final anyway, since the PJ64 version introduced some bugs. And it will probably be obsoleted in a few months when GLideN64 is released, which gives me even less insentive to update it's settings.

Could we make Glide64 for PJ64 use a separate configuration file? Maybe just clone Glide64 and change the name only?

project64:
I would love to drop it for GLideN64.

The aim of the settings is to make every install of pj64 portable. now it would not be that hard to move them off to there own file.

but what in particular are you finding is major hassle with them both in the same file?

Fanatic-64:
The fact that if I'm working on improving game settings, I also have to work on improving Glide64 settings. If I ignore the Glide64 portion, it won't be possible to update the plugin's settings without replacing the RDB. It ties 2 different settings to a single file.

And also, the mashup of PJ64 and Glide64 settings looks rather cluttered. I makes it a bit hard to tell which parameters are for PJ64 and which belong to Glide64. If Glide64 is eventually going to be dropped, we may as well sanitize the RDB already to make it easier to work with. I'm willing to edit it myself to remove the Glide64 settings, you would just need to readd Glide64.ini from the official version (it's settings have not been changed so far).

project64:
what is wrong with replacing the RDB file?

all glide64 settings should start with glide64-

i do not see having an extra file will help make things easier. I know your editing the RDB file and there are a few other people who do, but in general it is not something that an end user will ever do, actually 99.9999% of people I assume never even open the rdb file.

cxd4:
I never open the RDB file either (in text form), but I'm inclined to agree with Fanatic-64.

Maybe splitting Glide64 settings out of Project64.rdb really does do nothing more than "having an extra file", but I still think it's a matter of logical concept.  If we're going to start merging Glide64's, other plugins' settings with Project64 RDB settings for fewer files, we may as well kill the plugin system and just statically link plugins within Project64 rather than use a plugin system to begin with.

[...]

PR #54, opened 2015-02-06
Split glide settings in to Glide.rdb

project64:
fix for issue #40


cxd4:
10/10 would merge again

Predator82Germany:
Then it works with both Glide64? or what is better with that?

V1del:
@Predator82Germany It means the Glide settings will be decoupled from the general PJ64 settings making it easier to make changes to default settings per game which concern Glide64 only

Issue #23, opened 2015-01-24
Killer Instinct LLE resolution is not correct while using LLE plugins


theboy181:
While testing Killer Instinct in Mupen 0.5 I noticed that the exact same LLE plugins I was using in PJ64 were outputting a lower resolution (320x240). Should be 640x480, and I believe its interlaced. This is odd that it happens in this game and think this issue might be larger than this particular game.

LegendOfDragoon:
Mupen64 0.5 forces expansion pack to be enabled. So keep that in mind.

theboy181:
This game came out before the expansion RAM did and has no impact on the output in PJ64

LegendOfDragoon:
I see. Had to make sure you knew though.

cxd4:
Yes that is an issue with Project64.

Project64 does not administrate the VI_V_CURRENT_LINE_REG or other video interface registers for handling interlacing of raster scanlines sent thru the VI.

Currently only Mupen64 0.5 seems to have that implemented.

cxd4:
Just noticed something about this issue report.
It says "while using LLE plugins".

This is only an issue with Project64, if you are using pixel-accurate plugins (either LLE or HLE, or zilmar's Basic CFB Plugin, either way really).  It's not a visible issue with Project64 if you're using a hardware-accelerated triangle-rendering graphics plugin, such as z64gl, Jabo's Direct3D or Glide64, no matter whether it is HLE or LLE.

So while it's Project64's fault for maintaining a couple VI registers, it's generally a moot point as most graphics plugins don't emulate the VI on a scanline-accurate basis.  So the issue could be put off for quite some time...I'm not even sure how to emulate that thing myself.  I have a clue about how to do it from my conversations with angrylion, but I would have to examine Mupen64 more to see the comparison.

theboy181:
Thanks for the update. Could this missing function be causing some the the Gfx issues that spear while using those other plugins and pj64?

LegendOfDragoon:
theboy181 only considers AL's plugin to be LLE lol ;/ .

I'd love to learn from Mupen64 0.5's source, but have no clue how to compile it with MSVC. I know there's a makefile for Mingw, but I like using MSVC for debugging. If someone could help me out with that, I'd be very happy. I may try learning from MAME and CEN64 as well, in order to fix these kind of issues.

cxd4:
>Thanks for the update. Could this missing function be causing some the the Gfx issues that spear while using those other plugins and pj64?

[...]

PR #55, opened 2015-02-06
basic timing support for VI_V_CURRENT_LINE_REG alternating scan fields

cxd4:
Currently Project64 does not really emulate VI_CURRENT_REG.  It has a field counter which is always AND'd by ~1, forcing the LSB to 0, but this breaks many games which do not necessarily run at a native FB resolution of 320x240.

For example, _Star Wars:  Rogue Squadron_, which has a native resolution of 640x480, not 320x240.  That game is currently unplayable, nothing but a black screen with no video output, when using a pixel-accurate graphics plugin with Project64.  By emulating alternating scan-fields when the VI is blitting the active video in interlaced mode (also known as the "serrated pulses" bit), we can draw every alternating scanline:  The first 240 even scanlines in the first call to UpdateScreen, then the alternating 240 odd scanlines in the second call to UpdateScreen.  This constructs the pixel-accurate 640x480 image, thus making Star Wars RS playable.

This affects many other games using high resolutions (not necessarily because of the expansion pak), such as _Killer Instinct Gold_ as reported in:  https://github.com/project64/project64/issues/23

And I know the commit most likely isn't done in the way that project64 would have liked.  I just didn't understand the source structure well enough, so make any additional stylistic commits necessary.

### DIFF
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
@@ -2625,6 +2625,7 @@ int CMipsMemoryVM::SW_NonMemory ( DWORD PAddr, DWORD Value ) {
 	return TRUE;
 }
 
+extern unsigned int field_serration;
 void CMipsMemoryVM::UpdateHalfLine (void)
 {
 	DWORD NextViTimer = g_SystemTimer->GetTimer(CSystemTimer::ViTimer);
@@ -2634,7 +2635,6 @@ void CMipsMemoryVM::UpdateHalfLine (void)
 		return;
 	}
 	m_HalfLine = (DWORD)(NextViTimer / g_System->ViRefreshRate());
-	m_HalfLine &= ~1;
 
 	int check_value = (int)(m_HalfLineCheck - NextViTimer);
 	if (check_value > 0 && check_value < 40)
@@ -2647,10 +2647,15 @@ void CMipsMemoryVM::UpdateHalfLine (void)
 		g_SystemTimer->UpdateTimers();
 		NextViTimer = g_SystemTimer->GetTimer(CSystemTimer::ViTimer);
 		m_HalfLine = (DWORD)(NextViTimer / g_System->ViRefreshRate());
-		m_HalfLine &= ~1;
 	}
 	m_HalfLineCheck = NextViTimer;
 
+/*
+ * 2015.02.04 (cxd4)
+ * Basic support for VI serrated pulses to DAC alternating scan fields.
+ */
+	m_HalfLine &= ~1;
+	m_HalfLine |= field_serration;
 }
 
 void CMipsMemoryVM::ProtectMemory( DWORD StartVaddr, DWORD EndVaddr ) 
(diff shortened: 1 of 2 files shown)


project64:
Ok I can have a look at cleaning up the code and merging it.

some things like field_serration, is a global which is wrong it needs to be a member of the class, but the general idea and code is good and I will add it.

LegendOfDragoon:
Just tested Rogue Squadron and the menu now displays. Nice job.

cxd4:
Likewise for Indiana Jones and other games which were at the wrong resolution earlier, which were either unplayable or had blocky graphics half the actual resolution and image quality.

project64, yeah I really don't know anything about classes, because I don't practice them, so I agree you may as well keep consistent to make it a member of the class.

cxd4:
Games off the top of my head I can think of that this will fix with angrylion plugin:

_AI Shogi 3_
_Donkey Kong 64_
_Indiana Jones and the Infernal Machine_ (big playability fix)
_Killer Instinct Gold_
_Mario no Photopie_
_Perfect Dark_
_Saikyou Habu Shoug_
_Star Wars:  Rogue Squadron_ (big playability fix)
_Stunt Racer 64_
_Super Smash Bros._ (credits only, uses 640x480 in staff roll)
_The Legend of Zelda:  Majora's Mask_ (Bombers' Notebook only)
_Top Gear Rally_
_Virtual Chess 64_
_World Driver Championship_

Probably other games I've not even played.  I don't really test that much.

project64:
> Though you say it _"needs"_ to be a member of the class?  Why is that...it works perfectly over here.
Well it does need to be for the sync core to be accurate, when it runs the interpter and recompiler in sync. You would change the value in the recompiler first that would change the value in the interpter before it even got there.

cxd4:
Easy enough for me to take your word for it.  I never do classes.  Have fun with whatever you need to do.

Now just a small warning:  You will notice some games have jittery/shaky scanlines when animation/motion is acting over the screen.  Like in Star Wars RS if you decided to test that, pretty soon you can see what I mean.  This is not so much a "fault" of the code I offer, so much as it is the natural invariability that the common plugin spec + any other N64 emulator's CPU-RSP timing is not harmoniously synchronous.

It can sometimes be adjusted with the Counter Factor option...1/3/5 may give better/worse VI scanline timing than 2/4/6.

LegendOfDragoon:
Ya I noticed the shaky scanlines. One quick fix is to simply use frameskip in your plugin anyway. Might as well since that game is 30 fps xD.

I noticed it didn't shake on an old version of M64p that had zilmar spec. Maybe it does have to do with counter factor ;/ .

project64:
https://github.com/project64/project64/commit/7345ecd97e131907d2d1ce6de9438dcd60793d6c

cxd4:
About perfect, provided the 90% of it that I was able to comprehend, and the pattern of the rest of the source.

cxd4:
>Ya I noticed the shaky scanlines. One quick fix is to simply set refresh to 1 in your plugin anyway. Might as well since that game is 30 fps xD.

Rogue Squadron runs at 60 FPS.
In fact any game this commit fixes, probably runs at 60 FPS, so that the interlacing can stage the field serration across 2 vertical interrupts.  It would not be 30 FPS but 60.

That's why setting skip refresh = 1 in my angrylion plugin "fixes" it:  It declines to draw every odd scanline, only the even ones, returning us to the blocky, low-res, incorrect graphical result issue we started out with in Project64 before this commit.

Issue #59, opened 2015-02-07
AI Shougi 3 "causing texture errors"


cxd4:
I guess this has been known about in the RDB file for quite a while.
![aishougi](https://cloud.githubusercontent.com/assets/1396303/6090631/1b2887e8-ae4e-11e4-972a-6baa148aca02.png)

But 1964 0.8.5, 0.9.9, 1.? support it just fine, so why can't we have a solution in Project64 sometime?
![aishougi_fixed](https://cloud.githubusercontent.com/assets/1396303/6090639/3daf0e5e-ae4e-11e4-99a0-df58c9d72288.png)

Also, unintended rhyming.  ("file..while"; "fine..sometime"...idk lmao I swear I didn't mean that)

exhalatio:
1964 is fixed from PI DMA odd length to an even length.
to checked dma.cpp Peripheral Interface DMA Write section.

cxd4:
Good eye.  So according to schibo it is a corner case with unaligned PI DMA write.

I will experiment with his experiment.  He seems uncertain about the change himself, but it's been done in 1964 since at least 1964 0.7.0 (AI Shougi 3 works fine there too).

LegendOfDragoon:
I think this is great, to implement good parts from other emulators :) . Thanks @exhalatio . How did you know, btw?

exhalatio:
My pleasure @LegendOfDragoon .
Someone says schibo is leave a comment in source code, so I can find it :) .

cxd4:
>I think this is great, to implement good parts from other emulators :) .

Well, whatever.  It's a mixed blessing I guess.  It would only be arrogance for me to try to isolate the issue on my own, as I would most likely not have guessed it was related to the PI registers until it was too late.

PR #64, opened 2015-02-08
better hardware support for odd PI_WR_LEN_REG addresses

cxd4:
I have no idea how to test something like this on actual hardware to validate it.  I may be able to send some particular information which may or may not help confirm the accuracy of these changes.

In the meantime, after making this pull request, AI Shougi 3 looks like this in a pixel-accurate render:
aishougiyay

Before, it was broken.  See issue:  https://github.com/project64/project64/issues/59.

The basic element to this fix goes something like:
```c
if ((g_Reg->PI_WR_LEN_REG + 1) & 0x00000001)
    g_Reg->PI_WR_LEN_REG += 1;
```
Except, the safest approach was to declare new variables to contain the actual length and manage things better that way.  While I was at it, I felt inclined to `&= 0x00FFFFFF` the RDRAM address, since it's well-established that physical RDRAM can never comprise more than 16 MB of memory, although there is a gate limit of `&= 0x3FFFFFFF` (1 GB of addressing space...or was it 2, forget).  I suppose it's questionable what to AND it by, or even whether to AND it at all like 1964 does, since I can prove nothing without doing hardware tests.  Still, only the low 24 bits are documented to have a purpose, and this decision can always be changed in a later commit now that PI_WR_LEN_REG has been centralized to a single variable storage for more maintainability rather than saying `g_Reg->PI_WR_LEN_REG` all the time.

### DIFF
--- a/Source/Project64/N64 System/Mips/Dma.cpp
+++ b/Source/Project64/N64 System/Mips/Dma.cpp
@@ -84,11 +84,13 @@ void CDMA::PI_DMA_READ (void) {
 
 void CDMA::PI_DMA_WRITE (void) {
 	unsigned long PI_WR_LEN_REG;
+	unsigned long PI_write_data_length;
 
 	PI_WR_LEN_REG = (g_Reg -> PI_WR_LEN_REG) & 0x00FFFFFFul;
+	PI_write_data_length = PI_WR_LEN_REG + 1;
 
 	g_Reg->PI_STATUS_REG |= PI_STATUS_DMA_BUSY;
-	if (g_Reg->PI_DRAM_ADDR_REG + PI_WR_LEN_REG + 1 > g_MMU->RdramSize())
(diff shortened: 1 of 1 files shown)


project64:
I am currently working on the pull request for changing plugins on the fly, once that is done I will look at your code for merging here and your other pull request.

cxd4:
Sure, I get that.  It did seem like death-droid's last PR was kept waiting for quite a while there, kind of naively felt sorry for him I guess.  Thanks for dropping by a note.

project64:
OK committed  (https://github.com/project64/project64/commit/17a3561c237b1f6a03d77b22a4e685ee979f328e). Does that work enough to close this ?

cxd4:
replied on the commit

But as long as you have an answer for my question to why that other change then I'm fine to just close this.

Issue #57, opened 2015-02-07
BattleTanx (U) SP DMA errors


cxd4:
After a lot of complicated troubleshooting with the Evangelion SP DMA errors, which happened with RSP audio ucode in LLE from various saved states on various emulators' interpreter cores, we ran into very consistent, very easy-to-find problem with _BattleTanx_:
1.  load ROM
2.  press START
3.  "New Game"
4.  If you get tired of waiting for the stupid cinematics you can just keep pressing A.
5.  SP DMA read error!  No saved state necessary.
6.  only happens with LLE gfx, audio could be HLE doesn't matter

The fascinating thing is that when the SP DMA errors start varies...sometimes it starts right after you press START button to get through the menu, sometimes it happens after you begin a new game.  Interpreters shouldn't be giving random/unpredictable results like this though I would have thought?  Then again I'm not really a R4300 person.

cxd4:
Depending on whether or not project64 is able to associate this with any core CPU/interpreter faults within pj64 itself, I will most likely end up self-assigning this as a RSP interpreter bug with unimplemented SP DMA.  I don't even have this implemented in my own RSP plugin actually....

LegendOfDragoon:
Honestly, I don't even press a button. I just disable frame limiter and the error happens early on in the intro. The error happens on 1964 1.1 as well.

project64:
>  Interpreters shouldn't be giving random/unpredictable results like this though I would have thought?

yes, they should be consistent if all the input is the same, if you eeprom is different, or you start pressing keys timing and code paths will be different. Also if fixed audio timing is not on then that can also cause inconstancy.

cxd4:
Okay in that case new steps to reproduce:
1.  load ROM
2.  just sit there without pressing START button until SP DMA :)

Seems to be consistent every time this time, with no key presses on my part, and I have just turned on Fixed Audio Timing anyway.

So I guess that means the ROM really is attempting to SP DMA an unmapped RCP memory address...it's a shame because if I "fix" this SP DMA error in pj64 RSP interpreter, there will be few error messages left to interrupt emulation in case the RSP ucode gets corrupted by the CPU recompiler or something (causing ear-damaging audio)...I guess invalid RSP opcode would be one of the only error messages left.  So in a way, I actually don't want to fix this issue in RSP interpreter, because having that SP DMA error message can save us from having ear wrecks (not completely though in my experience -.-).

Still got to do whatever's necessary to fix.

oddMLan:
[...]

Issue #36, opened 2015-01-28
SP DMA segmentation faults in Neon Genesis Evangelion


cxd4:
Reported:
http://forum.pj64-emu.com/showthread.php?t=4839

Saved state and EEPROM battery save file:
http://mega.co.nz/#!EolUEB7b!VyJOeMuUjKalTu6DP-4C_9ApTCBoZgQqhOJwJV18lac

Patiently doing nothing from the point at loading his saved state and waiting for me to pretty much die and finish getting beat, the SP DMA error messages do start to finally happen in Project64 RSP **recompiler**, but I have not managed to get the SP DMA crashes to happen when running the Project64 RSP in interpreter mode, just recompiler.  Also, simply clicking "OK" repeatedly in the SP DMA error messages still lets you continue the game normally...no corrupt audio or anything.

cxd4:
Update--only happens in HLE gfx + LLE audio??

Running the RSP in LLE gfx, there is no SP DMA crash (which I find odd).
Changing graphics plugin doesn't seem to matter (duh).

It's just HLE gfx && LLE audio, that the SP DMA errors 3 times, if you click OK all 3 times it's like nothing ever went wrong.  So this is a very minor bug/inconvenience in the RSP recompiler I think, as interpreter does not have it.

cxd4:
R4300 CPU interpreter does not affect the issue, just whether RSP is recompiled or not.

32-bit Engine RDB setting affects whether or not I can get the saved state to load in the first place, beyond that not sure.

LegendOfDragoon:
Well, judging by the testing I've done, I don't think it's an opcode issue. There are some strange things about the RSP recompiler that I recently noticed. For instance, for some games, it will recompile the same block multiple times. I will try investigating stuff like that in the near future.

cxd4:
It probably is a very subtle and indirect cause.

After all it only happens if (gfx == HLE && aud == LLE) ... any other combination of HLE/LLE for RSP graphics/audio seems to not have this recompiler bug.  And even then all you have to do is click OK on the 3 or 4 SP DMA errors and there is no corrupt audio or graphics.  So this is a very minor bug indeed, but with an interesting subtle cause it would seem.

LegendOfDragoon:
I forgot to mention that I get them regardless of whether HLE gfx or LLE gfx.

LegendOfDragoon:
[...]

PR #66, opened 2015-02-09
should read 0 from un-mapped RDRAM address

cxd4:
WOW MY FIRST SUCCESSFUL CONSECUTIVE PULL REQUEST (I hope...) after 10 failures!

All I really did was re-enable the `RSP_SAFE_DMA` macro to make it easier to make a change accounting for reading from an un-mapped RDRAM address.  This macro should either be re-disabled in a future commit once this stability patch is applied to the non-safe version as well, or this pull request should just be "manually" merged (re-coded) through some sort of cherry-picking or whatever like zilmar usually does.

We'll assume 8 MB of allocated RDRAM in this example:  Say we have the expansion pak installed.
The max DRAM address is 007FFFFF_16.
What happens if a SP DMA READ requests the contents of `rsp_info.RDRAM[0x00800000]`?

According to the documentation in the pixel-accurate base, as well as a couple RCP DMA tests I asked angrylion about, it should read 0 from un-mapped memory, and all writes should be null.  I implemented this to fix _BattleTanx_:  https://github.com/project64/project64/issues/57 and also I believe the Evangelion case of SP DMA read access violations:  https://github.com/project64/project64/issues/36.

Due to the nature of the changes it's not like I can expect this to get merged directly, but I'm hoping it at least helps @project64 decide what could be done.  Again to minimize complexity of changes I had to temporarily enable the `RSP_SAFE_DMA` macro in this PR.

### DIFF
--- a/Source/RSP/dma.c
+++ b/Source/RSP/dma.c
@@ -30,19 +30,17 @@
 #include "RSP Registers.h"
 #include "memory.h"
 
-// #define RSP_SAFE_DMA /* unoptimized dma transfers */
+/* to-do:  re-undefine this once non-safe made safe */
+#if !0
+#define RSP_SAFE_DMA /* unoptimized dma transfers */
+#endif
 
 void SP_DMA_READ (void) {
 	DWORD i, j, Length, Skip, Count, End, addr;
 	BYTE *Dest, *Source;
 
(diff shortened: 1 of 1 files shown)


cxd4:
Sorry need to close this.  My master branch is way too far out of sync with upstream master causing conflicts whenever I try to do certain pull requests.

oddMLan:
Is this implemented yet?

Issue #67, opened 2015-02-10
Configure ability regresion


theboy181:
ScreenHertz= XX doesn't work any longer in the RDB

I like to use this to OC the VI's for testing purposes. Was this change on purpose? Can we please add it back in?

Thanks

project64:
it should be better calculated now, tho the value in the rdb/cfg file should over ride the calculation, I need to look at why it no longer does.

cxd4:
http://forum.pj64-emu.com/showthread.php?t=4685

recent regression...?

death-droid:
Fixed, can be closed now

theboy181:
Great stuff!!!

PR #74, opened 2015-02-11
Fix problem where ScreenHertz was not setting based off of RDB.

death-droid:
Fixes issue #67


project64:
I had reset the default setting in a previous commit still to be 60 so your current changes will not work (ttps://github.com/project64/project64/commit/167ceb11b192e41ca3c3531af5b511f666ff211c) my fix worked if it was in the cfg file, not if it was just in the rdb file. I could have done another look up to check, but probably not worth it. Easier to just go with your solution.

Issue #61, opened 2015-02-07
Unhandled CicChip(-1) in First DMA 


theboy181:
Unhandled CicChip(-1) in First DMA error in pj64 when trying to load Aleck64 ROMS

Some ROMS will load but into test menu (Eleven Beat). Missing RDB entry for proper testing too.

death-droid:
This is due to Aleck64 roms having a different CiC that Project64 does not recognize. However from what I've read 6102 may allow them to work.

Might make sense for us to actually use a default Cic rather than causing it to set to -1 and cause the emulator to be useless even if the ROM may still run under another CIC

death-droid:
Aleck64 ROMS use a CiC chip that is not support by Project 64.  CIC-NUS-6102 may work. Otherwise we need to implement CIC-NUS-5101. I can assume once that's resolved there is probably a bunch other problems will be ran into.

This is probably more up Zilmar's alley. 
@project64

theboy181:
These ROMS are loading fine in Mupen64 so I am unsure if it supports the CIC's then.

project64:
http://wiki.arcadeotaku.com/w/Seta_Aleck_64

I guess it just needs some tweaks to get them to run? I have never tried any so I have no idea

theboy181:
I managed to get all but one of the games running in PJ64. 

Basically it's just missing from the RDB. Need to set memory to 8 instead if 4 and I am unsure if there are other settings that should be tweaked. 

You still get the unidentified CIC error. Not sure if you just want to skip that message or support the CIC. 

Is there a guide on what all the settings for a ROM do and how they should be set?

It would be really benificial if these games could be identified and governed outside of the RDB. This would make loading PD ROMS a little less time consuming.

theboy181:
If you want I can get you the patches for the ROMS so you can further test. PM me on the forums.

cxd4:
What's the only ROM left that RDB configuration doesn't get it to work?
Aleck64, was it?

Is that a public domain image I can find somewhere or something?  Does it work on other emu?

If it's difficult to troubleshoot openly I can just PM or talk about it over irc, but maybe irc would be quicker.

theboy181:
Tower&Shaft I cant seem to get it to load. Unsure what the correct settings are supposed to be.

Its a ALECK64 modified ROM and its running in Mupen64 iirc

Is there a function that adds the ROM settings to the RDB so that I can get the CRC's added to it for editing?

cxd4:
[...]

PR #79, opened 2015-02-13
Default to 6102 CIC for some weird but legit ROMs.

cxd4:
I make this PR for roughly 3 reasons:

1.  fixes part of the error nagging in issue https://github.com/project64/project64/issues/61
2.  Improve debugging.  Right now Project64 merely says something like "Unknown CIC (-1)".  I want debugging to show the actual checksum that Project64 came up with, since we already know it's going to be an unknown CIC anyway.
3.  It might get other ROMs to work off the bat without nagging end users as well, since other emulators (1964, Mupen64) appear to also default to 6102 for whatever reason (is this normal practice...?!), and we can always have the nagging back when testing in Debug build mode instead.

### DIFF
--- a/Source/Project64/N64 System/N64 Rom Class.cpp
+++ b/Source/Project64/N64 System/N64 Rom Class.cpp
@@ -244,9 +244,11 @@ void CN64Rom::CalculateCicChip ( void )
 	case 0x0000011A49F60E96: m_CicChip = CIC_NUS_6105; break;
 	case 0x000000D6D5BE5580: m_CicChip = CIC_NUS_6106; break;
 	default:
-		m_CicChip = CIC_UNKNOWN; break;
+#ifdef _DEBUG
+		g_Notify->DisplayError("Unknown CIC checksum:\n%016X.", CRC);
+#endif
+		m_CicChip = CIC_NUS_6102; break;
 	}
-
 }
 
 CICChip CN64Rom::CicChipID ( void ) {


cxd4:
Ah, I see that the "Rom Information" display needs to show if it's an unknown CIC as well.

My commit fixes an error but falsely causes the UI to label a ROM as 6102 when it's really unknown.  I could append a second commit to this PR to fix that but maybe better to wait for feedback first.

cxd4:
Another reason this PR shouldn't be merged -- Microsoft's custom LLP64 model can't treat `long` as a 64-bit type, so the `%016X` in my call to DisplayError really reads the CRC as a 32-bit integer and prints the low 32 bits of the CRC only.  You would have to either create 2 32-bit variables to print both of them or use some C99 extension.

Figured I should keep it open for review anyway.

project64:
yes not getting merged for all that you say.

it would be better to show it when debugger is enabled, not just in debug build. That way we can see it but most users will not.

you can do:
printf("%I64d", x);
https://msdn.microsoft.com/en-us/library/aa261215%28v=vs.60%29.aspx

to view a 64bit value.

cxd4:
I would never write the %l64d since that's a nonstandard compiler-specific extension, but as long as anyone besides me write that then that's fine to me.  Just that I don't want to be the one to write it.

And yeah, getting unknown to behave like 6102 sounds like a better idea.  Basic reason I attempted this PR was just to not nag people getting some special ROMs to work who didn't have the debugger enabled, but the way you proposed it seems better.

PR #80, opened 2015-02-13
preliminary centralization of MIPS size types

cxd4:
Technically things like "WORD", "DWORD" etc. can mean many different things, especially when emulating MIPS as the target CPU on x86 which has different sizes.  I thought it could be more readable if we began taking some small steps to express variable/function declarations with exact type sizes.

For example, in the future, we might not have to say this:
```c
BOOL CMipsMemoryVM::LW_VAddr ( DWORD VAddr, DWORD & Value )
```
We could in theory document the MIPS system better by using fixed-size types like:
```c
BOOL CMipsMemoryVM::LW_VAddr ( u32 VAddr, u32 & Value )
```
It would be less likely to conflict with a MIPS word type and a <windows.h> DWORD typedef.

For whatever reason, SGI, too, used this style of types in the Ultra64 OS, but they're not the only ones.

Last and probably least, it's part of my uber-secret conspiracy to gradually make Project64 more portable.
As far as 64-bit portability it probably doesn't help all that much, as long as you assume MSVC compiler.

It's somewhat difficult to introduce this change globally for Project64 and each component, so I wanted to start off by just throwing the typedef's in `Source/Project64/N64 System/Types.h`.  Maybe in the future there could be a header file for these types that the RSP plugin and things like that could co-share.

### DIFF
--- a/Source/Project64/N64 System/Types.h
+++ b/Source/Project64/N64 System/Types.h
@@ -10,6 +10,63 @@
 ****************************************************************************/
 #pragma once
 
+/*
+ * Some compilers have <stdint.h> types built in to them internally.
+ * In those cases, the "dependency" on <stdint.h> can be overlooked.
+ */
+#if defined(__STDC__) && (__STDC_VERSION__ >= 199901L)
+#define C99_COMPLETE
+#include <stdint.h>
+#endif
+
+/*
+ * standard integer types usable by the MIPS ISA
+ * also used by the Ultra64 OS
+ */
+#if defined(_MSC_VER)
+typedef __int8                  i8;
+typedef signed __int8           s8;
+typedef unsigned __int8         u8;
+
+typedef __int16                 i16;
+typedef __int32                 i32;
+typedef __int64                 i64;
+typedef unsigned __int16        u16;
+typedef unsigned __int32        u32;
+typedef unsigned __int64        u64;
+#elif defined(C99_COMPLETE) || 1
+typedef int8_t                  i8;
+typedef int8_t                  s8;
+typedef uint8_t                 u8;
+
+typedef int16_t                 i16;
+typedef int32_t                 i32;
+typedef int64_t                 i64;
+typedef uint16_t                u16;
+typedef uint32_t                u32;
+typedef uint64_t                u64;
+#elif !defined(_STDINT_H) || defined(__LP64__)
+typedef char                    i8;
+typedef signed char             s8;
+typedef unsigned char           u8;
+
+typedef short                   i16;
+typedef int                     i32;
+typedef long                    i64;
+typedef unsigned short          u16;
+typedef unsigned int            u32;
+typedef unsigned long           u64;
+#else
+#error Unable to emulate all of the types from MIPS ISA.
+#endif
+
+typedef i16                     s16;
+typedef i32                     s32;
+typedef i64                     s64;
+
+typedef float                   f32;
+typedef double                  f64;
+
 typedef unsigned char    BYTE;
 typedef unsigned short   WORD;
 typedef unsigned long    DWORD;


project64:
how about standard int types
http://en.cppreference.com/w/cpp/types/integer

int8_t
int16_t
int32_t
int64_t 	

uint8_t
uint16_t
uint32_t
uint64_t

LegendOfDragoon:
If you're going to change the types, I'd prefer u32 over uint32_t, simply because it's easier and quicker to type xD. Up to you though.

cxd4:
>how about standard int types
http://en.cppreference.com/w/cpp/types/integer

Those are okay.  It's actually what libretro fork of Mupen64Plus enforces, which is nonetheless an improvement over saying long/int or DWORD/WORD all the time.

The only thing I don't like about them is that it's a disagreement behind how much C language specifications should enforce (the more requirements/guarantees you make in the standard, the harder it becomes for compilers to exist that comply with it).  C89 leaves type centralization up to the programmer; C99 (1999 ratification) proposes the <stdint.h> solution you just gave.

I guess I'm fine with either one.  MSVC technically is a C89 compiler not C99, but it has enough C99 extensions to cover things like what you just said, `long long` things like that.

cxd4:
Oh, and since MSVC has the types built-in in ternally as things like __int16 etc., you might get better support to use those rather than the <stdint.h> types necessarily.  For example, including <stdint.h> on MSVC could give you a warning for the `long long` 64-bit type definition, whereas if you just used my typedef __int64 i64; solution for MSVC you would have no warning.

>I'd prefer u32 over uint32_t, simply because it's easier and quicker to type

That would be another reason for my preference, but I guess the most important thing is portability and avoidance of name collision potential.  Both u32 and uint32_t are about equally as unlkely to have problems with either of those issues, but yeah, I like to keep lines of code shorter.

project64:
I much prefer the stand int type for better compatibiltiy if we ever try to make it more portable.

we can look at _MSC_VER for older versions of visual studio if it does not support c99 in the stdint.h in which case we can define them our self as in
typedef __int32 int32_t;
typedef unsigned __int32 uint32_t;
typedef __int64 int64_t;
typedef unsigned __int64 uint64_t;

cxd4:
That works just if you're careful to prevent including `stdint.h` or any MS headers that include it; otherwise some typedef's can conflict.  Some compilers have those types like int32_t built-in internally, which is why it seems more compliant to use names that weren't already reserved or taken up by published APIs, such as `i32` instead of `int32_t`.  (Upon doing C99, I'm guessing they probably had to name it `int32_t` instead of `i32` because the latter was already frequently employed by programmers before any standard existed.  The "_t" at the end of each name is probably just to uglify each name on purpose to minimize the risk of name collision.)

Still, like I said, either solution is viable.  If I must choose between the _t name style and DWORD/WORD/Windows API types, I'm willing to do the former.

project64:
what do you want to do with this pull request?

cxd4:
Stop me if you're in the process of applying `stdint.h` already.  I guess I'll do it myself.

I was going to look into somehow "editing" this pull request's commits, but I don't have the experience for that.  I'll just propose a new one, without the "centralization" of MIPS types but just using the compiler resource header for the _t type names you prefer instead.  If you don't feel up to discussing why you prefer the names so much then there's no sense in leaving this pull request open if there's nothing to change your mind about.  I'll just take whatever improvement there is.

project64:
mostly if I am going to move away from windows format, which I am fine with cause it does confusing with mips where a WORD is 32bit and a DWORD is 64 bits, where in windows a DWORD is 32 bits. So I can see the need to change, it is more just if I am going to change it I would prefer to settle on a standard then a custom type.

cxd4:
I've viewed them as about the same--standard types are custom types customized by some standards committee or another, for their own reasons.  When N64 OS was being done the i16/i32/i64 types were the standard types for RCP, but some people are modernists and enforce appliance of the "cutting-edge" additions that grow the internal C specifications in size.

GitHub, itself, would seem to be a modern example of this.  I noticed that GitHub has special syntax highlighting for `int32_t` and the like, though, but not "custom" typedefs.  I don't have any of your experience in the real world, so I imagine everyone is just accustomed to it I guess.  So thinking back on it, it's worth the reasoning that they're more recognized to people in general, and at least with your method I only have to include the compiler global `<stdint.h>` rather than creating/modifying an existing local header file and including it everywhere throughout the source.

cxd4:
Minor update--I'm presuming you desire no change to usage of `float` or `double`, as `stdint.h` doesn't address those AFAIK.  So Project64 source will continue to refer to `float` as 32-bit single-precision and `double` as 64-bit double-precision.

project64:
is there any compatibility issues with the using of float or doubles?

cxd4:
none of which I've ever read about, just the usual tolerance by the specs that `float` is not necessarily 32 bits and `double` not necessarily 64 bits...

If this had been the case then I'm not even that sure that typedef's would make it a whole lot more portable anyway, since the only other type even possible is `long double`...Ultra64 OS types had `f32` and `f64` for float and double which was something I tried to emulate in this pull request before closing it, but if you don't care then I guess I don't really, since "float" and "double" actually seems kind of readable anyway to refer to MIPS single- and double-precision, even if it is slightly less portable.

cxd4:
@project64 one more question I have to know.

Line 627 of `RSP/Interpreter Ops.c`, in function `VMADN`:
```c
temp.UW = (DWORD)RSP_Vect[RSPOpC.rd].UHW[el] * (DWORD)(long)(RSP_Vect[RSPOpC.rt].HW[del]);
```

This is very confusing...VM?DM does (s32) * (u32) and VM?DN does (u32) * (s32), but here in VMADN you have written (DWORD) * (DWORD)(s32) which makes it sort of open to interpretation...I'm not that experienced with nested type conversions so would it be more correct to change it to this:
```c
temp.UW = (uint32_t)RSP_Vect[RSPOpC.rd].UHW[el] * (int32_t)(RSP_Vect[RSPOpC.rt].HW[del]);
```
... or were you intent on this ...
```c
temp.UW = (uint32_t)RSP_Vect[RSPOpC.rd].UHW[el] * (uint32_t)(RSP_Vect[RSPOpC.rt].HW[del]);
```

PR #89, opened 2015-02-16
Overhaul the way we are setting version info, instead of using the obscure r...

death-droid:
...esource way, make it defined in the code in a version.h file. Provides a more simplified and easier to change system in my own opinion.


project64:
no, the reason I have this is that I can look at the version number in windows explorer I do not have to run/load the file to get the version number

death-droid:
Hmmm ok I see what you mean there, never even thought about that. Just realised I can retain all functionality of yours but make the information easily accessible in a version.h file.

I'll submit a different pull request later tonight, but if you still don't like it, then so be it. Just dont like how seems a tad overcomplicated to grab the version of an app within itself

project64:
yes I read the version directly out of the resource section and use that .. in theory it could be some defines that are written to the resource section and then those strings can be just compiled in where it is read.

PR #90, opened 2015-02-16
Handle files that are dragged and dropped onto Projec64's window.

death-droid:
Allows the main window to be able to handle files that are dragged onto it, aka allow roms to be played that are dragged onto the window


project64:
it really should be:
char filename[MAX_PATH];

but it is fine I will look at fixing it.

PR #91, opened 2015-02-16
many many rdb changes

dsx-:
updated rom browser compatibility info for many games as well as plenty of rdb settings to improve compatibility


dsx-:
complete list of over 500 changes in this commit can be found here: http://www.dsxcorner.com/emulation/rdb_changes_full.txt

PR #92, opened 2015-02-16
Instantiate fixed-size standard int types to RSP module.

cxd4:
I left the vector unit and the COP2 vector op-code functions alone, since this pull request started to get more and more complex and questionable with the total number of the changes, so here is what has been standardized to fixed-size portable C types, in chronological order of significance:
* RSP word, double-word, and `VECTOR` register union member types
* RSP memory loads and stores
* a few remaining RSP scalar instructions (those inherent from the MIPS R4000 superset) which had type signedness mixing, or the unnecessary use of sign-extended data types in some of the shift or bit-wise op-codes which could further impede non-2's-complement CPU portability

This PR isn't meant to be perfect or 100% complete; at times I had to hold Ctrl+Z and undo some final changes to happen in the future maybe, since things were getting a little too sizable.  I didn't exactly Ctrl+F, find/search for "(WORD)" or "(DWORD)" throughout the RSP solution; I just read through all the interpreter, debugger and recompiler stuff for emulation-relevant usage of them.  Anything that was tied to the Windows operating system or low-level calls to the OS, I left alone to continue to use types like `DWORD`.  I mostly only touched the types for the actual emulation of the RSP CPU to better document them and make them cross-platform; any Windows API stuff like memory protection, debugger windows and recompilers I left alone to continue to use Windows types.

I hope that this helps demonstrate basic type of changes that could be done to the other components outside of the RSP plugin.  Or maybe I can try to mirror what I did here for the main Project64 CPU component if it's that acceptable.

### DIFF
--- a/Source/RSP/Interpreter Ops.c
+++ b/Source/RSP/Interpreter Ops.c
@@ -64,7 +64,7 @@ void RSP_Opcode_JAL ( void ) {
 void RSP_Opcode_BEQ ( void ) {
 	RSP_NextInstruction = DELAY_SLOT;
 	if (RSP_GPR[RSPOpC.rs].W == RSP_GPR[RSPOpC.rt].W) {
-		RSP_JumpTo = ( *PrgCount + ((short)RSPOpC.offset << 2) + 4 ) & 0xFFC;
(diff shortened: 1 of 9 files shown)


cxd4:
Never mind, new plan.

Before even this pull request there is something else which must be done to both main CPU and RSP.

PR #95, opened 2015-02-17
Version h add

death-droid:
Makes resource files get information from Version.h, makes the info easier to access from within the source code.


project64:
I wonder if t is possible to get copyright year to be done by compile time.

death-droid:
Hmmm I reckon it would be, just call like the date functions and use the year out of that?

project64:
except it would have to be done by the per-processor using __date__

PR #96, opened 2015-02-17
Disassemble memory offsets as signed integers.

cxd4:
The commands stepper and debugging units of the RSP and VR4300 components to Project64 were using `sprintf` formatting to get the memory load or store offset as a hexadecimal string.  This is rather unintuitive because the offset is always, in the MIPS architecture, a sign-extended immediate operand (as opposed to the "immediate" name given to its counterpart), but that is a pull request for later.

I can give plenty of official examples where programmers have written something more like this:
```asm
LW      $at, -8($t0)
```
It's just more intuitive and more used in practice by N64 developers.  We desire the 32-bit content of memory address *(T0 - 8), not this weird zero-extended hexadecimal that Project64 would always output:
```asm
LW      AT, 0xFFF8(T0)
```
First, FFF8_16 is not a valid offset.  Memory offsets are always in the range { -32768 <= x <= +32767 }.  The correct hexadecimal output would be -0x8(T0), however decimals are probably better in practice due to the scalar arithmetic nature of the offset which is either added or subtracted.  Hexadecimals are better for bit-sensitive operations, absolute addresses or unsigned memory things.

### DIFF
--- a/Source/Project64/N64 System/C Core/r4300i Commands.cpp
+++ b/Source/Project64/N64 System/C Core/r4300i Commands.cpp
@@ -628,82 +628,82 @@ char * R4300iOpcodeName ( DWORD OpCode, DWORD PC ) {
 		sprintf(CommandName,"daddiu\t%s, %s, 0x%X",CRegName::GPR[command.rt], CRegName::GPR[command.rs],command.immediate);
 		break;
 	case R4300i_LDL:
-		sprintf(CommandName,"ldl\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"ldl\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LDR:
-		sprintf(CommandName,"ldr\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"ldr\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LB:
-		sprintf(CommandName,"lb\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"lb\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LH:
-		sprintf(CommandName,"lh\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"lh\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LWL:
-		sprintf(CommandName,"lwl\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"lwl\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LW:
-		sprintf(CommandName,"lw\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"lw\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LBU:
-		sprintf(CommandName,"lbu\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"lbu\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LHU:
-		sprintf(CommandName,"lhu\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"lhu\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LWR:
-		sprintf(CommandName,"lwr\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
+		sprintf(CommandName,"lwr\t%s, %i(%s)",CRegName::GPR[command.rt], (short)command.offset, CRegName::GPR[command.base]);
 		break;
 	case R4300i_LWU:
-		sprintf(CommandName,"lwu\t%s, 0x%X (%s)",CRegName::GPR[command.rt], command.offset, CRegName::GPR[command.base]);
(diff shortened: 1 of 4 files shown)


project64:
the +8 or -8 is fine ...

but the +0x380 is a problem, having the value of 896 sucks
even with same with -0x24 makes it -36 

so showing negative is ok, but do not like the decimal

cxd4:
LW *(T0 - 0x24) bytes or *(T0 - 36) bytes, or *(T0 + 0x380 bytes) or *(T0 + 896) bytes...I don't really see a unambiguous reason why one is universally more clear to read than the other.

Maybe a MIPS programmer will have an array of 1000 bytes, and it does LW $at, 999($t0)...this is much more readable than using hexadecimal to express the 999th element.  Your +0x380 value just happens to apply to an aligned address which might or might not be readable depending on the cause/reason for referring to a particular data structure to load from (if it's something like 0x380 into N64 controller pak/mempak then that would be more readable; if it's the 896th element of an array like the 896th musical note collected in Banjo-Kazooie then that's mostly a matter of opinion).

cxd4:
There are reasons for both decimal and hexadecimal in the same way that the C language would allow you to have typed either one...but typing 0x380 instead of 896 would be pretty overzealous with hexadecimals when just trying to refer to the 896th object out of a 900-element array.

That's why I said %i in the sprintf formatter, even though %d (decimal) does the same thing.  Ultimately what matters is that it's a sign-extended integer (hopefully but not necessarily 100% of the time a decimal).  We can try to add a +/- and force hexadecimal instead of this pull request but that will require complications as I'm not sure whether a sprintf formatter exists to natively handle that.  If so feel free to add on to the PR--the point of this was to fix falsely treating the offset field as an unsigned 16-bit binary mask or something, when it's really a signed offset.

LegendOfDragoon:
>so showing negative is ok, but do not like the decimal

I agree with this statement. Not only is hex more convenient for knowing whether or not it's aligned, but it's easier for me to locate the memory address when I'm debugging. I understand there's pros and cons to both decimal and hex though. It's just that for what I generally do, hex is more preferable.

Since it's using sprintf, perhaps you could add an option to choose between hex and decimal.

cxd4:
>Not only is hex more convenient for knowing whether or not it's aligned,

And I disagree with this statement.
Whether or not the offset is aligned has almost nothing to do with whether the address is aligned.

The offset is an arithmetic value, not a binary mask or unsigned value.  Arithmetic has nothing to do with hexadecimal values or whether they're unaligned, and it's hardly worth throwing an extra "0x" at the beginning every time.

cxd4:
http://stackoverflow.com/questions/9602974/how-to-print-signed-hex-in-c

Do you guys want to invent your own sprintf function to defy the C standards, or are we just going to prefer inaccurately disassembling MIPS memory instructions on purpose?

LegendOfDragoon:
I know that offsets are only part of the equation. The point is that it's easier for me to work with hex, for certain things.

Ideally, it would be better to invent your own sprintf function, but I certainly am too lazy to do such a thing, at this point in time.

cxd4:
Feasible enough as it may be, if I were to invent my own sprintf function all I would do is just multiply both the complexity/inefficiency of the code as well as the liklihood that this pull request is defined.

And this isn't just with MIPS either.  C/++ compilers for x86 such as GCC or MSVC both emit LEA instructions with offsets as signed decimals.  They don't produce things like DWORD PTR [eax + -0xFFFF8000] or w/e.  It would compile to -32768.

Of all my foreseeable problems with this pull request I really did not foresee this, for some reason.  Fine I'll add another stupid commit for forcing it to base 16, if that's what it takes to fix this unsigned offset thing.

project64:
I can write my own sprintf .. .but I have no idea what you would want from output from it. Or you just want a function that does the conversions and out puts a string.

I much prefer hex and find it easier to work with, I would be continually converting it if it was in decima
l

cxd4:
If you're honestly willing to go that far, then I guess that at least speaks for how sure you are about this.

I still think that either way--whether it's a +/- decimal or +/- a hexadecimal, it's still a readable improvement over the current way that it's done before this pull request -- + an unsigned hexadecimal that we have to figure out the sign-extension on our own manually.  Just the fact that it was an improvement to the correctness of the opcode syntax should have already been organized enough for this pull request to have been done with already, with some future commit to make it a hexadecimal if it's that important--but to the point of the pull request which is fixing the instruction syntax, it definitely isn't.  If I was just making these commits because of stylistic preferences and not assembler accuracy, I would have made tons of other commits to the commands stepper besides these two.

>I can write my own sprintf .. .but I have no idea what you would want from output from it. Or you just want a function that does the conversions and out puts a string.

Probably best for consistency to just have a function that lets you do something like this:
```c
my_sprintf(CommandName,"LB\t%s, %X(%s)",GPR_Name(command.rt), (short)command.offset,
	GPR_Name(command.base));
```
... where in your my_sprintf, %X would prefix a '-' or '+' if the hexadecimal was negative or positive.

You could make your own function a lot simpler and less time-consuming to create if you didn't do it that way, I'm sure, but then you lose the benefit of documenting the entire instruction's syntax with a single printf-ish string, which is already what we're doing throughout the rest of the source code anyway so probably best to be consistent.

project64:
sprintf(CommandName,"ldl\t%s, %s0x%X (%s)",CRegName::GPR[command.rt], (short)command.offset < 0 ? "-1" : "", abs((short)command.offset), CRegName::GPR[command.base]);

cxd4:
Only disadvantage to readability is that it sprintf's 4 values (%s, %s, %X, %s), when LDL has only 3 operands to the instruction.

But I know it can't be helped you're just picking poisons.

cxd4:
k, appended a new commit ^^^ for hexadecimals to show instead of decimals

project64:
I try my best to eliminate all macros out of the code base, not always possible, but when possible I do. So I dislike the macro. 

Same with creation of  variables, this has to exist more but where possible I like to avoid them.

project64:
you can use a function like:

void format_i_instruction (char * buffer, const char * opcode_name, OPCODE &command)
{
	int abs_offset = (offset < 0) ? -offset : +offset;
	const char sign_offset = offset < 0 ? "-" : "";
 	sprintf(CommandName,"%s\t%s, %s0x%X(%s)",opcode_name,GPR_Name(command.rt), sign_offset, abs_offset, GPR_Name(command.base));
}

 case RSP_LB:
	format_i_instruction(CommandName,"LB",command);
	break;

cxd4:
I can't do this.  This code just keeps getting worse and worse the more I try and bend it.  If I keep going through this I'm going to get a headache.

cxd4:
First, that function doesn't work, because it's not written in C.  Possibly in pj64 module it may work.

Second, "format_i_instruction" is misleading.  I-type or immediate-encoded instructions is not the target of this fix.  Only memory loads/stores using the offset immediate need this.  You can't logically name it "format_i_instruction" because that would apply to opcodes like ANDI or XORI, which should have an unsigned hex printed instead like you're doing already, not an offset.

Third, even if the name wasn't misleading, you're taking one broken thing and fixing it in exchange for breaking another:  consistency.  I respected your original code because you used sprintf to document all of the op-codes without necessarily having to involve speed.  This time, you're actually willing to destroy the beautiful pattern you had of using sprintf in every switch case, and create a custom function with an unrelated prototype syntax just to settle a simple decimal/hex disagreement with C standard practice.  If you're going to create a new function, it had best resemble the consistent pattern of using sprintf syntax you have used throughout the REST of your code already.

Fourth, fixing issues and inaccuracies comes first, **before** personal coding preferences.  Replacing macros with something else, just so you can use some version of Microsoft's own incomplete "pre-pre-processor" debugger on a mere text-writer function, is more of a stylistic preference than anything.  By the time too many commits have been made trying to guess what each of your preferences are, the pull request becomes irreversible with wasted commits--can't travel back in time without redoing it completely.  It would have been far less painful for all parties involved to have just merged the working solution part of the pull request, then adjust it to your own stylistic preferences.

PR #99, opened 2015-02-18
Fix RomClosed() issue for RSP #1.1 plugins

LegendOfDragoon:
RSP #1.1 does not have RomOpened() so m_RomOpen is never true. Since
m_RomOpen is never true, RomClosed() never gets called.


project64:
look good, just can you move the traces inside the if statement

LegendOfDragoon:
Ok done.

PR #100, opened 2015-02-19
Fix mistakes in IsNextInstructionMmx() and UseRspFlags()

LegendOfDragoon:
Both functions could be improved for further optimizations, but I wanted to ensure accuracy.


project64:
good fixes

PR #102, opened 2015-02-19
Cleanout chms and htmllib

death-droid:


Source/Project64/Multilanguage.h
```diff
@@ -88,12 +88,10 @@ enum LanguageStringID{
 	MENU_LANGUAGE		=	175,
 
 	MENU_HELP			=	180,
-		MENU_USER_MAN	=	181,
-		MENU_GAME_FAQ	=	182,
-		MENU_ABOUT_INI	=	183,
-		MENU_ABOUT_PJ64	=	184,
-		MENU_FORUM   	=	185,
-		MENU_HOMEPAGE	=	186,
+		MENU_ABOUT_INI	=	181,
```
project64: it is fine, but you did not have to change the values for: MENU_ABOUT_INI, MENU_FORUM, MENU_ABOUT_PJ64, MENU_HOMEPAGE.

I assume you went through every language file to make sure you have renumbered the strings

death-droid:
haha yeah :P I had a moment of OCD with the numbers jumping from 180 to 183 and just had to fix it up. Though yeah, rather unnecessary.

I did go over through every language files. I noticed some are incomplete in the amount of text they cover though. But then theres alot of languages supported

project64:
I probably should look more at language and make sure everything is translated .. but that is probably after 2.2 is released. Now everything is on git hub, probably easier for people to update and maintain language files

PR #114, opened 2015-02-21
No point in having an older version of the controller specs. for Nrage

death-droid:


cxd4:
The only emulator that really conforms to Controller 1.1 specs is Mupen64.  Nemu/1964 don't even include it, and Project64 violates zilmar's own specs by customizing a change to the CONTROL_INFO struct.  So since we have no real compliance with the 1.1 version of the specs between the emulators, the only available cross-emulator option is version 1.0 of the spec.  So I see no reason to remove it, while having a plugin spec available to begin with.

death-droid:
Considering this version of N-Rage is rebranded as N-Rage for PJ64 I think retaining the cross-emulator compatibility is pointless. Plus if Zilmar goes ahead with making the translation system of N-Rage connect in with Project64's, that will break the cross-emulator compatibility.

cxd4:
Naming and labels don't make things pointless.  Projects often rename to not be confused with upstream.

Whether zilmar wants to eliminate all of the portability or keep some of it behind for other reasons isn't really "pointless" in either case; it's more a personal decision.  I'm sure that the differences between 1.0 and 1.1 are a "minimal code increase" enough, that it could be reverted to a 1.0 plugin rather easily if somebody did a fork of Project64 and/or pj64 N-Rage.

I guess it is curious that the 1.0 specs are kept inside of the N-Rage folder though, rather than just in the specs listings.  Still, I'm not sure what that potential redundancy really inhibits.

death-droid:
Hmmm i see your point :P I guess the way I'm trying to push Project64 is more in the route of Dolphin where input and everything is all built within the emulator. Mind you i think would need a way better way of handling it all then how nrage does so that we can plugin non windows specific API's suck as SDL and stuff so it wont limit the whole thing just to windows.

cxd4:
Hmm...yeah.  Starting to kind of get where you're coming from.

I don't know what Dolphin's exact reasons were for getting rid of plugins, just that it was related to their own individual problems they were having with their own spec, rather than just to avoid plugin specs in general like everyone thinks (when they analogize that Project64 should do the same from Dolphin's example).  There is a big portability problem though--the specs are written to use too many Windows types.

I was able to get my keyboard plugin working on Linux Mupen64 though, but for things like joystick, I would only be familiar with SDL, which is also not compatible with the WinAPI types in the plugin spec.  So I guess it's open to different approaches.

PR #122, opened 2015-02-22
NRage GBCart, use actual address ranges rather than hacky method. No rea...

death-droid:
...l benefit to the hacky method as the code isnt that much cleaner and creates edge cases.

### DIFF
--- a/Source/nragev20/GBCart.cpp
+++ b/Source/nragev20/GBCart.cpp
@@ -535,8 +535,10 @@ bool WriteCartNorm(LPGBCART Cart, WORD dwAddress, BYTE *Data)
 		return true;
 	}
 
-	if (Cart->RomData[0x149] == 1) { // Whoops... Only 1/4 of the RAM space is used.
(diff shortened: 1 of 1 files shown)


project64:
there are a lot of changes there, are they just all cosmetic code clean up, or is there any change of functionality in the code?

death-droid:
Change of functionality. original code used to grab highest three bits I think. This code instead makes sure the address ita using is with the range the cart actually uses.
Ive seen reports of people having problems with the original code, this should ensure it is at least working like the carts do

death-droid:
Checked against https://github.com/djhworld/gomeboycolor/blob/master/src/cartridge/MBC5.go and subsequent code files to ensure their the right address ranges as well

cxd4:
So it's GBC information on how to handle GB carts in the transfer pak...

Anyway to have this tested in practical use of the plugin to confirm that these changes improve/fix anything with transfer pak emulation?  Just to make firm sure about the code in use.

PR #123, opened 2015-02-22
Added more missing games on glide64.rdb

Lithium64:


cxd4:
Nice work Lithium64.  You've figured out branches even.

PR #124, opened 2015-02-22
Remaining missing games added to glide64.rdb

Lithium64:


cxd4:
Why do you close?  Why cancel your own pull request?  Was something the matter with the commit?

PR #140, opened 2015-02-25
We need to set the id for XInput controllers separate from the controlle...

death-droid:
...r number where handed.

This is to stop XInput controller one from having to be used in first player.

Also remove the need for the bconnected variable, allows controllers that are connected at game start to be disconnected and reconnected for XInput.

### DIFF
--- a/Source/nragev20/NRagePluginV2.cpp
+++ b/Source/nragev20/NRagePluginV2.cpp
@@ -343,6 +343,9 @@ EXPORT void CALL InitiateControllers (CONTROL_INFO * ControlInfo)
 		}
 	}
 
+	//To handle XInput controllers better, we need to set id to 0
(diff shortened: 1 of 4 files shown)


project64:
looks fine at a quick glance, tho I really should try and test it out.

death-droid:
I tested it with my wired 360 controller and Xbox One controller, worked fine.
The way the code originally worked was polling the XInput controllers based on the number handed to it. Problem with that is that it meant player one had to use the XInput controller, even if it was player two that wanted to use it. Otherwise you would need two XInput controllers connected to just allow the seccond player to actually use it.

bConnected was a pointless bit of code, pretty much removed the ability to hotplug controllers.

project64:
ok, I still have not tested it but it probably is fine

PR #156, opened 2015-02-27
Fix Rat Attack entry in RDB

Frank-74:


dsx-:
lol woops, why did i put 0, my bad

project64:
> lol woops, why did i put 0, my bad

probably cause it is 0 in the cfg file and No in the rdb

cxd4:
0's better

project64:
yes, which is why it is in the cfg file .. but I kept backwards support to how the rdb worked originally.

cxd4:
That then would be for the most part a good thing in that case.

PR #163, opened 2015-02-28
Make core mempak handle each mempak in a seperate file

death-droid:
Make core mempak handle each mempak in a seperate file, rather than bundling them all together in a singular one.


Frank-74:
Found a bug with this.

Using NRage with Raw Data off, It creates a *_Cont_1.mpk file of 32Kb.
Restart Project 64.
Configure controller, select Raw Data on and choose a mempak.  Start a game.
Restart Project 64.
Configure controller, select Raw Data off.  Start the same game, and *_Cont_1.mpk grows to 64Kb.

cxd4:
Yeah better to avoid `sizeof` ... funny I thought I was just saying that.

I don't know whether or not zilmar feels ready to completely re-do the mempak from 1 file per game to a potential 4 files per game (even though my instinct tells me I would feel fine with this change), so I think I'll just start out small and start a pull request to fix the current mempak format's size from 256 to 128 KiB.

Frank-74:
I think if its to be 4 mempak's per game, it should use the NRage format of mempaks.  So that mpk's created will work in NRage as well.
If a single 128Kb .mpk file, fix NRage to read these files and load the mempaks into each controller if used.

cxd4:
Oh, is there a conflict between N-Rage mempak format and Project64 32-KB mempak format (rip out the first 32 of the 128 KiB)?

That could be caused by some differing guesses/opinions on how the mempak file system starts out at page 0 of the controller pak file.  AFAIK N-Rage mostly borrowed from pj64 1.4 source, but maybe he tweaked a thing or two.  That shouldn't be hard to look into.

Having N-Rage understand a 128-KB mempak file as valid, on the other hand...well, could implement it to ignore the latter 96 KB and truncate it upon any future writes.  Then you would lose the data for controller 2, 3, 4 so maybe do a warning first.  But N-Rage has every right to not recognize a 128-KB file as a valid mempak, and I don't think that should change.

death-droid:
@cxd4 haha yeah, was planning on doing that down the track a bit, rather then lots of  changes in the single pull request. And yeah unsure if @project64 will be willing to change it to this way.

@Frank-74, thanks for testing it out. I have a feeling of what is causing it. Considering the file handles are not closed at the end of emulation. The 32kb version of the save files should be completely compatible with NRages.
The mempak code of NRage and Project 64 are very, very similar. Personally i prefer the code out of PJ64 but that's more just personal preference.

One of the few times I actually agree with cxd4, but yes N-Rage should not recognize a 128-kb file as a valid mempak, as quite simply it isn't.

Frank-74:
NRage doesn't like the 32Kb PJ64 mpk.
nrage-pjmpk

cxd4:
Oh wow that's totally broken.  It seems to be assuming all 16 note slots, even if null, are being taken up as used space, without propertly interpreting the note indexing data...or there is some ultra-simple singular bug that's throwing the whole thing off.

death-droid:
Hmmm that is extremely odd. First things first is working out why the size is being expanded open 32kb ( @cxd4 replaced all the sizeof's in my current code but will push that once i fix some of these bizarre problems)
Also Frank how did you produce that problem, cant seem to reproduce it on my side >.<

Frank-74:
I deleted all .mpk files from save folder, started Golden Eye, let PJ create the GOLDENEYE_Cont_1.mpk.  End Emulation.  Close Project64.  Restart Project64, configure NRage to look in Save folder, open GOLDENEYE_Cont_1.mpk.

So it should be a blank formatted .mpk file.  But the formatting is different.  NRage uses #ff as a blank filler byte, there are other differences too.  Look in PakIO.cpp, section void FormatMemPak( LPBYTE aMemPak )

death-droid:
Hmm thanks for testing this for me @Frank-74. I've solved the 64kb problem. Will look into the GOLDENEYE one next.

Frank-74:
Its also very weird that using NRage 2.3b with Goldeneye, sometimes gives you blue water in Frigate.

I've merged this and will test after I get some sleep.  Been awake for over 24hrs...

project64:
I need to look at the code more to make this better. 

My first preference is to remove all the code out of nrage for handling mem packs and force it to go to project64, basically not allow nrage to select raw option if it is set to mempak. 

I am happy to take any fixes from nrage and put them back in the core. Each controller should have there own mempak as that is the way it is on the n64 and it is strange if they are shared, that could lead to errors .. granted I have not played with multiple people on pj64 which is why I never noticed.

death-droid:
"My first preference is to remove all the code out of nrage for handling mem packs and force it to go to project64, basically not allow nrage to select raw option if it is set to mempak."
Agreed
To be honest, part of me believes the transfer pak should also be emulated by the emulator and not the controller plugin. So quite simply all the plugin needs to do is handle button presses and such.

At this stage the mempak changes of mine are pretty much exactly how Mupen64plus formats its. And fairly similiar to NRages, though N-Rage has a fairly odd way of doing mempak things.

project64:
I originally gave the plugin the option of handling this for two reasons, 
1 if you were using an adaptoid or something of that nature it could use a real mempack
2. if there was a bug in my code (which was closed source) it could be worked around.

being able to use real mempak files and be able to covert to them is probably the best reason to have 4 separate mempak files.

I think the value I initialize the mempak was from me dumping the contents of a real mempak .. but I could be wrong about that it was a long time ago.

cxd4:
That's what I thought.  I had read some documentation on controller pack/pak file format (often discussed as "controller RAM" in the references), and it looked like you had gotten the data from an exact mpk, probably using an Adaptoid (which I wanted only for the singular purpose of reverse-engineering the file system).  One of the documented fields is a serial ID--just a serial number for the exact accessory sold.  You had some arbitrary hexadecimal filled in there which caused me to think you did a dump somehow.

Whether mempack/rumble/transfer should be handled by the emulator or by the plugin is about the same debate to me as whether plugins should exist for the other types.  I don't really see how arguing whether a RSP plugin specs should exist or whether RSP should always be handled internally inside the emulator itself is any different than saying whether or not the controller plugin should be allowed to handle mempak instead of the emulator.

project64:
well controller plugins should be able to still do there own mem pak handling that is fine. Just now that nrage is going to be the default and distrusted from the source, there is no reason to have 2 different implementations.  This version is designed to be tighter coupled with pj64 not really as a plugin that other people will use in a different emu.

PR #165, opened 2015-02-28
ABL turned off in Pokemon Stadium games

Lithium64:
Temporary workaround until the ABL be fixed, Vigilante 8 games are already with ABL turned off.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -4888,12 +4888,8 @@ Clear Frame=0
 Culling=1
 Emulate Clear=1
 Primary Frame Buffer=0
-SMM-Cache=0
-SMM-FUNC=0
-SMM-PI DMA=0
-SMM-Protect=1
-SMM-TLB=0
 Self Texture=0
+Linking=Off
 
 [91C9E05D-AD3AAFB9-C:50]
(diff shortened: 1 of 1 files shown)


cxd4:
Why did all the self-modifying code config get deleted?  The recompiler needs to know those.

deleted:
```
-SMM-Cache=0
-SMM-FUNC=0
-SMM-PI DMA=0
-SMM-Protect=1
-SMM-TLB=0
`

Would it also be an inconvenience for you to please have one commit per RDB game fix?
Like,
commit 1:  Pokemon Stadium 1 - temporarily disable ABL
commit 2:  Pokemon Stadium 2 - temporarily disable ABL

Sure the games are both very similar but also as ROMs they are still very different and could do self-modifying code differently etc. things like that.

Lithium64:
@cxd4

I tested all these games and they worked well without Protect Memory

cxd4:
The name of the pull request was to disable ABL, so a separate commit should turn off protect memory if that was the case, if not a separate pull request.

Also, if protect memory is disabled, then some other compiler method must be enabled in its place apart from default RDB settings, which was what happened when you took out all those lines.

PR #166, opened 2015-02-28
Fixed PJ_Glide New MemPak/Browse button crash

Frank-74:
Seems to work fine now.

### DIFF
--- a/Source/nragev20/FileAccess.cpp
+++ b/Source/nragev20/FileAccess.cpp
@@ -1096,7 +1096,7 @@ bool SaveLastBrowseDir( TCHAR *pszFileName, DWORD dwType )
 bool BrowseFile( HWND hDlg, TCHAR *pszFileName, DWORD dwType, bool fSave )
 {
 	TCHAR pszFilter[DEFAULT_BUFFER];
-	TCHAR *pszTitle = NULL;
+	TCHAR pszTitle[64];
 	DWORD dwFlags = /*OFN_DONTADDTORECENT |*/ OFN_NOCHANGEDIR;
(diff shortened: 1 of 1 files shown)


project64:
Yes bug in that title should not be a pointer, good fix but it still needs some fixing, we are saying the buffer size is DEFAULT_BUFFER when we load the string, so either the size should be declared as DEFAULT_BUFFER or the usage of the buffer should be 64, or better yet bet sizeof(pszTitle)

cxd4:
I would do the macro.

There's nothing at all wrong about `sizeof(pszTitle)`, just that it can become tedious.  It would be too hard to debug if there was an issue with buffer r/w lengths, so if you changed all ~40 instances of `DEFAULT_BUFFER` in the source to instead say `sizeof(pszTitle)`, you would then have to change all ~40 of them back to use the macro again if you wanted to debug any input/output issues of some sort.

So macros are more maintainable and better for debugging issues.  If we insist against the macro due to Visual Studio's own debugger (which I guess doesn't simply complete all text pre-processing first before applying the debugger) then that is just multiple debug options competing against each other...though you could just use constants.  So instead of `#define DEFAULT_BUFFER 64`, you could evade the logical text pre-processer via `static const int DEFAULT_BUFFER = 64;` and make the compiler do all the analysis instead of the pre-processor.  (Which I guess could be a positive or negative thing.)

Also, the `sizeof()` method is, like a macro, English text and open to interpretation.  Readers could mistake it for the size in bytes, the size in arbitrarily-sized or unknown-size elements, or even assume that all memory is contiguous in which case `sizeof` causes a bug.

Frank-74:
Fixed with https://github.com/project64/project64/commit/4fed3ea0ec38924173964458180b3d4b445bc92d, so I can close this...

PR #174, opened 2015-03-01
Language Updates

MELERIX:
ANSI encoding and few fixes for Spanish translation.


Lithium64:
Look's good to me

I had forgotten to communicate zilmar that the translation doesn't work properly if it is encoded as UTF-8

PR #177, opened 2015-03-01
Optimize a few instructions in RSP Recompiler

LegendOfDragoon:
I just added a basic optimization for a few instructions where RSPOpC.rt == RSPOpC.rd. Along with implementing Cop2_CF and Cop2_CT, and an MMX version of VABS.

I carefully checked my code and tested a few games.

### DIFF
--- a/Source/RSP/Recompiler Analysis.c
+++ b/Source/RSP/Recompiler Analysis.c
@@ -88,6 +88,7 @@ BOOL IsNextInstructionMmx(DWORD PC) {
 			} else 
 				return TRUE;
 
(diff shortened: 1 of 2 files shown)


death-droid:
Except your implementations of Cop2_CF and Cop2_CT are currently disabled in your code?

LegendOfDragoon:
It's using the same macro as Compile_Cop2_MF and Compile_Cop2_MT, so it shouldn't be disabled.

death-droid:
Ohhh awesome, didn't realise they were pre-existing macros

PR #206, opened 2015-03-04
Prevent calling NULL/uninitialized OpenGL extensions in Glitch64.

cxd4:
Glitch64 seems to use various GL commands acquainted with extensions in the range of OpenGL 1.1 to 1.4, so I think at least a video card with driver support for OpenGL 1.4 might be required.  If this is not available, then it has to fall back to Microsoft's software implementation of OpenGL, which is very painful and supports only OpenGL 1.1.

To begin with, the problem being addressed here is just that functions which could not be initialized to an available driver implementation for some graphics commands, never remain uninitialized.  Instead of crashing Project64 and raising an exception, it just displays an error showing the function/extension your GPU and/or drivers lack support for.

### DIFF
--- a/Source/Glitch64/main.cpp
+++ b/Source/Glitch64/main.cpp
@@ -268,19 +268,19 @@ PFNGLBLENDFUNCSEPARATEEXTPROC glBlendFuncSeparateEXT;
 PFNGLMULTITEXCOORD2FARBPROC glMultiTexCoord2fARB;
 PFNGLFOGCOORDFPROC glFogCoordfEXT;
 void dummy_glActiveTexture(GLenum/*texture*/)
-{ /* GLX render opcode 197 */
+{ /* GLX render opcode 197, req. OpenGL 1.3 (1.2 w/ ARB_multitexture) */
     DisplayError("glActiveTexture");
 }
 void dummy_glMultiTexCoord2f(GLenum/*target*/, GLfloat/*s*/, GLfloat/*t*/)
-{ /* GLX render opcode 203 */
+{ /* GLX render opcode 203, req. OpenGL 1.3 (1.2 w/ ARB_multitexture) */
     DisplayError("glMultiTexCoord2f");
 }
 void dummy_glFogCoordf(GLfloat/*coord*/)
-{ /* GLX render opcode 4124 */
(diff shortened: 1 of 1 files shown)


cxd4:
Surprisingly, while the DLL file size did go up (obviously), it seems to still be less than the binary size it used to be before the previous pull request to stream the planar coordinates into vectors.  At any rate, the Windows Explorer display of "3.13 MB" has never changed.

Frank-74:
I've not noticed the size go up or down.  With your glitch vector commit, size remained the same for me, though you say it reduced.  Maybe you have optimizations selected for size?  I only optimize for speed, which usually increases the size.

cxd4:
Because it always said 3.13 MB every time I replaced the DLL in my user pj64 folder from the Git pj64 folder.

I only said size went down or up when Windows Explorer told me a file by the target name already existed and asked me whether or not I wanted to replace it, at which point, it will say "(larger)" next to the destination file, the file I am using to replace the destination file, or neither one.

project64:
do we really want to show message box, I assume you doing that to see what functions are being called and find for working out what is being called, but not sure it is what we want for the end users.

Maybe show them if debug setting are enabled.

cxd4:
The message boxes are called only in the `dummy_` versions of the functions.

If you have a OpenGL 1.4 or later video card, none of those functions ever get called.
The point is that if they do get called, it means your video card doesn't support those functions.

if (your video card supports glActiveTexture)
/**/CALL glActiveTexture(...);
else
/**/CALL dummy_glActiveTexture(...); # which displays an error MessageBox

The error is so the user knows which OpenGL features are unavailable/disabled in their drivers and would cause the plugin to crash an exception to the emulator.

cxd4:
I was thinking about having the message boxes one-time only...use a static int to store whether the function has been called yet; if so skip the DisplayError.

I haven't really made up my mind on the matter myself.  I might even do what you said and take out all of the error messages (though making the plugin message-free would not mean that any graphics would draw on the screen...).  So I guess it depends how much is left to debug and sort out.

cxd4:
updated to
* match calling convention to Microsoft's own `GL/gl.h`
* not DisplayError if debugger is not enabled in PJ64 settings integration

PR #207, opened 2015-03-04
More Updates related with Translations.

MELERIX:
- Added few Desc where missing.
- Added a Missing String in English main lang file.
- Fixed some typos in the original english strings.
- Removed few Unused Strings.
- Sorted Strings by IDs.


project64:
I am not sure about the removal of the MENUDES_ items .. the question is it worth stripping them or is it better to fix the code so it uses them, I probably removed the usage by accident when I changed some of the UI code to c++

MELERIX:
ok, now should be ok, restored menu descriptions.

MELERIX:
is ok now ?

PR #210, opened 2015-03-05
Minor Updates for package_zip.bat

MELERIX:
- Output directory will be cleaned before start to copy files.
- Removed copy line for two .chm files that are outdated.
- Added copy line for a plugin that was missing.


project64:
looks fine except I was keeping Jabo_DInput.dll out on purpose to make sure the nrage plguin was good enough to replace jabo's.

I would prefer to use an open source plugin we can support and modify.

MELERIX:
I see, then Jabo_DInput.dll should be removed from installer.iss too ?

project64:
yes, I was planning to go back to the installer when I was getting closer to release and making sure it matches the output of package_zip

MELERIX:
ok, I'll make a new pull.

now closing this one.

Issue #213, opened 2015-03-05
Glide64 for PJ64 Resident Evil 2 regression


Lithium64:
This regression already existed on PJ64 2.1

Glide64 for PJ64
![glide64_resident_evil_ii_01](https://cloud.githubusercontent.com/assets/10705488/6501482/add38800-c2f7-11e4-970b-7893ddf39b05.png)

Glide64 Final
![glide64_resident_evil_ii_05](https://cloud.githubusercontent.com/assets/10705488/6501483/addfc75a-c2f7-11e4-935b-5f1edcfd9bdc.png)

Glide64 for PJ64
![glide64_resident_evil_ii_02](https://cloud.githubusercontent.com/assets/10705488/6501498/d766e522-c2f7-11e4-988f-541dac36df0f.png)

Glide64 Final
![glide64_resident_evil_ii_06](https://cloud.githubusercontent.com/assets/10705488/6501499/d77721ee-c2f7-11e4-8159-4369c4238a48.png)

Glide64 for PJ64
![glide64_resident_evil_ii_03](https://cloud.githubusercontent.com/assets/10705488/6501506/f4dbdffe-c2f7-11e4-8aea-05688d5ca4fe.png)

Glide64 Final
![glide64_resident_evil_ii_07](https://cloud.githubusercontent.com/assets/10705488/6501507/f4e820de-c2f7-11e4-9df2-90b1fff2c413.png)

AmbientMalice:
This is caused by "Software depth buffer rendering" being disabled. I'm trying to figure out the correct command to re-enable it, then I'll add it to the Glide64.rdb

Lithium64:
@AmbientMalice This didn't work the bug is still occurring

Lithium64:
@project64 Please re-open this issue

cxd4:
@Lithium64 , you are the creator issue and should be able to administrate the status of the issue (whether it's open or closed) without having to wait on project64 to re-open it.

This is no longer vBulletin forums; people can decide the status of their own issues.

Lithium64:
@cxd4 you cannot re-open your own issues if a repo collaborator closed them

http://stackoverflow.com/questions/21333654/how-to-re-open-an-issue-in-github

cxd4:
onice !!!

project64:
ok

cxd4:
ExtremeDude2, dsx, AmbientMalice and I have all gotten this fix to work.

[...]

PR #214, opened 2015-03-05
Fix RE2 depth issues with Glide64

AmbientMalice:
This fixes https://github.com/project64/project64/issues/213

### DIFF
--- a/Config/Glide64.rdb
+++ b/Config/Glide64.rdb
@@ -3045,6 +3045,7 @@ fb_smart=1
 fix_tex_coord=128
 n64_z_scale=1
 swapmode=2
+fb_render=1
 
 [2F493DD0-2E64DFD9-C:45]
 Good Name=Resident Evil 2 (U) (V1.0)
@@ -3057,6 +3058,7 @@ fb_smart=1
 fix_tex_coord=128
 n64_z_scale=1
 swapmode=2
+fb_render=1
 
 [AA18B1A5-07DB6AEB-C:45]
 Good Name=Resident Evil 2 (U) (V1.1)
@@ -3068,6 +3070,7 @@ fb_smart=1
 fix_tex_coord=128
 n64_z_scale=1
 swapmode=2
+fb_render=1
 
 [02D8366A-6CABEF9C-C:50]
 Good Name=Road Rash 64 (E)


cxd4:
http://ft.trillian.im/785abe041074fd64135ab3318aff1ec8767ab03d/6wOcP6Ysw5XjCWgHtRXessw9BQlOz.jpg

I don't think that fixed it.
I know the check box for software depth buffer rendering is checked this time after pulling from this commit, but it isn't fixed for me either.  If it is fixed for AmbientMalice then something external to the Glide64 configuration must be helping it.

Frank-74:
I'm wondering if PJ_Glide64 is from the Glide64 final version.  As there are other issues that don't happen with Glide64 final.  The bond portrait in Goldeneye for example.  Displays correctly in Glide final, yet is split in 2 with a line down the middle in PJ_Glide.

AmbientMalice:
@cxd4 @Frank-74 I can think think of a number of possible causes. They all revolve around Framebuffer Objects, and by association ATI/AMD GPUs.

FBOs are disabled by default in Glide64. Enabling them breaks RE2's depth. Traditionally, FBOs are associated with better performance on ATI hardware. I don't really know why.

Are either of you using AMD GPUs?

Frank-74:
Yes, on board ATI Radeon Xpress.

AmbientMalice:
Do you have "Use Frame Buffer Objects" disabled?

Frank-74:
Yes FBO's are disabled.  With Software depth buffer rendering enabled and FBO's disabled, it works.  Its just very slow with software depth buffer.

Lithium64:
@AmbientMalice HD 4650 1GB here with FBO disabled

Frank-74:
You need to uncheck Hide Advanced Settings in pj64 options to see the Emulation Settings tab in Glide with the software depth buffer option.

Anyway, now I'm getting a 'GL_INVALID_OPERATION" , "Update Screen" message box, as soon as the spinning logo disappears, only when I have Software depth buffer enabled.  Using the Glide debug build.

cxd4:
> Are either of you using AMD GPUs?

Nope I have a GeForce.

> Anyway, now I'm getting a 'GL_INVALID_OPERATION" , "Update Screen" message box, as soon as the spinning logo disappears, only when I have Software depth buffer enabled. Using the Glide debug build.

I put the OpenGL debugging only in UpdateScreen since that's the function that gets constantly called by Project64 all the time.  I think I should move the debugging to happen at the end of each of the Glitch64 GL command segments.

`GL_INVALID_OPERATION` means different things for different functions, so I might have to make the debugger more specific.  It can mean corrupt data in the vertex arrays, or it can happen with a complete loss of the GL context.

PR #216, opened 2015-03-05
RDB fixes to "Battlezone:  Rize of the Black Dogs"

cxd4:
A couple issues with this game were raised on IRC when theboy181 mentioned some flickering polygons on/off with all graphics plugins/RSP plugins/different emulator cores besides Project64.  Actually the real issue was that "Counter Factor" should be set to not 2, 3 or 4, but to 1.

This also improves the high-res video display with a pixel-accurate plugin and 8 MB RDRAM expansion pak installed, though I would rather not set RDRAM size from 4 to 8 MB in the RDB yet since Project64's VI timing is still off and has jittery scan-line interlacing.  So currently have it set to 4 MB for low-res graphics but less line glitching.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -845,8 +845,7 @@ Counter Factor=3
 [55D4C4CE-7753C78A-C:45]
 Good Name=Battlezone - Rise of the Black Dogs (U)
 Internal Name=BATTLEZONE
-Status=Broken (core)
-Core Note=broken graphics; use older PJ64
+Status=Compatible
 AiCountPerBytes=200
 Counter Factor=1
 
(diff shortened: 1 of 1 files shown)


dsx-:
still looks fucked for me http://puu.sh/gopRt/02cbe99d50.jpg

cxd4:
We don't test that plugin.  When this issue was reported on IRC, it was caused by the Project64 core, and angrylion's plugin didn't fix it.  Now no more issues were reported with angrylion's plugin.

dsx-:
am i able to have the details of this IRC? i wanna discuss some things

cxd4:
If you are new to IRC, then I guess you could join using a web client over HTTP.
I really do hate referring people to use webchat lol, but I guess it's just easier to explain for people starting.
http://webchat.freenode.net/

Just put in a nickname (that isn't currently in use by anyone else) and the CAPTCHA, click "Connect", then enter `/join #project64` to join the channel.  (Alternatively enter the channel name in the form first before clicking "Connect", then no /JOIN command is needed.)

PR #227, opened 2015-03-06
MusyX\MORT\Factor 5 sound fixes+misc.

AmbientMalice:
With these changes, games such as RE2 now use Fixed Audio timing with appropriate VI\AI values, and now have near-perfect audio that will no longer bug out - either randomly, due to low framerates, or when loading save-states. Infernal Machine no longer has crackling audio. (Yes, I'm obsessed with fixing that game.) San Francisco Rush 2049 seems improved. Rogue Squadron\Naboo have fixed sound. (Except for weird crackling during SW text scrolling.) Tarzan, Rugrats, and other games using Factor 5 tech were adjusted. Some ran fine on vanilla settings for some reason.

I disabled "Start Changed" for Naboo and Infernal Machine, since having it enabled causes significant stuttering in some menus.

FYI, MORT is the name of the speech codec used by Factor 5 games.

I decided on VI=2200 and AI=785 for most titles because those values hit a sweet spot between stopping crackle and keeping sound in synch. (Some games were happy with AI=400.) Some further tweaking might provide better results.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -1,4 +1,4 @@
-﻿// ============ RDB for PJ64 v2.2. GoodN64 v202b ====================================
+// ============ RDB for PJ64 v2.2. GoodN64 v202b ====================================
 // PJ64 v2.2 Official RDB
 // Not for use with PJ64 v1.6 or previous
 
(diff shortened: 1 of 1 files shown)


Frank-74:
I found TWINE to be ok already in 2.1 with just fixed audio timing disabled and HLE audio on, using Azimer's 0.7 DS8 WIP4.  The VI's were 2200 and AI's 800.  So only 15 AI's difference.

AmbientMalice:
In this case, I decided to pretend that Azimer's doesn't exist because although I like the plugin, it wrecks every single attempt to get MORT/MusyX audio to work consistently on both Jabo and Azimer's. It's especially teeth-grating with Resident Evil, where the timings have to be juuuuuust right or the FMVs are out of synch. For some reason, I can't get Azimer's AI handling to turn off - probably the most significant factor.

AmbientMalice:
The 15 VIs difference between 785 and 800 is the difference between smooth-ish sound and constant juddering. Especially with Fixed Audio timing, which throws everything out of whack.

oddMLan:
@project64 Is it true that you were planning to implement some sort of auto-detection of the VI/AI parameters for every game? If that's possible, I think it would fix a lot of audio timing in finicky games like Factor 5 and MusiX.

project64:
@oddMLan I want to get as accurate to the real n64 as possible, but until then if you can find something or there is something I can change to make things play better then I will make those changes

PR #239, opened 2015-03-07
Hydro Thunder enable 8MB and new audio timing.

AmbientMalice:
Enabling the expansion pack allows 4 player MP. Was only enabled in US version before.


AmbientMalice:
Just noticed I accidentally added a duplicated AI entry to (U) rom. Will make new pull request later.

PR #243, opened 2015-03-07
changes and corrections to RDB

dsx-:
these are good


Lithium64:
@dsx- Change Reg Cache=0 to Reg Cache=No

AmbientMalice:
Doraemon 3 works now? That's neat. Now if only there were an English translation.

Frank-74:
@AmbientMalice Yes, it works perfectly.  Jabo's has a slight mistake on one scene in the intro.  But Glide is perfect and a smooth 60VI's even on my old crappy gfx.  I recommend a change to Counter Factor=1.

PR #245, opened 2015-03-07
Resident Evil 2 rdb fixes

Lithium64:
Changed back RDRAM to 4, if set to 8 will cause resolution issues with Glide64

### DIFF
--- a/Config/Glide64.rdb
+++ b/Config/Glide64.rdb
@@ -482,6 +482,7 @@ fb_smart=1
 fix_tex_coord=128
 n64_z_scale=1
 swapmode=2
+fb_render=1
 
 [7C64E6DB-55B924DB-C:50]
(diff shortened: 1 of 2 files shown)


AmbientMalice:
I was wary of setting it to 8MB to begin with. Hopefully GLideN64 will handle the game properly.

AmbientMalice:
Out of curiosity, what do you mean by "resolution issues"? I have noticed that the screen flashes with garbage between camera angle changes more often and more noticably with 8MB compared to 4MB.

Lithium64:
@AmbientMalice Resolution issues like these

glide64_resident_evil_ii_01
glide64_resident_evil_ii_02
glide64_resident_evil_ii_03
glide64_resident_evil_ii_04

AmbientMalice:
I see. Thanks for the explanation.

PR #246, opened 2015-03-08
Hydro Thunder 8MB and new audio.

AmbientMalice:
8MB enables 4 player. Was only enabled in (U) version.


cxd4:
You can append to an existing pull request by pushing commits to the branch it's linked to.
That way we don't have to start a new pull request because the old one forgot a single addition.

AmbientMalice:
I did actually try that, but I couldn't figure out how to make the branch alterations show up in the pull request. I'll try to do that in future, though.

PR #247, opened 2015-03-08
Fix Resident Evil FMV audio.

AmbientMalice:
This fix isn't quite satisfactory, but with 2200\875, the last 30 seconds or so of the intro have no sound effects, only music. (A bug that could affect other FMVs throughout the game in theory.) 1500\530 has a wee bit of pop and crackle, but the FMVs now work correctly.


AmbientMalice:
There's a change the timings aren't quite right for the PAL version. I'll have to test later.

MELERIX:
I'm wondering if all of these values are calculated with a formula or are just related made by guessing and trying ?

AmbientMalice:
Mostly guesswork. But with FAT, most MusyX/MORT games tend to work best within certain ranges. RE2 used to use 1400/800 without FAT. Enabling FAT threw everything off, but 2200/785 FAT worked okayish except for the missing FMV sounds. So I went for 1500VI (Which is arguably the ideal baseline VI from what I've read) and tweaked AI values until the FMVs were synched and the crackling stopped. The PAL version may need a slight adjustment for best results, but it seems passable for now.

I'd prefer to not use Fixed Audio Timing, since it causes so many problems. But it's the only thing that fixes Factor 5 music and speech problems consistently, although a handful of such games seem okay without it.

MELERIX:
I noticed thay using Jabo's DirectSound 1.7.0.7 plugin with "Sync game to audio" option enabled in the same plugin, fixed all my audio problems (for example the music in Super Mario 64 & Diddy Kong Racing), without having to modify any value in rom settings, this is really weird.

AmbientMalice:
The problem is that MusyX games are using a completely different method of playing music to normal N64 titles. For most N64 games, FAT isn't needed.

I get the feeling that MusyX/MORT games were never properly tested in the past. :( Take the RSP Audio Signal fix, for example. Yes, it helped MORT speech play. But on its own, it did so unreliably. Sometimes speech drops out. Sometimes the hideous screeching starts due to a framerate dip and then the game freezes due to weird "wait until sample finishes and then trigger next event" behavior. (The infamous RE2 gun shop freeze is an extreme example. Infernal Machine will just sit there and start cycling through random speech samples if it gets out of synch.)

I'm just trying to do my part so PJ64 can actually run these games - at least until PJ64 has better audio timing code.

PR #248, opened 2015-03-08
"Fixed" Infernal Machine (U) sound.

AmbientMalice:
Okay, so it turns out that the (U) version of Infernal Machine *hates* the settings that provide good audio for the (E) version. (I have no clue why I didn't notice this until now.) After a lot of dicking around, I found that AI=540 provided a usable compromise, where the sound is in synch and doesn't crackle like crazy. The problem is that the main menu music is now too fast. None of the other Factor 5 games behave this way.

In an ideal world the (U) version would be placed into a refrigerator and have a small atomic device detonated next to it. Until PJ64 gets proper audio synch or Azimer does something magical, I don't see that version working properly.


AmbientMalice:
operation_upshot-knothole_-_badger_001

This is what I want to do to the (U) version of this game.

LegendOfDragoon:
Doesn't it sound good enough with Azimer's, with fixed audio timing disabled? If not, what parts specifically sound bad? I've been doing lots of testing on audio lately. I'm focusing on games that have sound issues on many different plugins (like WDC).

AmbientMalice:
Last I checked, without Fixed Audio Timing you eventually get garbled voices and long pauses due to the game trying to control the pace of cutscenes using voice acting samples or something like that, even if the game runs okay initially. (Also, without FAT you can't load save states properly.)

Forced HLE might be a different matter. I should test more with Azimer's.

This is mainly a quick and dirty fix. Azimer's runs the (E) version seemingly fine using the 2200/785 FAT settings so far as I can tell.

It's possible that Factor 5 screwed around with MusyX between the (U) and (E) versions. The game was apparently prone to crashing during music transitions on real N64 hardware.

PR #250, opened 2015-03-09
minor clean-ups to how MIPS GPR $zero is handled in VR4300 and RSP CPUs

cxd4:
In the master CPU (the modified R4300), only 7 interpreter op-codes completely prevented the possibility of overwriting $zero (in either 32- or 64-bit MIPS mode).  It was decided this shouldn't really be necessary, since it seemed more sizable, smaller and readable to just handle the read-only state of MIPS register $zero in the running CPU loop rather than the individual op-codes.

A similar change was made for the RSP, which should fix the vulnerability to overwriting $0 for RSP MFC0, MFC2, CFC2 and any of the primary op-codes for memory loads and stores (e.g., LB and SB).


oddMLan:
dat branch name tho.

cxd4:
I don't always name branches in a clear and unambiguous way, but when I do, I don't.

Issue #76, opened 2015-02-13
Indiana Jones gfx issues in CPU core


cxd4:
I was able to reproduce the findings of Ambient_Malice at this thread spot:
http://forum.pj64-emu.com/showthread.php?p=59953#post59953

I wasn't able to change RDB settings so far in a way that would fix it, and it apparently hasn't had the issues on 1964 0.8.5 interpreter or Mupen64Plus.  So I will have to try testing older versions of this repository to see as of which commit may have broken this (unless the issue goes back further than GitHub itself >.< hopefully not).

exhalatio:
Does mtc0 opcode revert to 1964 0.8.5, the GFX seem to be fixed anyway?

cxd4:
You suppose it may be an interpreter bug with MTC0 on the main CPU then?
Or I am simply not following.

LegendOfDragoon:
Could be. 1964 0.85's code looks a lot different for that instruction. I'm very intrigued.

exhalatio:
Yes I suppose a bug is in the MTC0 on main CPU.
I don't understand technical thing , but it seems there is a bug in The status register.

project64:
why do you think it might be in MTC0?

LegendOfDragoon:
@exhalatio probably studied emulator core code well. @project64 , you should take a look at 1964 0.85's source. You'll see that it implements that instruction quite differently.

project64:
in what way did they handle it differently?

AmbientMalice:
Filed a MTC0 error report for Supercross 2000 in LLE mode that might be relevant.

https://github.com/project64/project64/issues/104

exhalatio:
FixFpuLocations have two Routines, if forced as 1st routine work well.
This game does not seem to hit the (STATUS_REGISTER & STATUS_FR) of == 0 conditions.

AmbientMalice:
What about Battle for Naboo? Since both games seem to share the same bugs.

exhalatio:
Yeah, Battle for Naboo also seems to work well.
![naboo](https://cloud.githubusercontent.com/assets/7089513/6320800/7a625c2c-bb2b-11e4-8c3f-406d60d23700.jpg)

AmbientMalice:
The radar and skybox are broken. (Yes, I'm pointing out the obvious, but they work on other emus.)

Good work, anyway.

LegendOfDragoon:
Pretty sure he's using Jabo 1.7.0.57 though :P .

[...]

PR #251, opened 2015-03-09
force FixFpuLocations in Indiana Jones gfx issues

exhalatio:


AmbientMalice:
This seems to break Shadow Man. So as much as I'd like it to be added, I'm a bit... uneasy with its current state.

cxd4:
That shouldn't be all that surprising.  He knows that this fix is in the form of taking out the broken implementation, but removed/no implementation at all is not necessary better than an existing but broken one.  So this commit is just a bypass rather than a genuine fix; format investigation there on what really should happen is still pending.

Frank-74:
When the FR bit in the Status register equals [b]zero[/b], the Floating-Point general purpose registers (FGRs) are 32-bits wide.
To retain single-precision floating-point format data, sixteen even number registers out of thirty-two FGRs can be accessed.
To retain double-precision floating-point format data, even number registers are used for low-order bits of data, and odd number registers for high-order bits.
The registers are used as even-odd pairs, and can retain sixteen double-precision format data.

When the FR bit in the Status register equals [b]one[/b], the Floating-Point general purpose registers (FGRs) are 64-bits wide.
To retain single-precision floating-point format data, low-order bits of thirty-two FGRs are used.
To retain double-precision floating-point format data, thirty-two FGRs are used.

cxd4:
It's been years since I've read that manual.
I remember pretty much skimming through the CP1 stuff since I mostly cared just about the R4000 basics.

NEC also has some diagrams of the FR Status bit in their manual on their modifications to the R4300i:
http://datasheets.chipdb.org/NEC/Vr-Series/Vr43xx/U10504EJ7V0UMJ1.pdf

Searched over the whole thing and dug back up a few interesting pages.

Page 166 is a diagram of the CP0 register `Status`, showing the bit encodings describing the FR bit.

Page 208 tells you the 3 formats that a floating-point register can be accessed during an operation, with a VERY helpful diagram on the next page 209 showing the visual layout when FR=0/1.

>FPR is physically configured with General Purpose registers (FGRs). When the
FR bit in the Status register equals 0, the FPR is configured with two 32-bit FGRs.
When the FR bit in the Status register equals 1, the FPR is configured with a single
64-bit FGR.

>If the FR bit of the Status register is set to 1, all the thirty-two 64-bit floating-point
registers defined by the MIPSIII architecture can be accessed. If this bit is cleared,
the processor accesses the sixteen 64-bit floating-point registers defined by the
MIPSII architecture.

And everything else I think is repeated and re-stated junk throughout each of the COP1 op-codes.

Frank-74:
From the same manual, chapter 17, pages 557/558, shows the 2 routines ( 32bit and 64bit ) to retrieve the value of an FPR or to change the value of an FGR.
```
32 Bit Mode
value <-- ValueFPR(fpr, fmt)
/* undefined for odd fpr */
case fmt of
S, W:
value <-- FGR[fpr+0]
D:
value <-- FGR[fpr+1] || FGR[fpr+0]
end
StoreFPR(fpr, fmt, value):
/* undefined for odd fpr */
case fmt of
S, W:
FGR[fpr+1] <-- undefined
FGR[fpr+0] <-- value
D:
FGR[fpr+1] <-- value63...32
FGR[fpr+0] <-- value31...0
end
```

cxd4:
@AmbientMalice you mentioned at the bottom of this post that the issue is improved on 1.4:
http://forum.pj64-emu.com/showpost.php?p=59962

I mostly just tested right after the menu when you press start, since what you were taking a screenshot of in that post wasn't something I knew how to get to as quickly, but, I had a look at the way this function was being done in Project64 1.4 source:
```c
void SetFpuLocations (void) {
	int count;

	if ((STATUS_REGISTER & STATUS_FR) == 0) {
		for (count = 0; count < 32; count ++) {
			FPRFloatLocation[count] = (void *)(&FPR[count >> 1].W[count & 1]);
			//FPRDoubleLocation[count] = FPRFloatLocation[count];
			FPRDoubleLocation[count] = (void *)(&FPR[count >> 1].DW);
		}
	} else {
		for (count = 0; count < 32; count ++) {
			FPRFloatLocation[count] = (void *)(&FPR[count].W[0]);
			//FPRDoubleLocation[count] = FPRFloatLocation[count];
			FPRDoubleLocation[count] = (void *)(&FPR[count].DW);
		}
	}
}
```
(Wow that looks sexy it's even written in C!)

The key difference between this and how `FixFpuLocations` is done after pj64 1.4, up till current 2.x, is that under the `else` it reads `(void *)(&FPR[count].W[0]);` instead of `&m_FPR[count].F[1];` (address of first 32-bit half instead of the address of second 32-bit half) (the W and F mean the same data size so doesn't matter).  However if I change from 1 to 0 that doesn't fix anything at all.

Since the function is otherwise identical in both Project64 1.4 and Project64 2.x, I fear that the success of this hack fix may be a sign of only a completely unrelated interpreter bug somewhere else, which might be causing STATUS_FR to not get set when it should have been set.

AmbientMalice:
@cxd4 I'm glad my testing was helpful, even if the results aren't exactly desirable - I'm aware of how troublesome fixing interpreter bugs can be.

Frank-74:
Maybe this could be made an rdb option?

AmbientMalice:
@Frank-74 That occured to me, also. It wouldn't help PJ64's reputation as an emulator that fixes shit with hacks. There's a risk that a half-assed RDB fix would reduce the incentive to find out the real cause.

On a related note, I wish the Trading Post CPU recompiler freeze could be fixed. Either fix the recompiler or use a cheat to bypass the Trading Post. (The latter is terribly hacky, but it's plausible since the game bypasses the TP itself if the level has already been completed.)

Naboo and Infernal Machine are very close to working satisfactorily on PJ64. GLideN64 will emulate their graphics correct, even if the LLE mode isn't as fast as it could be. We're on the verge of PJ64 emulating games "out of the box" many folks gave up on years ago.

Oh, yea, and the RSP recompiler needs fixing, too because it freezes with Naboo\Indy.

theboy181:
I was looking at this game today, and its like the camera is in the wrong location? Is there a plugin that allows you control the position of the camera. I would like to see how this game looks with a proper camera position.

AmbientMalice:
@theboy181 The camera position is just a side effect. You need to compile PJ64 with the FixFPULocations hack.

exhalatio:
I can't solve this issue without Dirty hacks.
Must I close this pull request?

AmbientMalice:
@exhalatio I think @Legendofdragoon was taking a look at the issue, but I dunno whether he was making any progress. I don't think there's any need to close the issue.

cxd4:
We want to get to reverse-engineering it first and use this as an opportunity to compare emulator to hardware to game results; that's probably why this hasn't already been closed--a reminder.

exhalatio:
It makes sense.

LegendOfDragoon:
>@exhalatio I think @Legendofdragoon was taking a look at the issue, but I dunno whether he was making any progress.

Nothing conclusive yet, but I'm attempting to track down multiple issues.

Frank-74:
I've made it an RDB option in this branch. https://github.com/Frank-74/project64/tree/ForceFixFpuLoc-RDB-Setting

AmbientMalice:
@Frank-74 I'm 50/50 on using a hack. Regardless, it shouldn't be merged until the CPU interpreter/recompiler freeze bugs are fixed, IMO.

Frank-74:
I got tired of building a separate exe that is only useful for 2 games.

cxd4:
The trick is that finding the genuine fix is impossible for anyone inexperienced with the CPU ISA.

Regressions in interpreters are always far harder to fix than regressions in recompilers:  You have nothing to cross-reference with that is any more accurate, save for hardware reversing which you will have to do.  Basically, your cheat sheet is gone, and all one can do otherwise is guess hack workarounds or persist to substitute the thorough experience necessary for documentation (which is only capable of summarizing said necessary experience).

project64:
https://github.com/project64/project64/commit/c7f895737ab3abc257931c4f9a7d65890a037f02

this should be correct fix now

AmbientMalice:
Seems to work now. Good work. PR creator has disappeared so you'll have to close this yourself.

Game still crashes/freezes at the end of each mission, though.

Frank-74:
Works perfectly now.

This and #76 can be closed now.

PR #258, opened 2015-03-10
Ambient Needs to Stop Touching Things.

AmbientMalice:
Bunch of audio adjustments. Plus CF1 for Hydro Thunder. Plus I disabled RSP Audio Signal because it doesn't seem to be needed with FAT. Also deleted some duplicate entries. Plus some Infernal Machine adjustments.

Edit:
Re-enabled RSP Audio Signal for Tarzan, since it freezes without it. Also made some drastic VI\AI changes to NBA: Showtime - NBA on NBC. Now the game runs like dogshit but the ingame timers aren't running way too fast.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -314,24 +314,23 @@ Bad ROM?.AutoFullScreen=False
 
 // ---- END OF HEADER, START OF ROM ENTRIES --------------------------
 //================  0-9  ================
-[3B941695-F90A5EEB-C:50]
 Good Name=007 - The World is Not Enough (E) (M3)
 Internal Name=TWINE
 Status=Compatible
-ViRefresh=2200
-AiCountPerBytes=785
+ViRefresh=1500
+AiCountPerBytes=530
 Fixed Audio=1
-Sync Audio=0
+Sync Audio=No
 RDRAM Size=8
 
 [033F4C13-319EE7A7-C:45]
 Good Name=007 - The World is Not Enough (U)
 Internal Name=TWINE
 Status=Compatible
-ViRefresh=2200
-AiCountPerBytes=785
+ViRefresh=1500
+AiCountPerBytes=530
 Fixed Audio=1
-Sync Audio=0
+Sync Audio=No
 RDRAM Size=8
 
 [58FD3F25-D92EAA8D-C:50]
@@ -917,6 +916,8 @@ Emulate Clear=0
 Primary Frame Buffer=0
 Self Texture=0
 
+
+
 [7EAE2488-9D40A35A-C:4A]
 Good Name=Biohazard 2 (J)
 Internal Name=BioHazard II
@@ -926,9 +927,8 @@ Plugin Note=[video] depth problem (use Glide64)
 32bit=No
 ViRefresh=1500
 AiCountPerBytes=530
-Sync Audio=0
+Sync Audio=No
 Fixed Audio=1
-Audio Signal=Yes
 Counter Factor=1
 
 [7C64E6DB-55B924DB-C:50]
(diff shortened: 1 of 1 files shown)


Frank-74:
I can see lots of Sync Audio=0 still left.  Can you search/replace all?  Search for Sync Audio=0 and replace ALL with Sync Audio=No.  Should only take a split second.

AmbientMalice:
Glad to oblige.

AmbientMalice:
Glad to help.

Lithium64:
@AmbientMalice @Frank-74 sync audio=0 works fine no need to change it

Frank-74:
There are still some deprecated options in the RDB that no longer have any function.  CustomSMM is no longer used.  And I can't find a use for Emulate Clear or Clear Frame either.

Lithium64:
@Frank-74  Emulate Clear and Clear Frame is for Jabo's video plugins, Custom SMM don't know what this setting do and if still works, 

Audio Signal is needed for all these games even using hatcat rsp plugin.

AmbientMalice:
@Lithium64 how does removing Audio Signal break them? My understanding is that it is supposed to fix issues such as missing MORT speech and garbled sound in San Francisco Rush 2049. But using FAT and specific VI/AI values provides the same results, or at least seems to.

Lithium64:
@AmbientMalice Games like Tarzan freeze during gameplay if RSP audio signal is disabled

AmbientMalice:
@Lithium64 Even with FAT enabled? I played some Tarzan yesterday, and everything seemed okay. (It was only a few minutes, though.) Such freezes are usually caused by the mort speech codec trying to play. (RE2 freezes for the same reason.)

I suppose I should have tested more thoroughly.

LegendOfDragoon:
Better safe than sorry xD.

Lithium64:
@AmbientMalice I think if this setting's there is for some reason, I did several tests here and the game just stopped freeze when I identify that the audio signal was not really on.

AmbientMalice:
The RSP Audio Signal setting fixes broken MusyX/MORT sound in certain titles. The problem is that FAT (seemingly) also fixes broken MusyX/MORT sound, and does so more reliably. (Ignoring the wonky behavior of FAT, anyway.)

It seemed like double handling to have both settings enabled. I will properly test Tarzan later to determine whether RSP Audio Signal has any affect on stability with FAT enabled.

Lithium64:
@AmbientMalice I just tested with your changes and the game continues freezing but in other points

AmbientMalice:
@Lithium64 Okay, thanks. I'll revert the setting for Tarzan when I've got access to my main PC in a few hours.

Do you feel I should re-enable for games that don't seem freeze-prone? Rush 2049 used RSP AS to make the music clearer - but FAT handles that. My guess is this is MORT-related, so perhaps it should be enabled for 3rd party titles that use MORT? (Rush only uses MusyX, AFAIK.)

AmbientMalice:
Hmm... Yea, that's what I'll do. RE2, Tarzan, TWINE, and various other 3rd party titles - I'll turn RSP Audio Signal on. Better safe than sorry. Don't want people running into random freezes.

Factor 5's own games don't seem to need it, so I'll leave it off for them. (It's actually pretty weird how their own games seem to have voices no matter what.)

Lithium64:
@AmbientMalice You should play a bit each of these games to make sure they are not freezing

AmbientMalice:
The World is Not Enough seems fine with RSPAS

AmbientMalice:
To my befuddlement, I've discovered these games didn't have RSP Audio Signal enabled until I enabled it 5 days ago, at the same time I added FAT. https://github.com/project64/project64/pull/227/files

Almost finished testing.

AmbientMalice:
The other games seem fine. So this is ready to be merged if Zilmar wishes.

PR #273, opened 2015-03-14
Removed more default values from Project64.rdb

MELERIX:
Clear Frame=0/1/2
Self Texture=0/1
Primary Frame Buffer=0/1
Emulate Clear=0/1
Culling=0/1

Based in Documentation from:
http://en.helpdoc-online.com/project64_1.6/source/html/4%20configuration/plugin_video_d3d6_romsettings.htm
&
http://www.pj64-emu.com/manual-and-guides/configuration%3a-jabo-direct3d8/rom-settings-tab-d3d8-v1.6/#zAnswerList_307

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -174,21 +174,16 @@ Good Name=1080 Snowboarding (E) (M4)
 Internal Name=1080 SNOWBOARDING
 Status=Compatible
 Plugin Note=[video] framebuffer:screen (see GameFAQ)
-Clear Frame=0
 Culling=1
 Emulate Clear=1
-Primary Frame Buffer=0
-Self Texture=0
 
 [1FBAF161-2C1C54F1-C:41]
 Good Name=1080 Snowboarding (JU) (M2)
 Internal Name=1080 SNOWBOARDING
 Status=Compatible
-Clear Frame=0
 Culling=1
 Emulate Clear=1
 FuncFind=1
-Primary Frame Buffer=0
 SMM-Cache=0
 SMM-PI DMA=0
 SMM-TLB=0
@@ -236,7 +231,6 @@ Status=Compatible
 Good Name=A Bug's Life (E)
 Internal Name=A Bug's Life
 Status=Compatible
-Clear Frame=0
 Counter Factor=1
 32bit=No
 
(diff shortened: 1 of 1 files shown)


Lithium64:
I thought there is no documentation about these settings

cxd4:
Most of that is stuff from Jabo D3D6 1.5.2 plugin which I think is no longer included.  So I guess removing them from the RDB as long as the internal defaults within the plugin match them isn't exactly destructive.

MELERIX:
yes, just that I can't find info/documentation about plugin version 1.7, so no idea which values still are used.

@project64 have you info about plugin version 1.7 and which values still are used ?

cxd4:
only one of those that's still used is _Clear Frame_

Frank-74:
Clear Frame setting in RDB doesn't change the setting in the plugin.  And changing the setting in the plugin doesn't write it to the .cfg either.  Only on 1.6 plugins it works with the rdb in the emulator folder.

Culling works with all Jabo's from 1.5 to 1.7.  Even though the option is only selectable in 1.5.2.  Especially noticeable in games like Perfect Dark or GoldenEye.  Switch off culling and the frame rate will drop a lot.

MELERIX:
seems the plugin is not loading the values from the RDB, even if you change manually "Clear Frame" to 2 in RDB, the plugin still shows "Default" and even if you make changes in "Rom Settings" none of the options is saved.

so now I'm wondering if the plugin is broken ? or is something in Project64 causing the bug ?

Lithium64:
@Frank-74 @MELERIX for these settings working is needed leave a empty rdb on pj64 root folder

MELERIX:
tested and nothing... only Direct3D8-Direct3DPipe is saved but in Project64.cfg

Frank-74:
These are the only Jabo RDB options I could get working.
>// Jabo Options
>Direct3D8-Direct3DPipe	=0
>Direct3D8-DesiredAspect=0|1|2|3	// Default|Stretch|Force 4:3|Force 16:9
>Direct3D8-2xSai=0
>Direct3D8-ForceFilter=0
>Direct3D8-Brightness= 0|0-30
>Dsound-SyncAudio=0
>Culling=0

cxd4:
I know clear frame worked, at least in 1.6, as it was needed for several games to display right in Jabo's 1.6 plugin.  But maybe after 1.7 zilmar finally removed it so that is why you can't get it to work.

PR #276, opened 2015-03-16
Removed more default values from Project64.rdb

MELERIX:
CPU Type=Recompiler
FuncFind=1
RDRAM Size=4
SMM-Cache=1
SMM-PI DMA=1
SMM-TLB=1
SMM-Protect=0

Based on info from: https://github.com/project64/project64/blob/master/Source/Project64/Settings/Settings%20Class.cpp


MELERIX:
@cxd4 4MB is the default capacity in each N64 console, thats why is the default value in PJ64 too, and 8MB is mostly for the games that use "Expansion Pack" accesory.

http://nintendo.wikia.com/wiki/Nintendo_64_Expansion_Pak

cxd4:
mk and me saying that removing the RDRAM Size=4 setting from the RDB was a bad idea, implies that I'm unaware of the N64's default RAM capacity how exactly?

It doesn't.  It says there's no reason to remove a setting that's needed, which was just done unless we have tests showing the game is indifferent to 4 or 8.

MELERIX:
removing RDRAM Size=4 is just due "4" is already a default value in core, so there is no sense to have the same value in the RBD again, except that you are going to use the RBD with older Project64 versions and wants to mantain backward compatibility, but note that there is warning about this in the first lines of the RBD that says:

"Not for use with PJ64 v1.6 or previous"

btw, about which games should use 4 or 8, there is a complete official list of the games that use or require Expansion Pak, so I can update the whole RDB later based on that info.

cxd4:
>removing RDRAM Size=4 is just due "4" is already a default value in core, so there is no sense to have the same value in the RBD again,

Yes there is.
All default values are changeable, either via the source or by the settings to Project64.
The only reason the default setting for RDRAM size is 4 is because the user leaves it set that way.  If they prefer having 8 MB by default, then your commit breaks this ROM if it requires 4 MB.  Otherwise, why allow users to change the defaults?

Even if the users couldn't change the defaults, the fact that a ROM *requires* 4 MB to be set and not 8 MB is more than enough incentive to document this situation by adding RDRAM Size=4 to document that a game actually requires it.

MELERIX:
in the N64 console all the games uses 4mb by default (like in Project64 now), and there is only few games (63 according to Nintendo/wiki info) that uses or require 8mb (Expansion Pak). 

if in a future the core defaults values are changed or something related, we can update the "RDB info.txt" file, this file have info about the defaults, also if needed we can add the list of the games that require 8mb (Expansion Pak) there, so the users will be informed about this.

PR #285, opened 2015-03-18
Continue updating UI translations

samuelyuan:
Some of the text in the original files is outdated.


Source/Project64/Multilanguage/Language Class.cpp
```diff
@@ -269,13 +270,16 @@ void CLanguage::LoadDefaultStrings (void)
 	DEF_STR(ROM_USE_TLB,         L"Use TLB");
 	DEF_STR(ROM_REG_CACHE,       L"Register caching");
 	DEF_STR(ROM_DELAY_SI,        L"Delay SI Interrupt");
-	DEF_STR(ROM_SP_HACK,         L"SP Hack");
+	DEF_STR(ROM_VIREFRESH,       L"Vi Refresh Rate: ");
```
MELERIX: why you changed SP _HACK to VIREFRESH ?

samuelyuan: There wasn't enough space to add new strings, so I decided to remove one that was unused. Since the UI doesn't have any string called "SP Hack", I knew it was safe to remove

MELERIX: in translations now we have HACK_SP, previously it was called FAST_SP but is supposed to be the same thing.

maybe we have a typo in core strings ?

samuelyuan: I didn't know that. I wasn't sure if Fast SP and SP Hack were related. Some of the strings are never used, so they may have changed in one file, but never updated in another.

cxd4: All I know is in PJ64 1.6, it was called "SP Hack", and in 1.7.0.50+, it's renamed to "Fast SP".

So provided what MELERIX says is true, then the macro names were backwards from the text string names in the GUI when running the application.

MELERIX: ok, I've renamed Fast SP to SP Hack: https://github.com/project64/project64/pull/297

samuelyuan: Alright I added SP Hack back into the strings and put the other string somewhere else

Lang/English.pj.Lang
```diff
@@ -274,13 +276,16 @@
 #527#  "Use TLB"
 #528#  "Register caching"
 #529#  "Delay SI Interrupt"
-#530#  "SP Hack"
+#530#  "Fast SP"
```
MELERIX: also revert this line :P

Source/Project64/Multilanguage/Language Class.cpp
```diff
@@ -269,13 +271,16 @@ void CLanguage::LoadDefaultStrings (void)
 	DEF_STR(ROM_USE_TLB,         L"Use TLB");
 	DEF_STR(ROM_REG_CACHE,       L"Register caching");
 	DEF_STR(ROM_DELAY_SI,        L"Delay SI Interrupt");
-	DEF_STR(ROM_SP_HACK,         L"SP Hack");
+	DEF_STR(ROM_SP_HACK,         L"Fast SP");
```
MELERIX: revert this line too.

Lang/ChineseB5.pj.lang
```diff
@@ -1,426 +0,0 @@
-﻿/*********************************************************************************
```
MELERIX: remember that if you remove this file, you have to remove line that call this file from: Installer.iss and package_zip.bat

Frank-74:
SP Hack is still written to the CFG instead of Fast SP.  That needs changing as well.

samuelyuan:
All of the references to "SP Hack" should have been changed to "Fast SP".

Source/Project64/User Interface/Settings Config.cpp
```diff
@@ -88,6 +88,13 @@ LRESULT	CSettingConfig::OnInitDialog(UINT /*uMsg*/, WPARAM /*wParam*/, LPARAM /*
 
 	CConfigSettingSection * SettingsSection;
 
+	//Set the text for all gui Items
+	SetDlgItemTextW(m_hWnd, IDC_RESET_PAGE, GS(BOTTOM_RESET_PAGE));
+	SetDlgItemTextW(m_hWnd, IDC_RESET_ALL, GS(BOTTOM_RESET_ALL));
+	SetDlgItemTextW(m_hWnd, IDOK, GS(CHEAT_OK));
```
project64: I know this works ...I really should have just done a OK, CANCEL, APPLY .. since we now are applying a cheat string to a different code unrelated to cheat

PR #287, opened 2015-03-18
converted most components to Project64 for Allman-style braces

cxd4:
More often now, it seems habitual to make commits that simultaneously change both the braces style and spacing, which makes the actual purpose behind the commits a bit more buried (as far as the part of the algorithm that actually gets changed).  Since it can feel a bit too weird to have separate commits for changing braces style preceding adjacent changes right after, I thought I would try to just get it over with.

I guess I was partially successful.  It's too complex to do all of it in one pull request due to the size of all the changes combined, so I left the following things alone for the time being:
* the interpreter CPU and op-code functions to the RSP
* the RSP compiler
* the entire Project64 core component (turns out this should have its own PR)
* PJ64 Glide64 (seems to already be in Allman, possibly get replaced with GLideN64 anyway)
* external things like the SortRDB utility, third-party libraries, and any copyrighted components that I have no clue about the licensing or whether I should change any of their code at all
* all `do {` [...] while loops ... I never did decide for myself whether I should permanently prefer K&R- or Allman-style braces for do { } while loops, due to linguistics and such.  So I don't know what we want.
* `__asm {`, `try {`, `} __catch {`, `finally {` ... None of these things should even exist in C code, and I am not familiar with the compiler-syntactical linguistics of them or whether it's more accurate to use K&R for them in some special case.  They seem to still be a matter of block scope, but they have no branch prediction or special allocation, similar to my uncertainty with do { ... } while loops.
* Some code was neither in K&R nor Allman, like for example, `if (condition) { break; }`, or, `enum { single_constant = 0; }`.  I don't know whether to change them, so didn't decide.  Personally I wouldn't even use braces at all for the `if (condition) { statement; }` example, as I believe it would be more style-neutral to put the single-line branch to its own line after the "if (condition)" without braces.

I did on the other hand read through and update PJ64 N-Rage, Common, the basic RSP core components, the "Settings" component file....  I also removed excess whitespace and removed/adding some newlines if there were too many or not enough and unrelated code was looking too crammed together.

### DIFF
--- a/Source/Common/Ini File Class.cpp
+++ b/Source/Common/Ini File Class.cpp
@@ -28,13 +28,15 @@ void CIniFileBase::fInsertSpaces ( int Pos, int NoOfSpaces )
 	m_File.Seek(0,CFileBase::end);
 	end = m_File.GetPosition();
 
-	if (NoOfSpaces > 0) {
+	if (NoOfSpaces > 0)
+	{
 		stdstr_f SpaceBuffer(_T("%*c"),NoOfSpaces,' ');
 
 		do {
 			SizeToRead = end - Pos;
 			if (SizeToRead > fIS_MvSize) { SizeToRead = fIS_MvSize; }
-			if (SizeToRead > 0) {
+			if (SizeToRead > 0)
+			{
 				m_File.Seek(SizeToRead * -1,CFileBase::current);
 				WritePos = m_File.GetPosition();
 				memset(Data,0,sizeof(Data));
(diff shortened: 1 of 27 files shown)


project64:
looks good

cxd4:
Meant to have the rest of the components have these Allman changes, but like 4 days ago I let @ExtremeDude2 take over so he can work on it.  Just seemed way easier to let him do it than giving myself another headache.  I'm sure he'd agree considering the time it's taken already. :D

ExtremeDude2:
now the question is, how long will it take me? :o

cxd4:
https://media.pokebattle.com/pokeart/079.png

If he finishes blinking, you took too long.

But seriously, it could take you a month to do this; as long as you're the one doing it and not me I feel better.

Frank-74:
Code::Blocks has an 'AS' format option.  It will parse every source file for Allman-style in one click.

cxd4:
Yeah, but I guess I'm too much of a masochist to rely on automation sometimes.  Maybe I felt better just using my own eyes and looking out for other issues, other possible code changes like too many spaces/missing newlines/whatever besides the braces, so that was why I did that within this pull request as well.

But in the final result yeah, it may be a pretty good idea to use Code::Blocks or whatever feature to automatively check for and fix any remaining cases of non-Allman braces (although tbh most of them I am somewhat inclined to leave as-is).

PR #294, opened 2015-03-20
TLB Unmapping and Protect Memory UI fix.

AmbientMalice:
This should work. I think.


Frank-74:
This makes the UI show Protect Memory checked as default, and TLB Unmapping disabled as default.  So RDB setting Protect Memory toggles the TLB Unmapping option, and vice versa.

Maybe they were switched for a reason.

PR #295, opened 2015-03-20
Cleanup

MELERIX:


project64:
the remove thumbs.db is good, but we should have it in the .gitignore to stop them being added by accident.

MELERIX:
I added in .gitignore too, you can see that the pull have two commits, one where the files are removed and the other where is added in .gitignore.

PR #297, opened 2015-03-21
Fast SP is now SP Hack

MELERIX:


MELERIX:
closing, I'm now sure about this yet, need to be confirmed first.

project64:
it is not really a hack, it is just a caching mechanism for the stack register

MELERIX:
so Fast SP is the right name then, right ?

Issue #263, opened 2015-03-10
Project64 2.2.0.2 (from github) is not working with Socket A CPUs (AMD Athlon XP or AMD Sempron)


MELERIX:
you will see this error window when execute Project64...

![Text](http://i.imgur.com/tU40dMq.png)

project64:
that is with 2.2.0.2 downloaded from the pj64 forum? Or what version?

MELERIX:
yes, with current version 2.2.0.2 from github.

project64:
was a previous version working, did you compile it your self ?

MELERIX:
I have no idea about previous versions, because I just tested now with the computer of a friend in order to check if I can run Project64 in Windows XP, and only then I noticed this bug.

project64:
how did you get the binary file your testing?

MELERIX:
was compiled by me in Visual Studio 2013 Community Edition.

project64:
for xp it needs to be compiled in 2008 or you get that issue

project64:
if you use the one on the pj64 forum that is compiled in visual studio 2008

MELERIX:
I don't see any builds in the forum, may be is an area where I don't have access :(

anyway for me it seems to be more likely a CPU feature being forced in PJ64, because the CPUs I've mentioned above does not have SSE2 instructions, and if you open RCP Main.cpp at line 267 you will see the following code:

void DetectCpuSpecs(void) {
	DWORD Intel_Features = 0;
	DWORD AMD_Features = 0;

	__try {
		_asm {
			/* Intel features */
			mov eax, 1
			cpuid
			mov [Intel_Features], edx

			/* AMD features */
			mov eax, 80000001h
			cpuid
			or [AMD_Features], edx
		}
    } __except ( EXCEPTION_EXECUTE_HANDLER) {
		AMD_Features = Intel_Features = 0;
    }

	if (Intel_Features & 0x02000000) {
		Compiler.mmx2 = TRUE;
		Compiler.sse = TRUE;
	}
	if (Intel_Features & 0x00800000) {
		Compiler.mmx = TRUE;
	}
	if (AMD_Features & 0x40000000) {
		Compiler.mmx2 = TRUE;
	}
	if (Intel_Features & 0x00008000) {
		ConditionalMove = TRUE;
	} else {
		ConditionalMove = FALSE;
	}
}

and probably this is causing the issue.

its SSE2 being forced in PJ64 ?

project64:
[...]

PR #302, opened 2015-03-23
Fixed Windows XP Compatibility with VS2013

MELERIX:
MinimumRequiredVersion is now 5.01, otherwise VS2013 uses default 6.00,
making PJ64 incompatible with Windows XP.

it solve the following issue: https://github.com/project64/project64/issues/263#issuecomment-78684177

also fixed a minor thing in spanish lang file.

### DIFF
--- a/Lang/Spanish.pj.Lang
+++ b/Lang/Spanish.pj.Lang
@@ -231,7 +231,7 @@
 #464#  "Desactivar Protector de Pantalla cuando un rom esté en ejecución"
 #465#  "Mostrar Frecuencia de Cuadros"
 #466#  "Cambiar Tipo de Frecuencia de Cuadros"
-#467#  "Comprobar si project64 ya esta en ejecución"
+#467#  "Comprobar si project64 ya está en ejecución"
 
 //Rom Browser Tab
 #480#  "Máx # de Roms Recordados (Máx 10):"
--- a/PropertySheets/Win32.props
+++ b/PropertySheets/Win32.props
@@ -143,7 +143,7 @@
       <LinkErrorReporting>PromptImmediately</LinkErrorReporting>
       <CLRUnmanagedCodeCheck>false</CLRUnmanagedCodeCheck>
       <SubSystem>Windows</SubSystem>
-      <MinimumRequiredVersion />
+      <MinimumRequiredVersion>5.01</MinimumRequiredVersion>
     </Link>
     <Midl>
       <AdditionalIncludeDirectories>.;$(SRC);$(SRC)\3rd Party;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>


cxd4:
When and if there is 64-bit availability in Project64 releases, the version number for Win64 builds will need to be bumped to 5.02, as it can be nothing less for 64-bit binaries.

MELERIX:
@cxd4 if in a future we have a 64bit version of PJ64 I guess we can make a win64 target, with 5.02 as min, so it will be compatible with XP64 too.

cxd4:
I couldn't help but notice that, when the VS build toolkit is set to vs2013 XP (v120_xp), that the subsystem version implicitly becomes "5.01" without having to force it to that value.  For some reason, the combination of other settings in the VS2013 makefile is preventing this default from being set.  I wonder if that is the reason why these commits have not been merged yet.

It's tricky.  I tried to look at it, but once again merely opening pj64 in VS2013 seems to guarantee the need for a reboot after waiting through 30 minutes of frustrating unresponsiveness.  If this keeps up I may as well just uninstall Visual Studio 2013.  (Actually I was trying to look for VS2002 Express so I could link against msvcrt.dll.)  But the massive collision of so many source files, headers and folders and so much code in this one single repository seems to overflow VS2013's package server to the point where it's totally unusable for me.

MELERIX:
then seems we have troubles with: "PropertySheets", ".vcproj" & ".vcxproj" files, and this is causing others projects to not be compiled with v120_xp properly.

I'm inexperienced with VS2013 so no idea how to fix the whole trouble completely.

maybe "PropertySheets", ".vcproj" & ".vcxproj" files should be generated from scratch, and this could solve the issue completely.

so I'll close this pull, due it seems that does not fix the problem completely.

project64:
what about windows 2000 ?

cxd4:
I don't think that is possible when linking with VS2013 tools @project64 .

Never in any of my VS2013 projects have I been able to set a sub-system version less than 5.01.

Frank-74:
I never understood how microsoft got away with stealing dos and rebranding it as msdos and selling it back to the company they stole it from (IBM).  Their entire fortune was built from that theft.  They tried it before with a dos for Commodore and failed, but succeeded against IBM.  And what with Bill Gates supporting stuff like population control through vaccines and preparing for armageddon with seedbanks on the sea floor near the arctic to repopulate the earth's food supply, in proper James Bond baddy style.  I can't help but find nothing wrong in doing a google search for a licence key for any of their products.

cxd4:
About the same concept as how other majorly 'successful' companies like Google have gotten away with what they do I guess.  Anything with at least a partially hidden motive that goes hand in hand with government rule or convenience--in many cases, why would the government want to do anything about it?

PR #303, opened 2015-03-23
English translattion - Remove the "?" from Settings labels. Reword a few messages and labels.

toehead2001:
The question marks always seemed silly to me. "Graphics" instead of "GFX", as it is more widely understood.


toehead2001:
I do apologize. I made the mistake of running a rebase right after opening the pull request, and it stopped recognizing my commits.

PR #304, opened 2015-03-23
 English translation Clean Up

toehead2001:
The question marks always seemed silly to me. "Graphics" instead of "GFX", as it is more widely understood.

### DIFF
--- a/Lang/English.pj.Lang
+++ b/Lang/English.pj.Lang
@@ -18,7 +18,7 @@
 #12 #  "Extended Rom Info (.RDX)"
 
 //About INI title
-#20 #  "About INI Files"
+#20 #  "About Config Files"
 
 /*********************************************************************************
 * Menu                                                                           *
@@ -71,7 +71,7 @@
 
 //Help Menu
 #180#  "&Help"
-#181#  "About &INI Files"
+#181#  "About Conf&ig Files"
 #182#  "&About Project64"
 #183# "Support &Forum"
 #184# "&Homepage"
@@ -177,7 +177,7 @@
 #320#  "Select current Rom Directory"
 
 //Messages
-#340#  "Bad ROM? Use GoodN64 & check for updated INI"
+#340#  "Bad ROM? Use GoodN64 & check for updated RDB"
 
 /*********************************************************************************
 * Options                                                                        *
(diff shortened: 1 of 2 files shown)


MELERIX:
for consistency you also have to edit the strings in this file: https://github.com/project64/project64/blob/master/Source/Project64/Multilanguage/Language%20Class.cpp

Lang/English.pj.Lang
```diff
@@ -255,8 +255,8 @@
 #503#  "Self Mod Methods"
 #504#  "Default Memory Size:"
 #505#  "Advanced Block Linking:"
-#506#  "Start Emulation when rom is opened?"
-#507#  "Always overwrite default settings with ones from ini?"
+#506#  "Start Emulation when rom is opened"
+#507#  "Always override default settings with ones from ini"
```
MELERIX: maybe the right word here should be "replace" ?

toehead2001:
I think "Overwrite" and "Replace" will make the user think that the settings in the rom ini will become the emulator defaults...  which would be silly
"Override" would make them understand that the emulator defaults will be bypassed and the settings in the ini will be used instead.

You make a valid point about the hyphen. Perhaps it would be simpler to just to use the abbreviated forms (LLE & HLE) everywhere.

Thanks for letting me know about that other file.

toehead2001:
I made some revisions last night. Could I get this reviewed again?

toehead2001:
I just noticed that "ROM" has inconsistent casing. We have "rom", "Rom", and "ROM".

I would think "ROM" and "ROMs" (for plural) would be the correct casing. Correct?

cxd4:
Yes.  "ROM" is an acronym, and with rare exceptions, every letter in the acronym must be uppercase.

Even "rom" looks better than "Rom" ... really no reason to only uppercase the "read" part of the acronym.

oddMLan:
>FireFox

Not meant to derail the main discussion, but it's actually spelled _Fire**f**ox_ :P

toehead2001:
Another question.
We currently have "Instant Save", "Save State", and "State Save"

Do we want to go with "Save State" ? and if so, "Load State" or "Load Save State" ?

Personal, I prefer "Load State" and "Save State".

MELERIX:
in others emulators such as: Snes9x & VBA-M it says "Load Game" & "Save Game".

but, technically we are not saving & loading using the game, we are saving & loading using a CPU state, which is practically a "state" of the CPU instructions in a specific moment.

so technically "Save State" & "Load State" should be the right words, I guess.

it require a debate xD

cxd4:
Yeah, prefer as-is.  I usually just call them "save**d** states", not "save states".  As in the state of the machine...I'd hardly think of "save states" as a noun.  So actually I'd prefer the current way pj64 names that.

toehead2001:
>So actually I'd prefer the current way pj64 names that.

PJ64 currently has three ways of naming them... hence my question.

PR #309, opened 2015-03-23
minor clean-ups to Glide64 config

cxd4:
The first two commits were to fix the type overflow from passing long integers as `unsigned int`.  If you pass `0L` as a function argument of type `(unsigned int)`, it just looks stupid because you shouldn't have to convert from 0 to a signed long, and again to an unsigned int.  Furthermore only some lines of Glide64 config were doing this which make it look inconsistent and ugly.  Furthermore it's actually `0l` instead of `0L`, where the lowercase L looks like a 1 which made it even harder to read.

So that is the explanation for the first two commits.

The last two commits...I thought it was tedious to keep saying the longer, lower-level `RegisterSetting` function (introduced here:  https://github.com/project64/project64/commit/1c65f79e9017ec11a5b34ad648ee2abefb4bc5d4)...
```c
RegisterSetting(x, Data_DWORD_General, y, NULL, z, NULL);
```
...when only 3 of the Glide64 settings updating were relevant:
```c
general_setting(x, y, z); // Data_DWORD_General
game_setting(x, y, z); // Data_DWORD_Game
```

Take advantage of C functions to create higher-level functions over the Project64-API-specific `RegisterSetting` for the following advantages:
* shorter code and more readable to people who are just trying to understand the Glide64 settings, their values, what to set them to and when without the cruft of external API interference such as the Project64-specific `RegisterSetting` API function and many of its unused/NULL parameters
* smaller DLL file size, assuming MSVC will not do in-line expansion on the higher level function
* frees up instruction cache space for performance-relevant sections of Glide64 by minimizing the size of the performance-**ir**relevant configuration functions by calling shorter, higher-level functions instead of longer, lower-level functions
* shorter lines of code that no longer exceed 80 characters in length or sail past the end of your monitor screen :)
* portability of Glide64 as a graphics plugin outside of Project64- or N64-emulation-fixed environments

Probably macro functions would have also worked, except with macro functions, you cannot prevent in-line expansion from happening...which is undesirable here as we have no interest to in-line larger, more sizable functions for aspects of the Glide64 code where performance really is not at all a problem or relevant.


cxd4:
Added a 5th commit to fix compiler error regressions caused by my 4th commit (which was incomplete and forgot to update the number of arguments passed to the new function, which should be 3 not 6).

I had compiler errors in the 4th commit because when I tested I didn't have compiler errors, but Visual Studio kept locking up my system for a full 30 minutes, forcing me to clean re-build and complaining about old linker files from VS2013 conflicting with VS2008 etc. etc. etc. long story short I really tried to _repeatedly_ test the build to make sure I didn't overlook anything in my changes, but at the last minute I had made those errors out of frustration, sorry.

Source/Glide64/Config.cpp
```diff
@@ -1310,3 +1310,28 @@ void CALL DllAbout ( HWND hParent )
...
+{
+    RegisterSetting(
+        setting_ID,
+        Data_DWORD_General,
+        name,
+        NULL,
+        value,
+        NULL
+    );
+    return;
```
project64: any particular reason your calling return here?

cxd4: AFAIK every function has a return point.
The way I understand it, the `return` keyword is optional only in void functions that don't return an actual thing, but still need to leave the function.

I guess it depends on whether you view C as a low-level or a high-level language, but personally I think it's better to be explicit and formalize when you know it's time to be done with the function.  Although, I can't say it's any more or less maintainable to just remove the `return;`, so I think it's fine if you want to remove that.

I see it more as a matter of the C language permitting the omission of the `return;` in void functions, rather than creating it as an optional thing just if you wanted it, but I guess I could be wrong on that.

cxd4: haven't been doing it in every pull request though...maybe in this particular commit it triggered me to do it because of this:
```c
void general_setting(short setting_ID, const char * name, unsigned int value)
{
    RegisterSetting(
        ... x ...
```
... instead of ...
```c
void general_setting(short setting_ID, const char * name, unsigned int value)
{
    RegisterSetting(... x ...);
```
which caused me to remember my habit in this PR.  IDK, doesn't really matter, I'd have taken it back out if you wanted me to I guess.

cxd4:
Not to drag a dead horse but I have the urge to extend on my clarification.
A minor thing that just came to me after reading a lot of AziAudio that I need to make clear.

>any particular reason your calling return here?

I'd just like to emphasize that `return` is never "called".  It's not possible for anyone to be "calling" return because it's a keyword for an operator, not a function.

That's why I always insist on having a space between the keyword and the expression.
```c
if (expression) // dynamic operator of an expression
if(expression) // misleading, can confuse "if" for a function, which it isn't
function(parameter) // function of a parameter
function (parameter) // potentially misleading, but nonstandard
```
Because `return` is already required for all C functions to terminate, I do not personally believe that return can be considered as a function...if `return()` really was a function then it would be like a very abstract void function with no algebraic potential whatsoever, unlike `sizeof()`.  So I think it's a required keyword so far as C language functions are all concerned (except maybe void funcs), but not a function itself--more of a construction for low-level languages to create true mathematical functions.

PR #311, opened 2015-03-24
Indy Racing 2000 Work Real Good.

AmbientMalice:
Disabled register caching and disabled framebuffer for Glide64. This game runs terribly, and doesn't work at all with Interpreter. But at least it "works" again, like it did with 1.6


dsx-:
game works with register caching enabled if fixed audio timing is enabled as well...weird problem that should be looked into

PR #321, opened 2015-03-25
Fixed Brazilian Language File Codification

Lithium64:
Project64 only loads utf8 bom language files


MELERIX:
you should use the Standar UTF-8.

PR #322, opened 2015-03-25
English language fixes - Upper case "ROM", consistent terminology for saved States, spelling

toehead2001:


toehead2001:
>I'm not really sure whether "mempak" is a proper noun which would get upper-cased

Same thing would apply to "flashRAM", but I think it's best to have all items in that list capped... for consistency if nothing else.

cxd4:
Nintendo doesn't call it "flashRAM"; they call it just "flash [memory]" or "flash ROM", with the space.

It's only "mempak" where they mix in the slang, so that comparison wouldn't apply.
And just because zilmar capitalized all of them consistently doesn't mean that any of them should be capitalized.  Only proper nouns, the first letter of a *complete* sentence, title case etc. involve capitalizing the first letter.  Probably the reason why he wrote "Flashram" instead of "flash RAM" is because he's used to ThisStupidCamelNotationOfNamingVariablesInC.  But that doesn't make it correct English.

PR #331, opened 2015-03-27
attempting to synchronize [most?] language class changes to language files

cxd4:
Most of these changes are only to the English language files, as I know no other language but English (well, and C).  I don't know Spanish or anything else, so I apologize in advance if I've made any specially wrong assumptions in my attempts to synchronize my changes to the English language file to some others such as the Japanese or Spanish language file.

For example, I just figure that "Nintendo64" is not a valid word--you're supposed to have a space between the company name and the 64 because that's what Nintendo calls it.  I don't know if some non-English languages have some special rule that bans spaces or whatever, though I doubt it.  I just figure 95% of the language files kept saying "Nintendo64" without the space, because they all copied from Smiff's or Tooie's English language files.

I also tried to match some of my lower-case rules (unnecessary capitalization "fixes") to the other language files to try to keep them somewhat up-to-date.  Again I'm sorry if there was some sort of fundamental exception I had to overlook out of lack of special knowledge of the language.

But in general how do the changes look?


cxd4:
There, did the commit.  Triple-checked the rest with it and haven't found any mistakes of mine.

Though that probably means I made some anyway right? :P Does anyone know what they are?

MELERIX:
the changes in spanish file are ok at least.

cxd4:
Sounds good to hear.

I've went ahead to merge conflicts resulting from the other language pull request at https://github.com/project64/project64/pull/332.

PR #334, opened 2015-03-27
Change Glide64 Texture Filter Hotkey.

AmbientMalice:
Previously, pressing Backspace will change the texture filtering settings for Glide64.

With this commit, ALT+Backspace will change texture filter settings, preventing unintended switches for whatever reason - including but not limited to casually writing github posts while a game is running in the background.

### DIFF
--- a/Source/Glide64/Main.cpp
+++ b/Source/Glide64/Main.cpp
@@ -2033,15 +2033,7 @@ void newSwapBuffers()
         output (930.0f, 0, 1, (char*)wxDateTime::Now().Format(wxT("%I:%M:%S %p")).char_str(), 0);
       }
     }
-    //hotkeys
-    if (CheckKeyPressed(G64_VK_BACK, 0x0001))
-    {
-      hotkey_info.hk_filtering = 100;
-      if (settings.filtering < 2)
-        settings.filtering++;
-      else
-        settings.filtering = 0;
-    }
+     //hotkeys
     if ((abs((int)(frame_count - curframe)) > 3 ) && CheckKeyPressed(G64_VK_ALT, 0x8000))  //alt +
     {
       if (CheckKeyPressed(G64_VK_B, 0x8000))  //b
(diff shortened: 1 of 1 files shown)


cxd4:
Looks functional, though I know nothing of windows API functions.

I always did hate it at times where I couldn't type in a separate process with the emulation window idle or somewhere else in the background and receiving the window messages, since I use my keyboard for emulation as well.

The only concern here I could think of is--how did people know that Backspace was the keyboard shortcut formerly for changing the Glide64 texture filter?  If they had already read from some manual that this was how to change the filter, maybe now they will have a hard time discovering that it was changed to Alt+Backspace unless that was documented somewhere?

AmbientMalice:
Perhaps something could be added to the "About" screen, or new Glide64 documentation could be written?

cxd4:
I don't know, it would depend how people already knew what the old shortcut (plain Backspace) was.

If the original Glide64 had no documentation so far as this goes, then I would feel weird adding documentation for just this version of it.  If there was documentation, people would have read it to see that Backspace was the shortcut, then assume there was no reason to change this with PJ64 Glide64 and get confused to see that hitting just plain Backspace no longer works.

Do you have the problem with hitting Backspace in GitHub / web browser changing the PJ64 Glide64 texture filtering in the background with the original Glide64 as well, or just PJ64 fork of Glide64?  Because that does sound very annoying.  If it all possible it might be best solved to just make keystrokes sent to the active window not affect Project64 when it's inactive, but I don't know if people testing original Glide64 plugin would have had this issue only with pj64, mupen64 or if it can even be helped at all.  Because if it can't be helped, them maybe the only way would be to let plugins register the kotkeys through the specifications.

AmbientMalice:
Glide64 came with a chm help file that had the hotkeys alt+b, alt+v, and backspace documented in it.

I'm not even sure the average user should be tampering with texture filter settings. They're already locked behind the "advanced" menu options for the plugin. alt+v is a useful hotkey for users who temporarily need per-frame framebuffer for a game, though.

I'm pretty sure this has always been an issue. I remember playing Winback on 1.6 with Glide64 final and having the texture filtering changing because I paused the game and wrote some notes in a text document.

MELERIX:
I generally use the backspace key as "Z" buttom, so this change is really useful.

PR #342, opened 2015-03-29
Convert more files to allman

ExtremeDude2:
Includes Source\Project64\N64 System\C Core, Source\Project64\N64 System\Debugger, Source\Project64\N64 System\Interpreter, Source\Project64\N64 System\Mips, and Source\Project64\Multilanguage


cxd4:
yay!  u can haz win

ExtremeDude2:
nu uh, it isn't even close to all of it XD

PR #355, opened 2015-04-04
Backport Felipefpl's changes to Brazilian Portuguese language file.

cxd4:
The file he uploaded here seems to be based on the latest changes to the file by Lithium64.
http://ge.tt/api/1/files/7pgKalD2/0/blob?download

There were a few "reversions" to Lithium's changes, as well as to some of my changes (replaced "ROM" and "ROMs" with "Rom" and "roms" in a few places).  I will commit a few reversals to these [hopefully unintended] modifications shortly.


cxd4:
@Lithium64 what do you think of these changes?

I know a few of his changes somewhat overwrite stuff that you did when changing the language file, not to mention the "Author" bit.  I don't know any foreign languages so would need to run this by someone qualified.

Lithium64:
@cxd4 
I agree with a few changes

Some literal translations does not make sense

cxd4:
If you can, tell me what changes you disagree with so that I can update this pull request before it ends up getting merged.

Or would you rather we just have this pull request merged first, then you can update the bits you disagree with on your own?

Lithium64:
@cxd4 I will update the file and make a new PR

cxd4:
https://github.com/project64/project64/pull/357

PR #356, opened 2015-04-05
Line ending and File-encoding update.

MelchiorGaspar:
PLEASE I hope i got this Branch HELL right this time ;^_^;


cxd4:
You did it fine.  Don't worry anymore; you followed my instructions closely enough.

project64:
I have no issue with this, but what effect does this actually have?

MelchiorGaspar:
just cleaning up I guess they had the old file encoding
the .props files had LF Linux line ending


and the .vsprops had these file encoding  both for the xml header and the file it self...
UTF-8 is the current good standard...  
Windows-1252 is ancient, Latin file encoding and
Windows-1251, is Cyrillic script such as Russian, Bulgarian, Serbian Cyrillic and other languages.
 
.

MelchiorGaspar:
Consistent file encoding.

PR #363, opened 2015-04-10
Re-Volt CF=1 framerate improvement.

AmbientMalice:
I had avoided this in the past since it seemed unstable. But with @Frank-74's changes, the game now seems stable with CF=1, and it runs at 60fps with occasional dips instead of all over the place with dips into the teens as it does with CF=2.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -4744,6 +4744,7 @@ Plugin Note=[video] slow in options menu
 RDRAM Size=8
 ViRefresh=1500
 AiCountPerBytes=530
+Counter Factor=1
 Fixed Audio=1
 Sync Audio=0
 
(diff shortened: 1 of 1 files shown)


dsx-:
idk how i feel about smoother framerates, cf=2 is closer to actual n64 therefore more accurate but thats just me, no big deal i guess

AmbientMalice:
I take a moderate view on the issue since IMO, PJ64 is primarily about usability. Nobody is going to praise PJ64 for accurately emulating the framerate of a game like Re-Volt or Turok 2. Re-Volt was hammered by critics for its framerate on N64. If PJ64 can have smoother framerates than a real N64 with no side effects, then that's a near-universally good thing, IMO.

However, I do take exception to any sort of timing fiddling that results in significantly wrong game behavior. CF=1 causes actual bugs with Perfect Dark, which is why I find "60fps" hacks for that game currently unacceptable.

It's ultimately Zilmar's call, though. I'm just trying to do my bit to help PJ64 strike a balance between accuracy and end-user enjoyment.

Issue #367, opened 2015-04-11
Sin & Punishment (J) [T] - missing settings in Glide64.rdb


baptiste0602:
Could you please add settings for Sin & Punishment (J) [T] ? As you did in Project64.rdb.

cxd4:
Made a commit for it to probably end up merged.
https://github.com/cxd4/project64/commit/9d7303483787f53e9b3c2a91f6ed0ce1cb54b5d8

If you try this updated RDB, is your issue fixed?
https://raw.githubusercontent.com/cxd4/project64/yo_glide_i_herd_u_liek_translated_games/Config/Glide64.rdb

baptiste0602:
Sure, i use these seetings. Thanks for the commit.

cxd4:
Seems to have taken a while sorry, but it's merged now:  https://github.com/project64/project64/commit/b54e03b85d2474b00d3415a6d7b069319676bcf9

I'm guessing if that was done right then it satisfies this issue, which can be closed then.

baptiste0602:
Thanks and no problem for the delay, you do that on your free time :)

PR #368, opened 2015-04-11
copy Sin and Punishment RDB settings to English trans.

cxd4:
Have a complaint [here](https://github.com/project64/project64/issues/367) that `Project64.rdb` supports both the Japanese release as well as the English translation ROM to _Sin and Punishment_, while `Glide64.rdb` supports only the non-translated, original ROM.  So I used `Project64.rdb` to copy the settings into a new entry for the translated one.

### DIFF
--- a/Config/Glide64.rdb
+++ b/Config/Glide64.rdb
@@ -2965,8 +2965,6 @@ Good Name=Shadow Man (U)
 Internal Name=Shadowman
 depthmode=0
 
-// added 2015.04.11:  "Tsumi to Batsu:  Hoshi no Keishousha" English translation
-// [B6BC0FB0-E3812198-C:4A]
 [B73AB6F6-296267DD-C:45]
 Good Name=Sin & Punishment (J) [T]
 Internal Name=Sin and Punishment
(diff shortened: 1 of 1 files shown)


Config/Glide64.rdb
```diff
@@ -2965,6 +2965,15 @@ Good Name=Shadow Man (U)
 Internal Name=Shadowman
 depthmode=0
 
+// added 2015.04.11:  "Tsumi to Batsu:  Hoshi no Keishousha" English translation
+// [B6BC0FB0-E3812198-C:4A]
```
project64: do we need the comments?

cxd4: No.

I put them there just to help prevent future commits from changing the settings for only one version and not the other, since this would be one of the first--if not the only--new entry in the Glide64 RDB that wasn't in the official Glide64 INI.  I was just trying to prevent confusion; that was all.

I'll happily remove it anyway though, since it does sort of look too "striking" from the rest.

cxd4:
There seems to be removed.

PR #370, opened 2015-04-12
Fix losing audio while loading a save state

LegendOfDragoon:
This fixes the losing audio after loading a save state issue with
certain plugins in certain games. I was hesitant to submit this, since
the fault is the audio plugin itself, but this also fixes bad save
states. This will not fix games like Top Gear Rally.

Any plugin that initially had this problem, can still lose audio at any
given time, due to flaws in the audio plugin itself. It is best to
simply use a plugin that doesn't have this issue to begin with, for such
games.

### DIFF
--- a/Source/Project64/N64 System/N64 Class.cpp
+++ b/Source/Project64/N64 System/N64 Class.cpp
@@ -1571,7 +1571,7 @@ bool CN64System::LoadState(void)
 bool CN64System::LoadState(LPCSTR FileName) 
 {
 	DWORD dwRead, Value,SaveRDRAMSize, NextVITimer = 0;
-	bool LoadedZipFile = false;
+	bool LoadedZipFile = false, AudioResetOnLoad;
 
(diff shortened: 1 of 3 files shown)


AmbientMalice:
Assuming this works, great job. I'll get testing later.

project64:
I would prefer it a game specific setting then applying it to all games, so just make in a rdb setting, it does not need a UI element for it.

LegendOfDragoon:
Good point actually. Are you going to implement it, or should I do it?

AmbientMalice:
@Project64 Worth noting that working games can be randomly affected by the sound cutting out after save states. It happened during my Conker playthrough recently. Unfortunately, while this setting MIGHT fix those rare cases, it's not something that can be tested easily - if at all. I'm not sure how liberally you want this setting applied. If you want, I'll do a run through my game list to identify the roms that this demonstrably fixes, and make a pull request to add this new rdb option, and then any fringe cases can be looked into later.

PR #379, opened 2015-04-14
Fix cheat DB "Compleate" typo.

AmbientMalice:


AmbientMalice:
An old commit to disable FB is included because I'm dumb and don't know how to remove said commit without breaking stuff. Has no effect because that change was made ages ago.

PR #388, opened 2015-04-16
Harvest Moon 64 Glide64 fix.

AmbientMalice:
Disabled FB to fix half of screen being cut off. Kinda redundant with GLideN64 coming very soon, but I don't like broken things.

(No idea if disabling FB breaks anything, though.)


MELERIX:
is not supposed that these settings should be in Glide64.rdb ?

AmbientMalice:
You are absolutely right, @Melerix. I'm such a dunce.

PR #391, opened 2015-04-16
Correct Harvest Moon Settings.

AmbientMalice:
Thanks to @MELERIX for pointing out that I put fb_smart in PJ64's db instead of Glide's.


MELERIX:
there is more to move, see the changeset: https://github.com/project64/project64/commit/b5d6f3d6572bce1cd46f22a34c8e53d77763c903

AmbientMalice:
Ah, I think I missed (J) 1.0 I'm unfortunately away from my PC. So I'll have to fix it tomorrow morning.

PR #399, opened 2015-04-17
some changes for NRage default keyboard keys

MELERIX:
default keyboard buttons sorted in a similar way to N64 game pad, and analog stick control now uses WASD keys (which is the standard position for most of the games and game pads).


AmbientMalice:
Analogue on WASD? Personally, that's a terrible idea. What is the actual button layout now? I personally use the current settings, but change them slightly so A is on the C key, B on the X key, and Z on the Z key. That way ABZ are easy to reach, and you can ALSO easily operate the C-buttons via WASD by simply lifting your fingers. In my humble view, that's the best possible scheme for handling N64 on KB.

I feel movement should ideally be handled by the right hand and interaction and camera by the left hand. I mean no offense, however. Different people find different things comfortable.

MELERIX:
analogue is on the left side WASD (like in N64 and most of the game pads) and digital is in the right side with the arrows.

[Z] is in spacebar as always, (B) is in T an (A) is on G, c buttons are in U,H,Y,J.

AmbientMalice:
If you must put the C buttons on the right hand side, surely IJKL is more traditional? Other emulators have used those buttons for analogue movement. TBH, T and G for A and B does not seem intuitive at all. With this layout, it is extremely awkward to hold down the A button while moving. If you want to move and also rotate a camera, you have to move your hands ot stretch really far. When and B are located around and ZXC area, one can easily hold one down with an index finger while operating WASD with the ring and middle fingers.

Also, where does R and L go? Those were E and Q before, complimenting the WASD C-buttons. This worked reasonably well for N64 FPS games that tend to favor C-buttons for movement and analogue for looking. Or, for example, power sliding in Mario Kart. Hold down E\R while pressing A\X or A\C depending on how that was tweaked to accelerate.

MELERIX:
L and R are in I and K.

seems you play with inverted hands ?

generally in game pads or even FPS games you always move with left hand and right hand is for action buttons.

AmbientMalice:
>[B][CL][CU]
>[A][CD][CR]

This treats the N64 buttons as square, when in fact they're arranged as a diamond. If the player wants to press C-Up, they will intuitively press a button corresponding to an upwards direction. W, I, Up Arrow, etc. If they want to press C-Left, they will intuitively press a button which lies DOWN and to the LEFT. The problem with your arrangement is it destroys the natural 1:1 analogies for N64 controls. This is, in my view, a huge design mistake - one I see in emulators such as desmume and PPSSPP, where the diamond of ABXY or CIRCLE+SQUARE+TRIANGLE+CROSS is arbitrarily placed on a square grid as ASZX for no apparent reason when WASD is far more logical since it reflects the original diamond layout. By placing those buttons on WASD, the Q and E buttons form elegant and natural should buttons which can be held without hand strain.

MELERIX:
what do you think about using UHJK as C Buttons, for example:
 [T].... [U]
 [G][H][J][K]

AmbientMalice:
what do you think about using UHJK as C Buttons, for example:
[T].... [U]
[G][H][J][K]

I... suppose? I do think it is crucial that the R button be to the RIGHT of the diamond, which it would be if it were *I*. However, I think that shoulder buttons should always be arranged horizontally, so that would put L on Y?

You could shift it to the right and have

   I
J-K-L

for C-Buttons and U and O for L and R shoulder buttons. But I dunno...

MELERIX:
no idea where to put L and R buttons, may be on [Y] and [I] ?

AmbientMalice:
I don't want to come across as a whiny person trying to force my opinion onto you. You have just as much right to advocate for your layout sensibility as anyone. (Although I would like you to think about the triangle\diamond factor.)

I'd personally prefer PJ64 to offer a series of presets when PJ64 first boots, plus accessable from N-Rage.

However, please allow me to illustrate the underlying logic beneath the layout I favor.

n64_controller_ambient

First, please note the controller layout.

ambient_keys

Now take a look at the keyboard. Notice how all the important functions are arranged in a neat grid of 9 keys centered around the C-buttons. Notice how you can operate the shoulder button R with your middle finger while pressing A (C), B (X), or Z (Z) with your other fingers of your left hand. Same goes for the C-buttons.

This layout is the most usable "all-round" N64 emulator layout, in my somewhat biased view. It can be used for any genre of N64 game with relative ease. I'm not DEMANDING it be hailed as the greatest thing ever. I merely ask that you consider its logic.

MELERIX:
the trouble with that order, is that is inverted for the common gaming standard, practically you've put movement in the right hand and and actions in left hand, while the standard is left hand for movement and right hand for actions or mouse.

AmbientMalice:
>but the standard (in game pads and pc games) is to use left hand for movement and right hand for actions or mouse.

The N64's games themselves were not consistent when it came to which hand should operate movement. Games like Jet Force gemini switch between right hand control and left hand control depending on whether you're aiming a gun.

Also, traditionally, PC games used the arrow keys for movement and various keys on the left hand side of the KB for interaction. When you're playing a racing game on PC, not that I like the genre, it is traditional that arrow keys handle steering and various keys on the left handle interaction. The logic behind switching movement to the LEFT for PCs is simply because PC games started supporting mouse aim. Traditionally, movement was controlled by the right hand.

The N64 is not like other consoles because its games do not follow consistent logic where the left hand side handles movement and the right hand side handles interaction. This inconsistency is why designing a layout which works for every N64 genre is rather tricky.

AmbientMalice:
For the record, when I play FPS games, I tend to use a mouse-driven layout like this one.

fps_ambient

Remember that most significant N64 FPS games support or default to C buttons for movement. So having them on WASD already lines up perfectly. And A and B sit very nicely on E and R. Since Rare games use B for reloading weapons, it feels very organic to have B assigned to R in this context.

Basically, the common factor is keeping the necessary keys tightly contained and preferably aligned around a triangle of keys such as WASD. By having plenty of space in the middle of the keyboard, your hands aren't gonna be cramped.

MELERIX:
in game pads from: Nintendo, Sony or Sega (since 90s up to current days), the left hand was always used for movement and right hand for actions, and in others game pads as Xbox it has the same behavior, also in PC with FPS games or in Adventure & Action games like: GTA, Assassins Creed, and many others, have the same behavior.

thats why I consider this as a standard, and due n64 game pad uses the same standard, it makes sense to make a similar behavior in keyboard.

AmbientMalice:
>in game pads from: Nintendo, Sony or Sega (since 90s up to current days), the left hand was always used for movement and right hand for actions

N64 developers didn't abide by this, though. That's why movement using the right hand occurs in several important games. If you boot Turok 1, 2, Rage Wars, or 3, all those games are using the C-buttons for movement by default. The favored control schemes for GE\PD involve C-button movement. Conker and JFG, as a mentioned, swap movement between stick and C-buttons depending on context.

That said, there's nothing wrong with swapping C-Buttons from WASD to Arrow Keys. I just think that people picking up the emulator are going to logically expect the arrow keys to navigate menus. And for a vast majority of N64 games, the analogue stick is used for menus, not the C-buttons.

cxd4:
I never did like using Z, X, C keys for any n64 buttons...just felt dumb.  I have my fingers on the home row for easy ASDF/JKL; access; why should I copy off of some PSX emulator's config or w/e this is?

The analog stick is its own special non-digital thing with the N64 controller which requires more intrinsic support for handling abstract cases, like pressing Left and Right on the control stick at the same time which you can only do on the keyboard, plus some keyboards don't allow you to press certain keys simultaneously if you want to move northwest or something.  For all of those reasons, I always preferred to use the arrow keys, not WASD, for the analog stick directions.

That however was in the past.  Nowadays I'm much more suited to WASD for analog and arrow keys for digital pad, but only because I changed my preference over all the rest of the buttons.

Before I was actually pretty satisifed with:
```
START button:  space bar
A button:  'A' key
B button:  'S' key
Z-trigger:  'X' key (so X+A would be Z+A in a game, felt easy to me at least...)
L-trigger:  'Q' key
R-trigger:  'R' key (GRRRR look at me I'm a crab! ^_^)
C-buttons:  NUMPAD 2, 4, 6, 8
analog stick:  arrow keys
```
Pretty sure it wasn't necessarily the best setup, especially since the C-buttons when ever I so rarely may have needed to use them in combination with some other buttons, were all the way on the opposite side of the whole damn keyboard.

The nice thing about getting used to my new setup is that I have a much greater probability of only needing to use one hand instead of both hands:
```
START button:  'E' key
A button:  'J' key
B button:  'K' key
Z-trigger:  'O' key (so O+J would be Z+A in a game, felt easy to me at least...)
L-trigger:  'U' key
R-trigger:  'I' key (GRRRR look at me I'm a crab! ^_^)
C-buttons:  'L', 'P', '"', and ':'
analog stick:  WASD
```

AmbientMalice:
>I never did like using Z, X, C keys for n64 buttons...just felt dumb. I have my fingers on the home row for easy ASDF/JKL; access; why should I copy off of some PSX emulator's config or w/e this is?

Because Z is Z. That's the underlying psychology of it for me. Nothing to do with PS1 emulators. The Z-trigger has no equivalent on the keyboard except the humble Z key. And what do you know? It's located right next to X and C, two classic interaction buttons for PC games, and right below WASD and Q/E, which are also important function buttons on PC. That zone of 9 keys has been important in PC gaming gaming for ages. And in my view, it forms a perfect anchor.

If Z=Z, then it feels very natural that B and A should be located next to it in horizontal sequence, in my view.

Anyway, on another note, WASD and IJKL are the traditional movements buttons on a keyboard face. I don't think that PL:" is particularly intuitive. The I key is interesting because it resembles an Up Arrow. This forms an assocation with "Up", I feel. Also, J and L vaguely resemble angles pointing left and right.

cxd4:
Z=Z reminds me of the logic I used when I first started using keyboard for N64:
'A'=A button
'B'=B button

And guess how that worked out in terms of comfort? ;)

Z=Z and A=A is very uncomfortable with the Z+A button combo in games.  Try holding Z and pressing A; it's awful.  Now, 'X'=Z trigger isn't so bad.  'X'+A is manageable, and X is only 2 letters before Z in the English alphabet (not to mention right next to the 'Z' key anyway on the QWERTY layout).  So, if I had to pick and choose between 'Z'='Z' and 'A'='A', uh, I'd take the latter thx.

Anyway I said PSX because everyone always seems to use X for A and C for B or Z all the time.  You don't fall under that category at all; it's just some weird stupid widespread config I see everyone using like it's unique to them, and it's some strange default most controller plugins seem to be using.

>Anyway, on another note, WASD and IJKL are the traditional movements buttons on a keyboard face. I don't think that PL:" is particularly intuitive.

IJKL is more widely known than PL;', yeah, but it prevents WASD because you have the non-important, not-near-as-often-used C-buttons assigned to keys in the center of your keyboard's layout...if you had Z, L or R triggers assigned to keys to the right of IJKL then it just feels stupid because you're using your hands in the center of your keyboard for the C-buttons and all the way at the distant edges for the A and B or more important buttons.  Plus IJKL makes 'O'=Z too weird...it's better to have the mass of all important buttons centralized, and the C-buttons don't really count so, I'm pretty sure it's far from impractical to replace the old "traditional" notion of IJKL in the n64's case.  We do have 3 directional conflicts after all:  control stick, gray digital joypad and the 4 C-buttons.

AmbientMalice:
The N64 consists of two directional inputs. C-buttons and stick. (D-PAD rarely counts.) WASD, IJKL, and arrow keys form the three triangles which correspond to D-pad, stick, C-buttons in any order.

The shoulder buttons, in my view, should directly above left and right inputs. Q or E if using WASD, or U and O if using IJKL.

My layout is based around keeping 99% of the N64's functions located on a simple grid of 9 keys. Z(Z), X(B), C(A) mean that you can operate the C-buttons, press A, B, Z, and also operate the very important R shoulder button with one hand. Your right hand is either placed on the arrow keys (analogue) or IJKL (for those rare D-Pad games.)

Place your three middle fingers on QWE. Every single important function is now sitting under your left palm with my layout.

ZXAS is a nonsensical button layout used by many emulators that ignores the diamond layout of the buttons involved. WASD is a more logical layour for ABXY/Sony Face Buttons, in my view. It also puts the shoulder buttons of SNES/PSP within easy reach (Q/E) -- However, this doesn't work for consoles with four shoulder buttons.

cxd4:
DPAD rarely counts, but that's not a relevant argument because it still has to be mapped to something at all which preferably is comfortable when you need it, so for the purpose of a complete and playable setup on the keyboard there are basically 3 directional inputs in this challenge instead of just 2.  Since the arrow keys are the most distant from all of the other useful keys on the keyboard for normal play, I've since resigned to using them for the DPAD instead of using them for the analog stick.  Though either one is more than feasible.

>The shoulder buttons, in my view, should directly above left and right inputs. Q or E if using WASD, or U and O if using IJKL.

Well the shoulder buttons really have nothing to do with most analog stick combos, so it doesn't really make any sense to necessitate placing them near analog stick left/right keys.  So WASD with 'U' and 'I' as the L and R button works perfectly without conflict.

And, your use of the N64 analog stick is left-handed, not right-handed.  You don't need to force the eastern buttons of the controller to fall within the same access as your left hand on the left side of the keyboard.  So if you were going to use WASD to begin with for the analog stick, it's kinda inflexible to collage all the other buttons nearby it.

>ZXAS is a nonsensical button layout used by many emulators that ignores the diamond layout of the buttons involved.

Diamonds don't make my keyboarding any more comfortable.
I prefer cushions. :)  They cut less.

If all that really matters is just matching the respective N64 controller button layouts, then we may as well use my old config where I used the NUMPAD keys to the far right for the C-buttons.  Just because it has a "diamond layout" doesn't make it any more comfortable, unless it is a directional connection between the buttons, which is not the case with the A and B buttons (you were talking about ZXAS, which is not related to analog sticks).  ASDF row is better for the A and B buttons since A and S are right next to each other, and X is closer to the center of the keyboard than Z is which is all the way out towards the lower-left corner of the keyboard which shouldn't need to unnecessarily magnify the geometric radius you have to stretch your fingers over.  'Z' as in English keyboarding is rarely used letter; I agree with it being a corner key and not a center one.

MELERIX:
the current setting I've made is in this way...

....[W]........[T][Y][U][I]
[A][S][D]....[G][H][J][K]....[ENTER]
............[SPACEBAR]....................[ARROWS]

WASD = Analog Stick
T = B Button
G = A Button
UHJK = C Buttons
Y = L Button
I = R Button
ENTER = Start Button
SPACEBAR = Z button
ARROWS = Digital Pad

AmbientMalice:
I feel the analogue stick should be on the right side, not the left, since this is a traditional standard going back decades now. I don't believe in trying to replicate the N64 layout on a keyboard. I believe in taking the FUNCTIONS involved and laying out in a manner which offers maximum flexibility. I also believe the right hand should only be handling movement. Wanna use analogue? Arrows. Wanna use D-pad? Shift your hand to IJKL. All the crucial functions should be reachable with the left hand. Games that use analogue for aiming don't cause that much of a problem. Just aim with the arrow keys, move with WASD. Fire by hitting Z. This is a smidge awkward, granted. There is in fact some merit to using SPACE for Z. In fact, even if I kinda prefer Z=Z, SPACE=Z has less hand strain when constantly using WASD. HOWEVER, Z=Z is much better when playing a platformer like Mario 64, where you only have to press ABZ and C-Left/C-Right for camera.

So:

WASD=C-Buttons.
QE=Shoulder.
X=B
C=A
SPACE/Z=Z (Maybe support multiple buttons for the same input?)
IJKL=D-PAD.
Arrows=Analogue.

The diamond/triangle layout is not just about comfort. It is kind of important that your C-buttons have four correct directions. The whole reason ABXY assigned to ASZX is nonsense is because it completely destroys games that expect you to move forward by pressing Y, left by pressing X, and so on. You must always remember that diamond face buttons can and will be used as directional inputs by game devs.

AmbientMalice:
@MELERIX I'm not convinced arranging A and B arranged vertically on the keyboard is wise. There's a reason FPS games use F for "use" and G to toss grenades (generally), as an example. Or Z, X, C to switch from prone, crouch, stand. Keyboards are horizontally oriented.

The rest ain't bad, though. I do think analogue should use IJKL simply because it LOOKS like arrow keys. It's a good feel. (Yes, and it's traditional.)

AmbientMalice:
It occurs to me that this is fundamentally a dispute over left-right alignment. I feel the right hand should only handle movement or aiming. Your models spread stuff like shoulder buttons to the right hand. Maybe this has some "spreading the load" merit. I'm not sure.

I still think arrow keys should be movement and IJLK (Or UHJK) should be D-pad. The N64 is an analogue-oriented console. The arrow keys are the NATIVE directional inputs for a PC, going back forever.

AmbientMalice:
The reason I put A and B as C and X is because in a game like Mario 64 or Banjo or whatever, you need to be able to easily switch between jumping, crouching, etc, and also ajusting the camera with WASD (or equivalent). Camera adjustment is kind of important. If A and B are located directly below the C-Button keys, then you can easily operate C buttons and A+B by moving your fingers up and down a single line.

With this logic, the ONLY suitable place for A+B is on the bottom row of keys, beneath whatever is used for C-Buttons.

If we supported dual inputs, you could support shoulder buttons on QE *AND* UO. Just a thought. What if PJ64's default featured multiple inputs for A and B and Z and L and R so that people would find the most comfortable? So long as analogue has single input, there should be no issue.

Some people do want dual inputs for PJ64. For example, to bind C-buttons to 360 face buttons AND the right stick. Maybe this would be good solution.

MELERIX:
well, seem my pull is useless, due everyone here seems to play with inverted hands in keyboard xD

AmbientMalice:
@Melerix, I think this is a topic worthy of discussion. I certainly don't want you to feel drowned out. I appreciate your position. I'm sure left handed people might prefer WASD for analogue, too.

You want analogue on WASD? Maybe we could work out a layout with dual binding that supports both Arrow keys and WASD. D-Pad on Num Pad.

cxd4:
I am not surprised MELERIX. :(

I did not have a lot of faith in this pull request the moment I saw it existed.  It's nothing that you did wrong!  It's not that you did anything at all wrong!  It's that there are literally tens of [more than] "comfortable" keyboard N64 gaming setups that people have all gotten to work for them.  It's a huge religious war, and nobody is going to convince everybody to agree on a single best keyboard layout.  Really, the #1 benefit to any (and I mean any, no particular one) keyboard setup, is how used to it you've already been.  Changing from one keyboard setup to another simply takes time and rewiring of old habits in your brain, and it's too hard to convince even someone who uses a masochistic keyboard config that a more convenient one is better, as long as they're used to the masochistic one enough that it "works" for them, due to the challenge in migrating old habits.

Anyway.  I guess to troll around with AmbientMalice some more...

>I feel the analogue stick should be on the right side, not the left, since this is a traditional standard going back decades now. I don't believe in trying to replicate the N64 layout on a keyboard. I believe in taking the FUNCTIONS involved and laying out in a manner which offers maximum flexibility.

I actually don't see what tradition has to do with any of this.  It's traditional that some old games from back in the 80s or 90s used the arrow keys for movement, as well as the fact that others just as old used WASD for movement, as well as the fact that others could have used anything on either the left or the right side of the keyboard for directional movement and still made a comfortable habit out of it.  "Tradition" is a limited measurement because, while some ideas have been around for longer than others, none of them are standard or scientifically established, and often have counterpart software that disagrees and does things a different way.  It may seem "traditional" to you that most software used the right half of the keyboard for directional or analog movement; personally I remember quite a few games that used actually the left half.  What is traditional I would have to say, is the constant disagreement itself.

Now, as far as maximum flexibility.  Anybody can be trained to feel used to (and happy enough with) the left half or the right half of their keyboards being used for analogue.  Could I have stuck myself to the constant habit of IJKL instead of WASD for control stick input, and switched which half of the keyboard is for digital buttons?  Certainly.  However, you said we're talking about **flexibility** here, not what's traditional.  The keyboard mostly is symmetrical, and I see no reason why IJKL side of the symmetry is any more beneficial to flexibility than WASD is.

>Wanna use analogue? Arrows. Wanna use D-pad? Shift your hand to IJKL.

I agree that having to shift the hand for a different set of keys is best undoubtedly for switching to the D-pad; many N64 games would have no reason to use it and for good reason.  The problem here however is that it's really the ARROW keys that are isolated from every other key we use for the buttons.  It's the arrow keys that require the most reaching during day-to-day typing of Internet messages as well as many game keyboard setups, because, it otherwise is going to be a two-handed task.  Therefore, it makes even _more_ sense, to transfer the task of the isolated DPAD input, to the isolated arrow keys.

Why not do it the other way around though, as you said?  Arrow keys should be analog stick not digital.  Well if you do it that way, then only one (most likely your left) hand has access to the buttons.  You of course have proposed a solution where the left hand has easy access to the buttons; the right hand is mostly exclusive to just the arrow keys.  Therefore, it's not that it can't be done.  The highest number of solutions to a logical problem like this, however, arises in a setup where it can be done in many more ways--the more possible solutions there are to a problem, the higher the rational probability that at least X of them are natural, optimal ones.  Coming up with a solution where all of the buttons, non-analog-stick keys, can be easily accessed by a single hand is _possible_, but not highly enumerative of all the possible solutions.

It is much easier, to balance the control.  In the setup I recall giving earlier, my left hand is mostly only ever used for WASD, but it also has access to the START button ('E') and that is it.  Much as you said, my other hand is dedicated to all the remaining buttons:  L and R ('U' and 'I'), Z ('O'), and A and B ('J' and 'K').  It's more than just intuitive:  It works very well.  The downside is the same as with any:  Took me days to get used to it away from the old keyboard setup I was into the habit of.  Of course, since the C buttons can at times be required in conjunction for combo with other buttons, it is necessary for those keyboard mappings to be placed in the proximity of the button keys that I just mentioned, though this has the unfortunate effect of the "PL;'" description.  Still, so long as it's been about the positions of the keys, it detracts nothing.

>I'm sure left handed people might prefer WASD for analogue, too.

I actually am left-handed, come to think of it.
Though I still preferred the arrow keys for the analog stick.  It was a difficult habit for me to break, but I stopped, and now prefer WASD.

>I still think arrow keys should be movement and IJLK (Or UHJK) should be D-pad. The N64 is an analogue-oriented console. The arrow keys are the NATIVE directional inputs for a PC, going back forever.

And this is why I always preferred to assume in my younger life, that the arrow keys had to have been the perfect choice for the control stick--it is the unique section of the keyboard, probably the closest thing on the whole keyboard we have to resembling the N64 analog stick (pressing LEFT arrow and RIGHT arrow at the same time makes much less sense than just pressing 'A' and 'D' at the same time, and this logic compares to control stick vs. digital pad etc.).  Even in spite of that!  In the grand scheme of comfort by itself, I ended up feeling better with WASD.  Someone else convinced me to reprogram my n64 input plugin to use better controls; it took a few days but I realized his setup felt kinda more comfy.

AmbientMalice:
@cxd4 What about multiple buttons for the same function? That seems like a good way to make the emulator's defaults more comfortable for both left handed and right handed people.

cxd4:
Back when I had the arrow keys for the control stick, I also just set them to the digital pad as well and didn't care since basically no game I ever played used them (except Zelda, which I had to change because DPAD left or right had a more troublesome function in Zelda sub-screens than control stick left or right).

So you probably ask about using different keys for the same result.  You could assign both WASD and arrow keys to both be usable as the control stick, but then we have less known options to choose from for deciding between how to assign the C buttons and the digital directional pad.  NUMPAD is probably out of the question since many keyboards don't even have one of those, like my laptop's.  So that just leaves IJKL, which I can't use because those are for my right-handed access to all the N64 buttons.

PR #401, opened 2015-04-18
forced crlf in few files using .gitattributes file

MELERIX:


cxd4:
How is this different than what Melchior was trying to do?

MELERIX:
> How is this different than what Melchior was trying to do?

the only difference is that he did it with tortoise and I did it with .gitattributes file.

> (except the difference that clicking that link to view the commit froze my whole system for a minute)

that is due the amount of files converted to crlf xD

cxd4:
I have to say, I am unfamiliar with this .gitattributes file, so I'm not educated on the details.

Still, if ever Project64 were ported to Linux, wouldn't we still want the RDB files, etc., to be pulled to LF line breaks instead of CRLF ones when pulling or cloning the source on a Unix machine?

MELERIX:
well, if someday there is a PJ64 version for Linux or Mac, the .gitattributes file can be edited to be more specific with some files, for example you can force some specific files to be LF (for Linux), and other files to be CR (for Mac).

here is the documentation: http://git-scm.com/docs/gitattributes

cxd4:
In general I think they should all just be the same encoding as the operating system being pulled on, i.e. all should be LF for Unix and all CR for Mac.

MELERIX:
that can be done using eol=auto, in a future if needed.

cxd4:
It also is default.

I cloned Project64 on Linux once, and all the text had LF and not CRLF.
Do a commit to project64.git on Linux, send a pull request...works all the same as if it was CRLF.
I would like to keep it this way. =]

PR #403, opened 2015-04-19
Added FPS addtion to the Advanced Settings

theboy181:
You can now select the to show the DL/s and VI/s while running a game.  This can be useful when checking different CF factors and attempting to configure better RDB settings.

The formula for DL/s might not be 100% correct in the original code, as sometimes the FPS doesn't seem correct in a few games.

The goal was to have the DL/s and the VI/s to show up in the bottom right corner, but my lack of comprehension on WTL and the status bar is limited.

For now I figure this will be acceptable solution.


Source/Project64/User Interface/Frame Per Second Class.cpp
```diff
@@ -49,16 +49,17 @@ void CFramePerSecond::Reset (bool ClearDisplay)
 		g_Notify->DisplayMessage2(L"");
 		return;
 	}
-	
-	if (m_iFrameRateType == FR_VIs) 
-	{
-		DisplayViCounter(0);
-	}
+
+	if (m_iFrameRateType == FR_VIs)
```
project64: Display DisplayViCounter is never being called now

I assume you meant if (m_iFrameRateType == FR_VIs || m_iFrameRateType == FR_VIs_DLs)

cxd4:
You get the occasional but minor extra spaces/tabs in some of your commits as well @LegendOfDragoon.  It's no big deal, but either Git GUI preview all the code changes before committing helps with that, or `git diff` in the command-line Git server to show all changes not yet committed (or `git diff --cached` to instead analyze for differences just the files staged for commit but not yet committed).

PR #404, opened 2015-04-20
Added FPS addtion to the Advanced Settings (revereted code cleanup)

theboy181:
Added Advanced Settings VI/s and FPS


Source/Project64/User Interface/Frame Per Second Class.cpp
```diff
@@ -58,7 +58,7 @@ void CFramePerSecond::Reset (bool ClearDisplay)
 
 void CFramePerSecond::UpdateViCounter ( void )
 {
-	if (m_iFrameRateType != FR_VIs && m_iFrameRateType != FR_PERCENT)
+	if (m_iFrameRateType != FR_VIs && m_iFrameRateType != FR_PERCENT && m_iFrameRateType != FR_VIs_DLs)
```
project64: this might be right .. but I am not sure it looks like it could be wrong at first glance, I need to work out exactly if it is wrong or not

Source/Project64/User Interface/Frame Per Second Class.cpp
```diff
@@ -102,79 +102,129 @@ void CFramePerSecond::DisplayViCounter(DWORD FrameRate)
...
-{
-	_this->m_iFrameRateType    = g_Settings->LoadDword(UserInterface_FrameDisplayType);
-	_this->Reset(true);
-}
+				Total = 0;
+				for (int count = 0; count < NoOfFrames; count++)
+				{
+					Total += Frames[count];
+				}
+				g_Notify->DisplayMessage2(L"VI/s: %.2f", Frequency / ((double)Total / (NoOfFrames << 3)));
```
project64: is the display here just the same as FR_VIs or is the DL count meant to be here as well

Source/Project64/User Interface/Frame Per Second Class.cpp
```diff
@@ -102,79 +102,129 @@ void CFramePerSecond::DisplayViCounter(DWORD FrameRate)
...
-		if (CurrentFrame > (NoOfFrames << 2)) {
-			__int64 Total;
-			
-			Total = 0;
-			for (int count = 0; count < NoOfFrames; count ++) {
-				Total += Frames[count];
+
+	void CFramePerSecond::UpdateDlCounter(void)
+	{
+		if (m_iFrameRateType != FR_DLs)
```
project64: this can just be one line with &&

project64:
it looks like it is going to keep switching between showing fps and dls

never both at the same time ?

Source/Project64/User Interface/Frame Per Second Class.cpp
theboy181: I think this is needed to make sure the VI/s Counter updates.  if you press F4 the VI/s don't update at all without this line.

PR #411, opened 2015-04-20
optimized cheat codes parsing a little bit

cxd4:
though I imagine nobody will C my point of view.  HAHAHAHA GET IT???


cxd4:
hmph.  tough crowd.

PR #412, opened 2015-04-20
optimized `AsciiToHex` performance and code size

cxd4:
A few places in the RSP debugger and Project64 debuggers/other stuff is the function `AsciiToHex`.  I decided that a switch jump table was not as optimal as a 128-byte look-up table.

Unfortunately I am telling the truth when I say that I don't understand anything about C++ classes or class members (used to know them when I was like 16 but not anymore), so the obvious problem of the re-definition of this function copy-pasted across 4 or 5 different modules of Project64 is not one which I have the C++ knowledge to address.

Also I forgot to create a new branch first, so this ended up being my master branch.  Oh well.

### DIFF
--- a/Source/Project64/N64 System/Cheat Class.cpp
+++ b/Source/Project64/N64 System/Cheat Class.cpp
@@ -488,7 +488,7 @@ static const signed char ASCII_to_hex[128] = {
      -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
 };
 
-DWORD CCheats::AsciiToHex (const char * HexValue)
+DWORD CCheats::AsciiToHex(const char * HexValue)
 {
 	DWORD Count, Value;
 
--- a/Source/Project64/N64 System/Cheat Class.cpp
+++ b/Source/Project64/N64 System/Cheat Class.cpp
@@ -476,42 +476,36 @@ int CCheats::ApplyCheatEntry (CMipsMemory * MMU, const CODES & CodeEntry, int Cu
 	return 1;
 }
 
+static const signed char ASCII_to_hex[128] = {
+     -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+     -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+     -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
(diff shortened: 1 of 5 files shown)


cxd4:
There's also some other look-alike to this function in `numberctrl.h`.  But I decided I would rather avoid touching/breaking that.  Probably still not worth merging due to the duplication problem not being taken care of from the start anyway.

Going to add one commit because it just occurred to me I should say break, not return.

lioncash:
Couldn't AsciiToHex just be simplified to:

```cpp
DWORD AsciiToHex(const char* hex_value)
{
    // Could potentially check errno before returning, etc.
    return std::strtoul(hex_value, nullptr, 16);
}
```

As far as I can tell from really quick glance at the codebase, AsciiToHex is used where a `const char*` parameter would suffice.

cxd4:
Actually yes, used to use that function a lot in some places myself.

Therefore really not sure why I totally forgot about it the entire time till you mentioned it.

As far as I could tell the usage of `AsciiToHex` was generally in places where there most likely was already a C standard library function somewhere to do it.  It could be that in this case he preferred to have low-level control over how string manipulation was being done--the key difference here I think is that if an invalid character (not '0'-'9', 'A'-'F' or 'f') for hex digit conversion was encountered, `AsciiToHex` would just end the conversion there, while `strtoul` would IIRC just return 0?  Though I'm not sure a detail that specific applies anywhere in the code to make a difference.  So I didn't think much of calling a standard library function if zilmar wasn't already to begin with.

Though I imagine that would definitely solve the code duplication issue with repeatedly defining `AsciiToHex` across multiple modules, either with my 128-byte look-up table every time or with that giant switch case for an indexed jump to the i-cache every time.

cxd4:
And yes no reason to not say `const char*` instead of `char*` in this case since read-only properties to any memory should hot have any effect on the conversion to a 32-bit integer.

Went ahead and did that commit just for the sake of clarity.  I'm not actually expecting this pull request to get merged, so maybe wait if @project64 has any thoughts about settling on a simple call to `strtoul` instead of reinventing the wheel.  In a good (probably not Windows) C implementation, calling strtoul probably is faster than my 128-byte LUT anyway.

project64:
strtoul is probably fine, probably be better then having my own function. It should be const char * since the function should not modify the string value. Granted I wrote that function when the code was C code and not using stl/c++

cxd4:
>Granted I wrote that function when the code was C code and not using stl/c++

That should have nothing to do with anything.

cxd4:
Neither method matters due to the size of the code which is not portable.

PR #416, opened 2015-04-22
Cheat Class: Replace is_valid_hex_digit

lioncash:
isxdigit does the same thing.


cxd4:
This is more portable, so I agree with it.

Binary size has not changed but that probably is because Project64 statically links the CRT (/MT Multithreaded DLL) which can remove the `is_valid_hex_digit` procedure but also add whatever nested mess Microsoft usually likes to do for solving most of these basic C library problems.  But I like the change itself in that it's more standard, readable.

PR #421, opened 2015-04-27
[RSP] using standard library function `strtoul` instead of `AsciiToHex`

cxd4:
It was decided in an earlier pull request that it may be better to just use the standard C library function, `strtoul`, instead of multiply defining our own `AsciiToHex` function.

In essence,
```c
#include "rsp.h"
DWORD = AsciiToHex(string);
```
... has the rough equivalent of ...
```c
#include <stdlib.h>
unsigned long = strtoul(string, NULL, 16) & 0xFFFFFFFFul;
```

The `... & 0xFFFFFFFFul` is necessary because @project64 designed the AsciiToHex function to have a maximum string length of 8 characters, so it reads up to only 8 ASCII characters and therefore returns a 32-bit-or-less number.  If you pass a string longer than 8 characters to `strtoul` and `AsciiToHex`, the result could very well be different, if `sizeof(long) != sizeof(DWORD)`, in which case the internal C library implementation of `strtoul` might be working with something else like 64-bit numbers.

So this really is a preventative measure to help portability later down the road.  I'd strongly recommend against the "wait and see" approach of waiting until the end to know a problem should be fixed, because then things just pile up if you want to allow flexibility in your design--both to your benefit as well as everyone else's.  So it's best to prevent portability issues now, not wait until the last minute to solve them.

I only did this change for the RSP because I am waiting for feedback against what I have done here.  Otherwise I was planning on making the same basic changes throughout the rest of the Project64 source, so if I have something wrong in my approach for some reason then please strike now, not later.

Sanity test on the RSP commands stepper after this commit to make sure I didn't break anything:
sp_commands

### DIFF
--- a/Source/RSP/RSP Command.c
+++ b/Source/RSP/RSP Command.c
@@ -432,7 +432,7 @@ void RefreshRSPCommands ( void )
 	if (InRSPCommandsWindow == FALSE) { return; }
 
 	GetWindowText(hAddress,AsciiAddress,sizeof(AsciiAddress));
-	location = strtoul(AsciiAddress, NULL, 16) & 0xFFFFFFFCul;
+	location = strtoul(AsciiAddress, NULL, 16) & ~3u;
 
 	if (location > 0xF88) { location = 0xF88; }
 	for (count = 0 ; count < RSP_MaxCommandLines; count += LinesUsed )
@@ -545,7 +545,7 @@ LRESULT CALLBACK RSP_Commands_Proc (HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM 
 			SCROLLINFO si;
 
 			GetWindowText(hAddress,Value,sizeof(Value));
-			location = strtoul(Value, NULL, 16) & 0xFFFFFFFCul;
+			location = strtoul(Value, NULL, 16) & ~3u;
 
 			switch (LOWORD(wParam))
 			{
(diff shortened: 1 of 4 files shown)


Source/RSP/RSP Command.c
```diff
@@ -431,7 +432,7 @@ void RefreshRSPCommands ( void )
 	if (InRSPCommandsWindow == FALSE) { return; }
 
 	GetWindowText(hAddress,AsciiAddress,sizeof(AsciiAddress));
-	location = AsciiToHex(AsciiAddress) & ~3;
```
project64: what is wrong with the ~3

cxd4: `long` could be a 64-bit type.

`location = strtoul(AsciiAddress, NULL, 16) & ~3;`
If `strtoul` returns 0xFFFFFFFFFFFFFFFF, then `location = 0xFFFFFFFFFFFFFFFC`.

`location = strtoul(AsciiAddress, NULL, 16) & ~3;`
If `strtoul` returns 0xFFFFFFFFFFFFFFFF, then `location = 0x00000000FFFFFFFC`.

I forced a 32-bit result to keep consistent with your max string length limit of 8 characters in your old `AsciiToHex` function but if you don't care about maintaining that property I don't care I can just remove it.

cxd4:
I canceled the FFFFFFFC change.

PR #425, opened 2015-05-01
Update ZLib to 1.2.8

death-droid:


cxd4:
"Showing with 5,249 additions and 2,298 deletions."

to 1.2.8 from what?  like, 0.6.4 or something, no idea never looked at zip but that's a big code change

death-droid:
haha yeah, there's a surprising amount of changes from 1.2.5 to 1.2.8. There is also a fair few portability improvements among a fair chunk of bug fixes according to their changelogs.

PR #429, opened 2015-05-02
Fixed Audio Timing Improvements

Azimer:
- A user-defined FAT value of 0 will cause it to calculate dynamically
- The second buffer shouldn't be part of the GetLegth returned value
- Rounded the GetLength values up and aligned to 4 bytes (16 bit stereo sample size)
- Implemented ai_busy and changed the behavior of ai_full
- Changed the second buffer behavior

I wanted to keep the default behavior the default where if the FAT value is set by the user, that is what will be used.  If we determine the calculated method is better, setting this value to 0 will cause it to calculate.

The DMA registers were handled incorrectly.  The first write triggers AI_DMA_BUSY and the second write triggers AI_FIFO_FULL.  Once the first DMA is complete, AI_FIFO_FULL is unset and an interrupt occurs.  AI_DMA_BUSY is only unset when there are no AI DMAs going.

I rounded the return on GetLength.  I've noticed in some games (Top Gear Rally) if the number of bytes returned are not a multiple of 4 it will corrupt graphics.

Let me know if you have any questions or request any changes.

### DIFF
--- a/Source/Project64/N64 System/Mips/Audio.cpp
+++ b/Source/Project64/N64 System/Mips/Audio.cpp
@@ -26,6 +26,7 @@ void CAudio::Reset ( void )
 	m_Status = 0;
 	m_BytesPerSecond = 0;
 	m_CountsPerByte = g_System->AiCountPerBytes(); // should be calculated ... see below, instead allow from user settings
+	if (m_CountsPerByte == 0) m_CountsPerByte = 500; // If the user has no defined value, grant a default and we will calculate
 	m_FramesPerSecond = 60;
 }
 
@@ -35,10 +36,10 @@ DWORD CAudio::GetLength ( void )
 	DWORD TimeLeft = g_SystemTimer->GetTimer(CSystemTimer::AiTimerInterrupt), Res = 0;
 	if (TimeLeft > 0)
 	{
-		Res = (TimeLeft / m_CountsPerByte) + m_SecondBuff;
+		Res = (TimeLeft / m_CountsPerByte);
 	}
 	WriteTraceF(TraceAudio,__FUNCTION__ ": Done (res = %d, TimeLeft = %d)",Res, TimeLeft);
-	return Res;
+	return (Res+3)&~3;
 }
 
 DWORD CAudio::GetStatus ( void )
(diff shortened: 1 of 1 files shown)


AmbientMalice:
I look forward to testing these changes. FAT 4 LYFE.

Frank-74:
Tested with TWINE, RE2, Perfect Dark, South Park Rally and a few other games.  Set all games to AI=0, Fixed Audio on, Sync using audio off.  Only South Park Rally needed VI's changed to 1400.
I'd say this is an improvement to timing overall.

AmbientMalice:
@Azimer, if dynamic FAT gets good results in general, maybe PJ64's default setting should be FAT=Yes, AI=0, and audio sync disabled. Then problematic games can be adjusted on a per-case basis.

Just a thought.

dsx-:
OH BABY

thanks loads azimer, FAT + AI=0 restores 1.7 FAT behavior on rush games, oot, etc, sounds amazing with every plugin (even jabo's)

Azimer:
Glad the results are going so well.  It's basically the same idea of what was in Project64 1.4 Audio Fix.

dsx-:
yup, no more gaps/crackling/whatever

the changes you made seemed to do something to top gear rally...one of the best ways of getting decent audio from the game was to use pj64 1.7/2.0/2.1/2.2 & your "v0.30 (Old Driver)" plugin and set the plugin's buffer to minimum, but the audio would randomly get more/less glitchy over time, varying from stable to unstable; any plugin would have good audio at some point but then gradually get more garbled, then get better again after another minute or whatever, weird behaviour. But now, using the same 0.30 plugin with minimum buffer, along with FAT+AI=0, the audio is near-perfect and stays this way forever. 14 years since i first investigated this issue, it's finally over. now i guess all that remains is getting the same results from current plugins.

Frank-74:
Found that with this in GetLength, Goldeneye crackles without FAT.
```cpp
	return (Res+3)&~3;
```
I changed it back to `cpp return Res;`, and instead did this in SetViIntr.
```cpp
		m_CountsPerByte = (int)((double)CountsPerSecond / (double)m_BytesPerSecond);
		m_CountsPerByte = (m_CountsPerByte+3)&~3;
```
No longer crackles with or without FAT.

PR #431, opened 2015-05-03
Convert BOOL to bool where possible.

lioncash:
Quite large, however it removes almost all usages of BOOL related defines/typedefs from the core, save for the Windows API where it requires BOOL specifically.

### DIFF
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
@@ -96,7 +96,7 @@ void CMipsMemoryVM::FreeReservedMemory()
 	}
 }
 
-BOOL CMipsMemoryVM::Initialize()
+bool CMipsMemoryVM::Initialize()
 {
 	if (m_RDRAM != NULL)
 	{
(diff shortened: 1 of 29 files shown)


Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
```diff
@@ -2281,59 +2276,54 @@ int CMipsMemoryVM::MemoryFilter( DWORD dwExptCode, void * lpExceptionPointer )
...
+		*Value = 0;
 		return true;
 	}
+
 	if (PAddr >= 0x10000000 && PAddr < 0x16000000) 
 	{
 		g_Notify->BreakPoint(__FILEW__,__LINE__);
 #ifdef tofix
 		if (WrittenToRom)
-		{
```
project64: I prefer to have the brackets, I know it adds more lines but I find it more consistent

cxd4:
>I prefer to have the brackets, I know it adds more lines but I find it more consistent 

Brackets I have always thought were relevant more to scope.
For example when there is a full branch predictor or jump involved.

Something like
```c
if (expression)
{
    dynamic_ops();
}
```
... I used to do a lot and probably wouldn't mind still writing now, as the braces are safer, less hazardous and easier to maintain.

But for something where there really is no branch predictor, like:
```
if (expression)
{
    continue;
}
```
or
```
if (x > 0)
{
    x = 0; /* really should compile to just SHR eax, 31, no branch involved */
}
```
then I find the braces misleading in those particular cases to the static code being generated and probably less readable.  So they're more "consistent" yes, but too much consistency isn't always a good thing.  (Like consistently saying only decimal never hexadecimal, or consistently writing hex numbers in C but never expressing as decimals.)

That said, I can't help but agree that it seems a more divided purpose of clean-up away from the main point of this commit, and would probably have been easier to revert/argue over if it was its own commit.

lioncash:
I've re-appended them.

PR #434, opened 2015-05-04
Remove unnecessary struct/union typedefs

lioncash:
These are superfluous in C++, structs can be declared as is.

I've left the plugin interface structs alone for obvious reasons.

### DIFF
--- a/Source/Project64/Multilanguage/Language Class.h
+++ b/Source/Project64/Multilanguage/Language Class.h
@@ -20,11 +20,11 @@
 typedef std::map<int, std::wstring, std::less<int> > LANG_STRINGS;
 typedef LANG_STRINGS::value_type               LANG_STR;
 
-typedef struct
+struct LanguageFile
 {
 	stdstr Filename;
 	std::wstring LanguageName;
-} LanguageFile;
+};
 
 typedef std::list<LanguageFile> LanguageList;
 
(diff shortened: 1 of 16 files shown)


cxd4:
The C method's better though. :)

Better to declare structs in main source files than header files...with headers it should be declaring it as a type for strictness and emphasis.  Seems less readable to declare the same thing multiple times when really at a concrete level it's just a type of object.

lioncash:
In C++ that is all implicit. It's the same thing. The only time you would need a typedef in C++ is to resolve ambiguous definitions with structs that have the same name, but are in different namespaces; this isn't the case with PJ64.

cxd4:
Implicit is good for less typing, but at least to me, not necessarily more readable.

Maybe there are some cases where implicit code is more portable to different targets than over-thoroughly explicit code is, but for typedef vs. outright declaring I think it's an unnecessary risk that C++ allows.  Just appears more readable to be able to have a prototype of sorts so that the headers will work more portably, whichever the language.

lioncash:
We have very different definitions of "unnecessary risk", then.

cxd4:
https://lh6.googleusercontent.com/-2kxWegNZf6w/VGnsuPx4eWI/AAAAAAAAJjk/gveaZex_CHM/w800-h800/5.png

Issue #436, opened 2015-05-04
[Regression] Everything is broken.


AmbientMalice:
This commit seems to have broken PJ64. https://github.com/project64/project64/commit/6db122ec30161e94adca0c068d713480ae39381b I can't get any games to load after it.

death-droid:
Should still be functioning fine, try running clean on the solution then rebuilding.
The commit makes no real change in the functionality of the code.

AmbientMalice:
I did a fresh git pull.

Now this is odd.

Most games, a prime example being Mario 64, are refusing to boot.

But Aerogauge boots.

I'm pretty sure this commit is at fault. I'll do some more testing.

cxd4:
>The commit makes no real change in the functionality of the code.

I would not be too certain of that, even though you may be right.  `bool` true and false is not the same as C true and false or the WINAPI kind.  If at any point he changed the C version for an extern/instance of it used in the code to the C99 or C++ version using the Boolean type, but not the other instance, it's possible it can change functionality.  The C version was also more powerful...if you convert logical/bitwise applications of true/false to the limited C++ version it can alter things potentially.

Anyway bool may suck, but it's at least more portable/readable than WINAPI types, so I shouldn't complain.

death-droid:
Ok actually had a chance of testing out the commit, and it indeed breaks everything.

death-droid:
Found out where exactly the code was breaking and made a quick pull request to solve it. Will have to look into it a bit more deeply at some stage though

AmbientMalice:
Excellent work. I'll close this.

cxd4:
This is why dividing large changes into multiple commits is a good thing.

If it was more than one commit we could just keep checking out until the exact region where he accidentally broke everything.  Instead what could maybe have been 10 commits was all done in 1 commit, so it takes too long to figure exactly where it went wrong.

PR #438, opened 2015-05-04
Revert "Convert BOOL to bool where possible."

death-droid:
Commit breaks emulation. Needs further work before it should be in main.


death-droid:
Resolves issue #436

PR #439, opened 2015-05-04
System events seems to rely on remaining as an integer type. Changing it...

death-droid:
... to bool causes the whole emulator to stop functioning correctly.


project64:
converted it back to BOOL (https://github.com/project64/project64/commit/9aa9061f074403311b35a0d79f2a741984b7bcc0)

death-droid:
Closing :) Zilmar has now implemented the fix

PR #440, opened 2015-05-05
Revert DoSomething from bool to BOOL #2

Azimer:
This fixes the interpreter


Azimer:
This should take care of it @Frank-74.  If all is good I think it's ready to merge @project64.

Frank-74:
All is good.  I've tested CPU Interpreter with some games.

PR #441, opened 2015-05-06
Make AudioResetOnLoad work with Fixed Audio Timing

LegendOfDragoon:


AmbientMalice:
The AudioResetOnLoad hack wasn't needed for games running FAT, AFAIK. Unless this serves some other purpose?

LegendOfDragoon:
@AmbientMalice It actually was needed. I made a save state with FAT disabled and had no audio when loading save state with FAT enabled.

AmbientMalice:
Ah, I see.

PR #452, opened 2015-05-11
Added N64 Mouse support in NRage

LuigiBlood:


Source/nragev20/NRagePluginV2.rc
```diff
@@ -524,25 +524,29 @@ END
...
 
 VS_VERSION_INFO VERSIONINFO
- FILEVERSION        VER_FILE_VERSION
- PRODUCTVERSION     VER_PRODUCT_VERSION
- FILEFLAGSMASK      0x3fL
- FILEFLAGS          VER_FILEFLAGS
- FILEOS             VER_FILEOS
- FILETYPE           VER_FILETYPE
- FILESUBTYPE        0x0L
+ FILEVERSION 2,4,0,4
```
project64: any particular reason changing the version structure has to do with the mouse?

LuigiBlood:
That might have been VS2013 changing everything when I added a new checkbox maybe...?
I didn't really see that at first.

project64:
I find I check the difference before checking in and revert the changes VS sometimes makes so it is only the relevant changes I need.

LuigiBlood:
I should have fixed the problem now. I used the original file and changed only the lines that should be changed.

PR #463, opened 2015-05-15
Disable 32bit for all 64DD cart hacks.

AmbientMalice:
Sim City 64 complains about SP DMA errors if 32 bit mode is enabled. I think for safety, it should be disabled for all these titles.


AmbientMalice:
I did some more testing. The SP DMA errors aren't actually 32 bit mode related. They're fixed, bizarrely, by enabling Fixed Audio Timing. This also fixes flickering\glitched sprites.

PR #478, opened 2015-05-23
Simplify platform-specific Property Sheets

JunielKatarn:
In order to stop relying on 'Conditional' predicates for project configuration, add as many settings as possible into platform base property sheet.
Also, instead of testing for platform value (Win32|x64), create the approrpiate Configuration/Platform files for x64. (By creating the base platform property sheet mentioned above, adding new combinations should be trivial. This could open the door to adding ARM platform in the future, and easily configure its specific settings).

Testing done:
Built all combinations (Debug, Release) x (Win32, x64).
Results identical to current master branch.


project64:
couldn't we just rename Win32.props sheet to Platform.props and this basically would have the same effect?

JunielKatarn:
if you mean renaming but NOT creating a new Win32.props, then no.
The goal of this change is to be able to interpolate the platform/config combination in the VCXPROJ files (i.e. "$(Platform).$(Configuration).props"). Also, platform-specific prop sheets should be fully independent from each other (thus the need for the base 'Platform.props').

If you mean renaming instead of moving content, then yes.

Is there anything else I can do for this pull request to be promoted?

project64:
So you are creating 4 new property sheet so there is no condition on 4 settings.

JunielKatarn:
Pretty much. The long-term idea is not only to get rid of conditions in VCXPROJ files, but to make them as simple as possible.
This will make platform-specific fine-tunning and performance enhancements easier.

Also, adding a new platform would require to create roughly two files (Debug, Release), which is not bound to happen a lot (maybe in the future we will have ARM as a platform?). So, I'm confident we won't get a large or unmanageable number of .prop files.
That said, the extra files (Platform.Config.props combinations) will be basically pass-through imports.

project64:
the reason there is so many property sheets at the moment is there is the set from vs2008 and a duplication for vs2013. The thing in the vs2008 property sheets the conditioning does not work so well in it, where in the .props files (vs2013) the conditions work really well. 

I am more inclined to actually try to move to a single .props file as it is easy to see where something applies and there is one place to edit and not like 8.

cxd4:
Yeah having vs2008 and vs2013 versions of everything started to make some questionable situation, worth it probably.  Everything VS2010 and afterwards uses the cleaner, XML-ish style of property config, whereas everything 2008 and earlier uses the old system for conditions.

JunielKatarn:
So, is this the proposed scheme?

[Props Directory]
- Add (moved from Win32.props) Platform.props. Concentrate all configs there, using conditinals for platform-specifics.
- Discard x64.props.

[VCXPROJ files]
- Import "Platform.$(Configuration).props" insteado of "Win32.$(Configuration).props".

project64:
yes I think, maybe the property sheets should have all vs2008 sheets in one dir and vs2013 in another one .. but then if vs2013 ends up being just one file then that probably is not needed. 

so for vs2013 it would be just inherit Platform.props

JunielKatarn:
Hi. I have updated the code review. The new changes simplify VCXPROJ files even further.
Please provide any feedback necessary.

JunielKatarn:
Hi @project64 . This change simplifies .VCXPROJ files a lot, and makes configuration of potential new platforms much easier. Would you take a look and approve if no further work is required?

PR #480, opened 2015-05-24
warning fixes

cxd4:
warning fixes for the 32-bit build

No reason to fix 64-bit warnings because you risk breaking the build's functionality even more without the ability to test it / isolate what the genuine fix really is first, so only 32-bit warnings fixed.

The fourth commit name:
"[Project64] fixed warning C4100: unreferenced formal parameter"
I was supposed to say [zlib] not [Project64].

### DIFF
--- a/Source/3rd Party/zlib/contrib/minizip/crypt.h
+++ b/Source/3rd Party/zlib/contrib/minizip/crypt.h
@@ -32,7 +32,7 @@
 /***********************************************************************
  * Return the next byte in the pseudo-random sequence
  */
-static int decrypt_byte(unsigned long* pkeys, const z_crc_t* pcrc_32_tab)
+static int decrypt_byte(unsigned long* pkeys)
 {
     unsigned temp;  /* POTENTIAL BUG:  temp*(temp^1) may overflow in an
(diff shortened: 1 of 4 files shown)


project64:
generally I am not touching 3rd party code which is why I probaly did not change it my self .. oh well looks fine

cxd4:
Not my favorite thing to touch either, but if it's going to be embedded into this repository then it's going to share its syntax problems, bad/nonstandard coding practices, portability issues and whatever else is wrong with the 3rd Party code with any problems the Project64 code might have.

For example I thought this was dumb.
```c
int function(unsigned byte)
{
    return (byte & 0xFF);
}
```
Since 7-zip is always compiled on a system where sizeof(int) > sizeof(char) (or else would violate c90 rules), it's impossible that making this change would break anything about 7z support:
```c
unsigned char function(unsigned byte)
{
    return ((unsigned char)(byte & 0xFF));
}
```
Even though the signedness and size vary, this should break nothing while using smaller types to be more flexible to implicit type conversions.

Maybe by the time death-droid has a newer upgrade to the 7-zip libraries to pull into the repository the official 7z guys will have the malpractice fixed on their end.

PR #484, opened 2015-06-05
(Glide64) Remove dynamic inline asm detection of SIMD function pointers.

cxd4:
This seems to be half of the reasons why PJGlide64 does not yet build as 64-bit graphics plugin.

It does run-time detection of SIMD extensions and dynamically adjusts the default HLE function pointers to point to special-case functions with 3dnow!, SSE2, SSE3, NEON etc. machine code done in terms of inline asm.

First, inline asm is retarded.  Any modern, optimizing compiler can perfectly well vectorize simple loops of small steps like `for (i = 0; i < 4; i++) v32[i] &= s32[i];` and anyone who would write non-portable intrinsics or asm for something that direct is easily full of it.

Second, many of the SSE implementations are way larger than the native C implementations (which sometimes can only have 3 elements to operate on, therefore wasting a vector element) and pollute the instruction cache--to the point where you would consider not using SSE at all for those cases.  (In other words, even non-vectorized ANSI is faster than such messes.)  Maybe there are several better ways to do the SIMD in ways that aren't as poorly written, but in either case these functions need to be logged over for hazard-proofing and relevance in a profiler to performance.


Source/Glide64/3dmath.cpp
```diff
@@ -151,48 +151,119 @@ void calc_sphere (VERTEX *v)
...
-    len = sqrtf(v[0]*v[0] + v[1]*v[1] + v[2]*v[2]);
-    if (len > 0.0f)
-    {
-        v[0] /= len;
-        v[1] /= len;
-        v[2] /= len;
-    }
+
+    len = v[0]*v[0] + v[1]*v[1] + v[2]*v[2];
+    if (len == 0.0f)
```
JunielKatarn: Is this comparison safe? (floating point comparison).
What if you don't get an "exact" zero (say, 0.0000001).

Source/Glide64/3dmath.cpp
```diff
@@ -151,48 +151,119 @@ void calc_sphere (VERTEX *v)
...
 void TransformVectorC(float *src, float *dst, float mat[4][4])
 {
-  dst[0] = mat[0][0]*src[0] + mat[1][0]*src[1] + mat[2][0]*src[2];
-  dst[1] = mat[0][1]*src[0] + mat[1][1]*src[1] + mat[2][1]*src[2];
-  dst[2] = mat[0][2]*src[0] + mat[1][2]*src[1] + mat[2][2]*src[2];
+    float products[4][4];
+    register int i, j;
+    const int limit = 4 - 1; /* either xmm[127..0] or just xmm[95..0] works */
+
+#if 0
```
JunielKatarn: If the #else section is unconditionally better, why keep this one?

Source/Glide64/3dmath.cpp
```diff
@@ -151,48 +151,119 @@ void calc_sphere (VERTEX *v)
...
+    for (j = 0; j < limit; j++)
+        for (i = 0; i < limit; i++)
+            products[j][i] = mat[j][i] * src[j];
+#else
+    for (j = 0; j < limit; j++)
+        for (i = 0; i < limit; i++)
+            products[j][i]  = src[j]; /* _mm_set1_epi32(src[j]):  PSHUFDW */
```
JunielKatarn: How is this two-loop scheme better than the #if 0 section?

Source/Glide64/3dmath.cpp
cxd4: That's the point, you see.  You're not able to divide by an "exact" zero.
That's why the check is there.

Source/Glide64/3dmath.cpp
cxd4: You assume that it's unconditionally better; both methods of expressing the algorithm have their readability vs. parallelization advantages, much in the same way you should believe in commenting your sources with C comments.  Unless you're of the type who believes comments are a detriment to C programming I'm not sure why you would ask me about why to keep/throw out the #if 0 version of an identical algorithm--in fact it's even easier to debug.

Source/Glide64/3dmath.cpp
cxd4: Because there's no such thing as a SSE2 instruction that multiplies rd = rs * rt all at once.  You need to divide the storage to its own step, then multiply by a source operand.  It's more organized--the divided split loops show these micro-steps better in the architecture's limitation.

And again, you're supposed to be able to change #if 0 to #if 1 in case you need to debug anything about the code during testing; that's the advantage over just /* commenting out */ all of your code.

Source/Glide64/3dmath.cpp
JunielKatarn: Yeah, I assumed you added that change because it would somehow run better. Your SSE2 explanation makes it clear.
One suggestion, though. There may be others who, like me, are not familiar with SSE2 parallelization. Can you add a small comment explaining that you keep the #if 0 section for easier debugging, but also the reason of the default implementation change?

cxd4: Perhaps, something like this?  https://github.com/cxd4/project64/commit/7d6ff5d29c2994a42d9782bf99d4dbc0648ba48e

The problem with SSE code is that even if it is faster (which it very well can be), it ironically can use up more space in the byte-code cache.  Even if this code is faster it can have a less direct effect outside of the functions--therefore in my opinion we should always default to optimizing algorithm for compact size and small binary builds, before testing with a profiler to see where the bottlenecks are and inverting the approach to maximum performance.  (Plus the #if 0 stuff helps people understand the loops...it's kind of easy to see something like (2x + 4x + 6x + 8x + ...) iterated within the C source, than only a loop on += 2*i*x or whatever.)

cxd4:
Change of plans, as 64-bit portability should be in a separate branch and fork written in C.

PR #499, opened 2015-06-19
Avoid defining a macro to a comment as the macro definition.

cxd4:


cxd4:
This one I don't care so much about; I'm happy to just cancel/close the changes if no one else is interested.  It wasn't presented in the right way but this PR was started merely because of defining a macro to a comment (while this generally has no negative implications it can alter the syntactical nature of the macro's existence in subtle ways and I think if it's not intended then it's best to prevent that).

PR #500, opened 2015-06-19
DRAM address range bounds checking was off by one byte.

cxd4:


Source/RSP/dma.c
```diff
@@ -39,7 +39,7 @@ void SP_DMA_READ (void)
 
     addr = (*RSPInfo.SP_DRAM_ADDR_REG) & 0x00FFFFFF;
 
-	if (addr > 0x800000)
+	if (addr > 0x7FFFFF)
```
cxd4: The alternative solution:
```c
if (addr >= 0x800000UL)
```
as without the "UL" at the end, a 24-bit signed int type could equate that to
```c
if (addr > INT_MIN)
```
which would have caused the MessageBox 99% of the time and have been a bug, so I like 7FFFFF.

oddMLan:
I like your new branch naming convention. (y)

V1del:
10/10 branch names, would get finger stuck on one key again :+1:

cxd4:
9999/9999 would branch again

PR #503, opened 2015-06-19
[PJGlide64] TransformVector C is better than TransformVectorSSE.

cxd4:


tony971:
Bumping this and #504 as they're both needed for x64 glide.

cxd4:
Probably the reason it took so long to merge this pull request was that it didn't make clear that `TransformVector C` is the default function pointer initialized to `TransformVector`.  Glide64 likes to initialize an optional function named `TransformVectorSSE` to `TransformVector` if it detects SSE support at run-time.

I see no point to this because the C code can optimize to SSE code in modern compilers.  It's also better than writing SSE in inline asm instead of C intrinsic functions.

PR #505, opened 2015-06-19
[PJGlide64] imul16: no need for inline asm here

cxd4:
Recreated from old pull request dead from merge conflict:  https://github.com/project64/project64/pull/502.

PR #506, opened 2015-06-19
Convince JunielKatarn to stop making crappy reports.

cxd4:
Appendage to https://github.com/project64/project64/pull/505 to remove the surviving remnants of inline asm linkage.


JunielKatarn:
HAHAHAHA

OK, tested again. Builds and runs fine.
Atta boy!

cxd4:
You would be in the thousands of people who have tested this change already, considering I already did the exact same changes as these to the libretro fork of Glide64, which is written in C.

There were no mistakes in either of my commits; you just ran into a link error on your own build's end.

PR #508, opened 2015-06-19
[RSP] Ternary, SHMERNARY!  Blasphemous.

cxd4:
I think zilmar wants to do it this way:
```c
if (bool)
{
    z += y;
}
else
{
    z += x;
}
```

I would rather say this:
```c
z += bool ? y : x;
```


cxd4:
On second thought bool ? y : x would be kind of arbitrary.  The if/else isn't so bad for that case since it can't be compiled to static code anyway, so my example wasn't the best.

PR #511, opened 2015-06-19
[RSP] re-ordered VMULF branch prediction:  corner case least likely

cxd4:
It's supposed to be if (A && B), not if (NOT a || NOT b).

### DIFF
--- a/Source/RSP/Interpreter Ops.c
+++ b/Source/RSP/Interpreter Ops.c
@@ -498,7 +498,16 @@ void RSP_Vector_VMULF (void) {
 	for (el = 0; el < 8; el ++ ) {
 		del = EleSpec[RSPOpC.rs].B[el];
 
-		if (RSP_Vect[RSPOpC.rd].UHW[el] != 0x8000 || RSP_Vect[RSPOpC.rt].UHW[del] != 0x8000) {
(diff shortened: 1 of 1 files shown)


project64:
is the code actually wrong, it seem like this is meant to be an optimization as the most likely jump in executed first?

cxd4:
There are three small reasons why I did this change, but the original code has nothing wrong in it.

I just think it's more readable.

First,
if (!a || !b || !c || !d || !e)
is a pet peeve of mine.  It's just logically more readable to say
if (a && b && c && d && e).
usually don't like double negatives in logic gates.

Second, like you said it is not very likely that source == 0x8000 && target == 0x8000.  The main algorithm in my opinion should be under the `else`, where no branch was taken.

Taking a branch from an if (expression) implies an exception, something unusual that you don't normally do to solve the problem.  It's just more readable to say
```c
if (n == 0)
    return UNDEFINED;
return (m / n);
```
than it is to say
```c
if (n != 0)
    return (m / n);
return UNDEFINED;
```
because the way the problem SHOULD be solved most of the time, is what the `else` is made for.

Third, yes my change has to do partly with optimization, though it won't make the MSVC build any faster since MSVC ignores branch prediction advice anyway.

cxd4:
The only reason I don't want this to be merged anymore is because there is a different pull request that matters more than this one, and conflicts with it.

PR #513, opened 2015-06-20
Document the RSP op-code matrices + static disassembler debugger code.

cxd4:
These changes are mostly demonstrative.

They're to give an idea of what **could** be done to improve the RSP plugin, but they are not complete.  Therefore they're less for merging directly, more for discussing.  But if it looks good then I guess merge.

Basically with the **type** of changes done in these commits, we can eradicate the `switch` statements in the RSP command stepper source code, thus saving many KB of DLL file size as well as improving the maintainability of the code (centralized to a look-up table of opcode strings instead of copy-pasting the same duplicated lines of code 64 or 32 times per type of op).

### DIFF
--- a/Source/RSP/RSP Command.c
+++ b/Source/RSP/RSP Command.c
@@ -793,6 +793,18 @@ void RSP_Commands_Setup ( HWND hDlg )
 	SetWindowPos(hDlg,NULL,X,Y,WindowWidth,WindowHeight, SWP_NOZORDER | SWP_SHOWWINDOW);
 }
 
+static const char unused_op[] = "invalid";
+static const char* mnemonics_primary[8 << 3] = {
+    "SPECIAL","REGIMM" ,"J"      ,"JAL"    ,"BEQ"    ,"BNE"    ,"BLEZ"   ,"BGTZ"   ,
+    "ADDI"   ,"ADDIU"  ,"SLTI"   ,"SLTIU"  ,"ANDI"   ,"ORI"    ,"XORI"   ,"LUI"    ,
+    "COP0"   ,unused_op,"COP2"   ,unused_op,unused_op,unused_op,unused_op,unused_op,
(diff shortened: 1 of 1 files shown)


Source/RSP/RSP Command.c
```diff
@@ -1238,59 +1279,28 @@ char * RSPSc2Name ( DWORD OpCode, DWORD PC )
...
+			command.Ascii[2],
+			command.Ascii[1],
+			command.Ascii[0]
+		);
+	}
+	else
+	{
+		sprintf(
+			CommandName,
+			"%s\t$v%d[%d], 0x%04X(%s)",
```
cxd4: Here is where we can start to document actual opcode formats to the RSP.
```c
"%s\t$%d, $%d, $%d" // R-type instructions:  rd, rs, rt
"%s\t$%d, $%d, 0x%04X" // I-type instructions:  rt, rs, immediate
"%s\t0x%07X" // J-type instructions:  (target << 2)
"%s\t$v%d, $v%d, $v%d[%s]" // most vector unit computational instructions
"%s\t$%d, 0x%03X" // I-type, register-immediate RSP-encoded
...
```
These are just some of the ideas @project64 that I think can be worth warranting a discussion in the future.

cxd4:
Finished a custom RSP disassembler yesterday.
It turns out, there are 16 possible instruction formats on the RSP's MIPS.
https://github.com/cxd4/rsplog/blob/master/disasm.c#L114

So this pull request again is just a partial illustration.  I did the full condensed disassembler at:
https://github.com/cxd4/rsplog/blob/master/disasm.c#L232

cxd4:
This merge is a surprise, but since it's merged I think the same change can be done to almost the rest of the stepping functions as well.  So I think I will see about doing more PRs to distribute these changes to the rest of them.

PR #518, opened 2015-06-21
[RSP] unified VR[vt] shuffling to its own parallel step func

cxd4:
I'm sure this is too long to scrutinize and merge, but I guess that can't be helped.

It's exactly one of the greater kinds of changes the whole vector unit implementation in pj RSP plugin needs desperately for cleaner code, and it needs to be highlighted in some form whether it is able to get merged or not.

### DIFF
--- a/Source/RSP/Interpreter CPU.c
+++ b/Source/RSP/Interpreter CPU.c
@@ -402,6 +402,20 @@ void BuildInterpreterCPU(void) {
 	RSP_Sc2[31] = rsp_UnknownOpcode;
 }
 
+VECTOR RSP_shuffle(VECTOR target, unsigned int element)
+{
+    VECTOR result;
+    register int i;
+
+    result = target; /* "uninitialized" variable warnings */
+    element &= 0xF;
+    for (i = 0; i < 8; i++)
+    {
+        result.HW[i] = EleSpec[0x10 | element].HW[i];
+    }
+    return (result);
+}
+
 DWORD RunInterpreterCPU(DWORD Cycles) {
 	DWORD CycleCount;
 	RSP_Running = TRUE;
--- a/Source/RSP/Interpreter CPU.h
+++ b/Source/RSP/Interpreter CPU.h
@@ -24,6 +24,8 @@
  *
  */
 
+#include "types.h"
+
 #define NORMAL				    0
 #define DO_DELAY_SLOT 			1
 #define DELAY_SLOT 				2
@@ -40,3 +42,5 @@ extern DWORD RSP_NextInstruction, RSP_JumpTo, RSP_MfStatusCount;
 
 void BuildInterpreterCPU(void);
 DWORD RunInterpreterCPU(DWORD Cycles);
+
+VECTOR RSP_shuffle(VECTOR target, unsigned int element);
--- a/Source/RSP/Interpreter Ops.c
+++ b/Source/RSP/Interpreter Ops.c
@@ -491,15 +491,14 @@ void RSP_COP2_VECTOR (void) {
 
 /************************** Vect functions **************************/
 void RSP_Vector_VMULF (void) {
-	int el, del;
+	int el;
 	UWORD32 temp;
 	VECTOR result = {0};
+	VECTOR target = RSP_shuffle(RSP_Vect[RSPOpC.rt], RSPOpC.rs);
 
 	for (el = 0; el < 8; el ++ ) {
-		del = EleSpec[RSPOpC.rs].B[el];
-
-		if (RSP_Vect[RSPOpC.rd].UHW[el] != 0x8000 || RSP_Vect[RSPOpC.rt].UHW[del] != 0x8000) {
-			temp.W = ((long)RSP_Vect[RSPOpC.rd].HW[el] * (long)RSP_Vect[RSPOpC.rt].HW[del]) << 1; 
+		if (RSP_Vect[RSPOpC.rd].UHW[el] != 0x8000 || target.UHW[el] != 0x8000) {
+			temp.W = ((long)RSP_Vect[RSPOpC.rd].HW[el] * (long)target.HW[el]) << 1;
 			temp.UW += 0x8000;
 			RSP_ACCUM[el].HW[2] = temp.HW[1];
 			RSP_ACCUM[el].HW[1] = temp.HW[0];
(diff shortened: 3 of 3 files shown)


cxd4:
essential change is

Before:
```c
void VAND(unsigned vd, unsigned vs, unsigned vt)
{
    VECTOR result;
    int el, del;

    for (el = 0; el < 8; el++)
    {
        del = EleSpec[RSP_Vect[RSPOpC.rs].UHW[el];
        result.UHW[el] = RSP_Vect[vs].UHW[el] & RSP_Vect[vt].UHW[del];
    }
    RSP_Vect[vd] = result;
}

After:

void VAND(unsigned vd, unsigned vs, unsigned vt)
{
    VECTOR result;
    VECTOR source = RSP_Vect[vs];
    VECTOR target = RSP_shuffle(RSP_Vect[vt], RSPOpC.rs);
    int i;

    for (i = 0; i < 8; i++)
    {
        result.UHW[i] = source.UHW[i] & target.UHW[i];
    }
    RSP_Vect[vd] = result;
}

**Comment by cxd4** [comment]
Also I suppose if implemented it can help towards removing the EleSpec arrays from the interpreter and recompiler cores in some future change, as it's not needed for hardware-accurate shuffling which has an algorithm that can be expressed clearly without LUTs.

**Comment by project64** [comment]
ok del might not be needed as much now that writing to a temp vector to write output so overright is not an issue. 

but for example in your RSP_Vector_VMUDL 
it was:
del = EleSpec[RSPOpC.rs].B[el];
temp.UW = (DWORD)RSP_Vect[RSPOpC.rd].UHW[el] * (DWORD)RSP_Vect[RSPOpC.rt].UHW[del];

now
temp.UW = (DWORD)RSP_Vect[RSPOpC.rd].UHW[el] * (DWORD)target.UHW[el];

this seems wrong as something like vmudl v0, v1, v2[1] seems like it could be wrong (it probably fine). But I do not think the code is any cleaner with you moving the vector register in to the temp and using that as a multiplier. That you have the RSP_shuffle that is finding the element and copying it I find you are hiding information and can make the code less clear on what is going on.

I like trying to keep the interpreter as less changed as possible, mostly I am trying to avoid accidentally introducing bugs in.

**Comment by cxd4** [comment]
>this seems wrong as something like vmudl v0, v1, v2[1] seems like it could be wrong (it probably fine).

I don't understand why you would expect a lesser chance of my function being safe for `vmudl v0, v1, v2[1]` ... it is the same for all shuffle masks, no less stable than the LUT you are using to re-order the elements that way.  (In fact that's currently all the func uses.)

Any chance you have of being correct on that is the same chance of it being impossible to optimize the RSP recompiler for vector register shuffling--reorder RSP_Vect[rt] to be in parallel to the order of RSP_Vect[rd] for direct recompiler SIMD to emulate the RSP opcodes.  Unless I'm misreading something about what you said there's no way my function method is any less fine than your LUT method.

>That you have the RSP_shuffle that is finding the element and copying it I find you are hiding information and can make the code less clear on what is going on.

There is no information that I am hiding that your `EleSpec[]` array doesn't hide.

That array shows no algorithm at all about RSP hardware shuffling; if my function can be dismissed as hiding information then so can the original array you were using, which I am sure came straight from Michael Tedder's RSPINFO.TXT (too bad it still has nothing to do with the algorithm on the hardware).

The only reason why the function I added reads your EleSpec[] array anyway (which of course gives the redundant look that I am hiding information, since you were referencing that array directly to begin with) is just as a temporary means of avoiding breaking the interpreter with bugs.  But eventually that array would best be deleted.

**Comment by cxd4** [comment]
It's virtually the same thing as what you did.

You created `VECTOR Result = { 0 };` to prebuffer the computed result of the vector source and target, then move the whole 128 bits to RSP_Vect[sa].

My function does a 128-bit shuffle of [vt] to arrange it to be parallel to the order of [vs] so that not only `del` can be deleted from all of your functions, but the loops have less work to do in every function and fewer instructions in all of them.

Surely you wouldn't rather interleave mixed operations and prevent yourself from being able to use a direct `memcpy` to do the whole thing in a single hit?

**Comment by project64** [comment]
I agree It's virtually the same thing, that is why I do not see much value it.

I do not care as much as speeding up the RSP interpreter, if it effects the re compiler and can make that faster, then I like that.

**Comment by cxd4** [comment]
It's the same *now* because the function has not yet been changed to get rid of the EleSpec dependency.

The function:
* gives you the ability to make it not the same thing
* lets you do more things than just reading from a LUT in every opcode
* can be marked __noinline to make the interpreter even slower if you want :)

Doesn't matter; it's clear that this is less about speed, more about smaller functions and smaller code, in both the interpreter as well as the recompiler.  The fact that it could help speed out is incidental; in fact I'm qute sure that this pull request makes the interpreter even slower, not faster, contrary to what you think my intentions may be.  Now, this function does make it easier to make the interpreter faster, but that's not what is being done in this pull request, as the function is still very slow and calls your EleSpec LUT either way.

By the time the RSP re-compiler is vectorized to re-order the elements in a single instruction (or two instructions since Intel sucks and is a dumb architecture to make recompilers for in general), your `EleSpec[]` array becomes obsolete anyway and should be removed from both cores.

**Comment by cxd4** [comment]
Nobody ever documents vector instructions that way ...
```c
void vector_add(int16_t * VD, int16_t * VS, int16_t * VT)
{
    for (int i = 0; i < 8; i++)
        VD[i] = VS[i] + VT[EleSpec[rs].HW[i]];
}
```
That, sir, is not readable at all.

It should just have been:
```c
void vector_add(int16_t * VD, int16_t * VS, int16_t * VT)
{
    shuffle_vector(&VT[i], rs);
    for (int i = 0; i < 8; i++)
        VD[i] = VS[i] + VT[i];
}
```
Did it not ever occur to you that since all of the RSP vector instructions scramble the elements of VT in a order defined by rs, that the shuffling could be a global part of the instruction/register decode phases of execution on the hardware?

If you truly want the interpreter to focus on readability and accuracy, consider not keeping so much cruft behind and unify what goes on in every single function, to be stored only in the function calling said table of functions.  DLL size, maintainability to future structural changes (like rewriting from C to C++ using classes AHHHHHHHHHHHHHHHH Salesman!) and centralized interface to this one function for optimizing also the recompiler are just some of the benefits.

I don't find LUTs readable at all.  I think they are only warranted when there is no known pattern to explain the functional algorithm that is going on--there is a clear pattern with RSP shuffles so there is no need for you to keep the EleSpec[] LUT, in either of the cores (especially not the re-compiler).

cxd4:
Alright, forget this I changed my mind.

There is one good reason why I can think of to cancel these changes and that is because it makes you have to say VR[vd][i] = VR[vs][i] + shuffled[i] instead of VR[vt][EleSpec[i]] or whatever...I can understand wanting to say RSP_Vect[] all the time instead of adding a new temporary vector like `VECTOR Result` or `VECTOR shuffled` or whatever.  (Then again, had this PR been merged, it would have allowed you to delete the `VECTOR result = { 0 };` from every function and be immune to the overwrite.)

Ultimately it boils down to a tradeoff between readability _opposed to_ hardware accuracy, which maybe the hardware does use a look-up table to reorder the vector elements like you do already.

Besides, I just thought of something more important about the interpreter algorithm to change than this, and for certain I think it is no more likely to get merged :P but the changes must be shown.

cxd4:
I suppose the only way to prevent that from being an issue was for you to change your funcs from:
```c
void VMULF(void) {
```
to
```c
void VMULF(int16 * VD, int16 * VS, int16 * VT) {
```
and just shuffle VT to a temporary vector before calling the COP2 function pointers table, that way you would not have had to declare a new variable in all of the functions.

However that might not be reasonable since most functions maintain no argument list to begin with.

PR #519, opened 2015-06-22
[RSP] guessing an enumerated list may be more maintainable here?

cxd4:
Now for my final OCD pull request before I take a good break from harassing the repo owner.

Looking at the interpret CPU header, none of the macros seem to need the exact integer they're defined to.  For example
```c
#define DO_DELAY_SLOT       1
```
It could be defined to 2, 3, 4, 9001, anything and still have the same effect (or so a brief scroll through the core code and testing my compiled changes from this PR with World Driver Championship suggests).

The nice thing about `enum`s is that all their values can be subtly but instantly updated without changing the actual line of text in a Git commit, but just their relative position to other lines in the enumerated list.  As long as each constant only needs a unique, arbitrary value (nothing in particular) then I think it sounds perfect that way.  It also avoids the potential language vulnerability issues of macros.

### DIFF
--- a/Source/RSP/Interpreter CPU.h
+++ b/Source/RSP/Interpreter CPU.h
@@ -24,17 +24,19 @@
  *
  */
 
-#define NORMAL				    0
-#define DO_DELAY_SLOT 			1
-#define DELAY_SLOT 				2
-#define DELAY_SLOT_DONE			3
-#define DELAY_SLOT_EXIT			4
-#define DELAY_SLOT_EXIT_DONE	5
-#define JUMP	 				6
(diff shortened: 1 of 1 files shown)


project64:
if you are going to change it to an emu, then it should have a type and the variables using it use that type, then it will be type checked .... wait it typechecked in c++ .. not sure how that will work in c code .. maybe I should just change all the files from c to c++

cxd4:
lol.  do whatever you want in that case.

No clue about C++ but as far as I'm concerned, enum == C macros with more maintainability to them.

PR #526, opened 2015-06-24
[RSP] unified signed clamp for all opcodes reading the accumulator

cxd4:
Have fun analyzing this one. :) It moves the hardware function of clamping 32-bit accumulator slices to 16-bit result slices to a central function rather than doing the chains of segmented if/else in each vector opcode.

So yeah it's more readable and maintainable, but I hope that nobody plans to test the interpreter for a joyride or do extensive debugging if this gets merged because my changes here totally tank the interpreter's speed.  Though I have to say it is kind of fun to make interpreters slower, after having done the opposite for so many years.


cxd4:
son, i am accumulate.

PR #527, opened 2015-06-24
[RSP] Use del more often in the rest of the ops (except VMACQ).

cxd4:


cxd4:
20% of this I do/did care to have merged, but not half of it, so mostly I don't care.

PR #528, opened 2015-06-27
shiren save fix

Nekokabu:


Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
```diff
@@ -2376,13 +2376,56 @@ bool CMipsMemoryVM::LB_NonMemory(DWORD PAddr, DWORD* Value, bool /*SignExtend*/)
...
+
+			switch (Address & 0x3)
+			{
+			case 0x0: Address += 3; break;
+			case 0x1: Address ++;   break;
+			case 0x2: Address --;   break;
+			case 0x3: Address -= 3; break;
+			default:
+				return false;
+			}
```
cxd4: Just personal suggestion but maybe instead of doing that switch
```c
switch (Address & 0x3)
{
case 0x0: Address += 3; break;
case 0x1: Address ++;   break;
case 0x2: Address --;   break;
case 0x3: Address -= 3; break;
default:
    return false;
}
```
I was thinking it could be more maintainable code without a switch by using this?
```c
Address ^= 0x000003;
```

Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
```diff
@@ -2393,18 +2436,59 @@ bool CMipsMemoryVM::LH_NonMemory(DWORD PAddr, DWORD* Value, bool/* SignExtend*/)
...
+
+			switch (Address & 0x3)
+			{
+			case 0x0: Address += 2; break;
+			case 0x2: Address -= 2; break;
+			case 0x1: 
+			case 0x3: 
+			default:
+				return false;
+			}
```
cxd4: same comment for this, except
```c
if (Addr & 1)
    return false;
Address ^= 2;
```

Nekokabu:
I think that Shiren is not using  SB,SH/LB,LH, when the SRAM access.
and fix fix DMA Sram access when not word aligned.

PR #529, opened 2015-06-30
MemoryFilter support for 64-bit

cxd4:
Whether porting to one thing or entirely another, this horrible function needs to somehow be made into something slightly less horrible.  I don't know why a function of this type of design should even exist in any emulator at all; therefore I don't fully understand all of its design and propose simply the idea until it's really taken care of.

So if this micro step to supporting this feature in 64-bit is too big or too small to be proper, then what is the proper way to change it?


cxd4:
Closing!  Do not merge, never mind.

I thought this was causing 64-bit builds to not work but actually it was the TLB memory map issue I solved in my other pull request I filed alongside this one.

So MemoryFilter is never reached in our test cases, so rewriting it for 64-bit is not required for games to work in 64-bit Project64.  Please ignore this pull request and just try to merge the other one.

PR #530, opened 2015-06-30
Fix constant run-time check failures in MSVC debooger.

cxd4:
Before these changes, I can't step through anything in MSVC's debugger because of thousands of loop iterations of the following code in 64-bit:
```c
void CMipsMemoryVM::Reset( bool /*EraseMemory*/ )
{
 	if (m_TLB_ReadMap)
 	{
		memset(m_TLB_ReadMap,0,(0xFFFFF * sizeof(DWORD)));
		memset(m_TLB_WriteMap,0,(0xFFFFF * sizeof(DWORD)));
		for (DWORD address = 0x80000000; address < 0xC0000000; address += 0x1000) 
 		{
			m_TLB_ReadMap[address >> 12] = ((DWORD)m_RDRAM + (address & 0x1FFFFFFF)) - address;
			m_TLB_WriteMap[address >> 12] = ((DWORD)m_RDRAM + (address & 0x1FFFFFFF)) - address;
 		}
```
which makes this repeated (but unsurprising) complaint:
`Run-Time Check Failure #1 - A cast to a smaller data type has caused a loss of data.  If this was intentional, you should mask the source of the cast with the appropriate bitmask.  For example:  
char c = (i & 0xFF);`

All I had to do was fix the TLB array sizes from 32- to 64-bit (in _WIN64 builds only) to silence the run-time check failures so that I can continue stepping without the constant interruptions.
```c
memset(m_TLB_ReadMap , 0, (0xFFFFF * sizeof(size_t)));
memset(m_TLB_WriteMap, 0, (0xFFFFF * sizeof(size_t)));
for (address = 0x80000000UL; address < 0xC0000000UL; address += 0x1000) 
{
    m_TLB_ReadMap[address >> 12]  = ((size_t)m_RDRAM + (address & 0x1FFFFFFF)) - address;
    m_TLB_WriteMap[address >> 12] = ((size_t)m_RDRAM + (address & 0x1FFFFFFF)) - address;
}
```

Please note that it is not a very smart idea to omit the "UL" at the end of the loop header delimiters.  When it comes to suddenly having Project64 in 64-bit we are in uncharted territory and there is no reason to omit the UL and cause the compiler to think we really mean
```c
for (address = (__int32)-0x80000000; address < (__int32)0xC0000000; address += 0x1000)
```

### DIFF
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
@@ -48,26 +48,29 @@ CMipsMemoryVM::~CMipsMemoryVM()
 
 void CMipsMemoryVM::Reset( bool /*EraseMemory*/ )
 {
+	size_t address;
+
 	if (m_TLB_ReadMap)
 	{
-		memset(m_TLB_ReadMap,0,(0xFFFFF * sizeof(DWORD)));
-		memset(m_TLB_WriteMap,0,(0xFFFFF * sizeof(DWORD)));
-		for (DWORD address = 0x80000000; address < 0xC0000000; address += 0x1000) 
+		memset(m_TLB_ReadMap , 0x00, 0xFFFFF * sizeof(size_t));
+		memset(m_TLB_WriteMap, 0x00, 0xFFFFF * sizeof(size_t));
+		for (address = 0x80000000UL; address < 0xC0000000UL; address += 0x1000)
 		{
-			m_TLB_ReadMap[address >> 12] = ((DWORD)m_RDRAM + (address & 0x1FFFFFFF)) - address;
-			m_TLB_WriteMap[address >> 12] = ((DWORD)m_RDRAM + (address & 0x1FFFFFFF)) - address;
+			m_TLB_ReadMap[address >> 12]  = ((size_t)m_RDRAM + (address & 0x1FFFFFFF)) - address;
+			m_TLB_WriteMap[address >> 12] = ((size_t)m_RDRAM + (address & 0x1FFFFFFF)) - address;
 		}
 		
 		if (g_Settings->LoadDword(Rdb_TLB_VAddrStart) != 0)
 		{
-			DWORD Start = g_Settings->LoadDword(Rdb_TLB_VAddrStart); //0x7F000000;
-			DWORD Len   = g_Settings->LoadDword(Rdb_TLB_VAddrLen);   //0x01000000;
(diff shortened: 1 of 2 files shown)


cxd4:
@project64 : @LegendOfDragoon has confirmed that with these changes, several ROMs (Super Smash Bros., Absolute Crap 2) now work in 64-bit Project64.  Testing has shown that this pull request will get emulation working in 64-bit Project64; please state any objections you might have to these changes so that we can finally start playing games in 64-bit emu.

AmbientMalice:
This commit seems to break PJ64. At least in the 32 bit builds. (Which it shouldn't.) No clue why.

cxd4:
Hm oops can you describe what's broken please?

I can try to figure it out shortly but an initial sense of direction could help.

AmbientMalice:
Games won't boot. Can't recall the error, but it was very generic.

cxd4:
Gah.  I figured it out.

It's a recompiler-only regression that happens in the breakpoint I added to TranslateVAddr:
https://github.com/project64/project64/pull/530/files#diff-76ac10927d20b69a79bbb614ea16f4f4R545
```c
#if defined(_WIN64) || defined(_M_IX86)
    g_Notify -> BreakPoint(__FILEW__, __LINE__);
#endif
```

It's because I said _M_IX86, which means 32-bit...the breakpoint was meant only for 64-bit.

A macro typo on my part.

cxd4:
k, fixed in that second commit I just pushed^

I still feel like I'd rather someday, re-do this PR into smaller PRs anyway.  Much of the changes in this pull request were little whitespace changes, but only as a means of highlighting all the TLB array references I found through VS2013.  (Highlighting them in the commit meant nit-picking white space changes.)  Anyway when I feel like it I can re-do this PR into smaller changes.

But my changes are not wrong, for 64-bit.  All this PR really does is change the allocation of the TLB memory addr array from `DWORD` to `size_t`, which is the hardest change to confirm is safe.

project64:
looks ok, but can no longer merge from conflicts

cxd4:
The merge conflicts are on purpose.

I assumed since you haven't merged this for months that it was too complicated for you to agree on merging, so I ended up having you merge this PR instead:  https://github.com/project64/project64/pull/564

As a result, I basically caused the merge conflict intentionally.

cxd4:
I leave this pull request open to continue to reference it to continue dividing it into smaller, sub-PR's.  It is easier to find this pull request when it's in the open PRs list, not the closed PRs list, so that's why I haven't closed this yet.

I will close it when the rest of the changes are all split to other PRs and merged.

PR #534, opened 2015-07-11
Update AI Shougi 3 RDB

Nekokabu:
change Status to Compatible


cxd4:
Oh, yeah.  This was fixed a while back by handling 16-bit-unaligned PI DMA addresses.

PR #535, opened 2015-07-11
Fixed Audio Timing Overhaul.

AmbientMalice:
I'm hesitant to make such a drastic PR, but figure that this needs to be done eventually, and since the FAT save state bug has been fixed, most of my personal reservations are gone.

The basic idea of this PR is to make dynamic fixed audio timing the default behavior, with the understanding that some games may not like that, so they'll need to be tested and tweaked later. Sync audio has been disabled by default and FAT has been enabled. AI has been set to 0 by default.

The only major problem I can see is that Azimer currently doesn't handle dynamic FAT as well as Jabo does. I'm getting some intermittent crackling from Azimer's. Not sure why.

On the topic of Azimer's, World Driver Championship sounds very good with FAT disabled using Azimer's, so I disabled FAT there.


LegendOfDragoon:
I think disabling Sync Audio by default and setting AI count per bytes to 0 by default are good changes, but I'm not too sure about setting fixed audio timing on by default. At least half the games I play are better with FAT off (SSB, WDC, Rogue Squadron, Turok, etc).

I hope we'll have enough testers to turn it off for all the games that are worse with it on.

AmbientMalice:
>At least half the games I play are better with FAT off (SSB, WDC, Rogue Squadron, Turok, etc).

Could you explain the shortcomings? WDC is a unique case.

Rogue Squadron needs FAT to fix its audio juddering. I'm not aware of FAT causing issues.

Also, FAT fixes a myriad of subtle and not-so-subtle timing problems. Turok 3 has slightly off time animations without it. It also cures errors in the DD Sim City game.

Also, any N64 game with multiplayer must be be FAT-enabled for the purposes of netplay. This is where Smash Bros becomes a massive problem.

LegendOfDragoon:
>Could you explain the shortcomings?

Half the time I hear an unusual amount of crackling, I go and check my settings to see what I did wrong and find out that I left FAT on. One good example is Body Harvest. I don't know of an audio plugin that plays the audio perfectly for that game, but it's still much better with FAT off.

>Rogue Squadron needs FAT to fix its audio juddering. I'm not aware of FAT causing issues.

In game, the VI/s drops below 60 if using any sort of audio sync + FAT. Without any audio sync, the crackling is awful, with FAT enabled. Pretty sure with Azimer's XA 2 WIP 5, it sounds better overall with FAT off, in that game.

>Also, FAT fixes a myriad of subtle and not-so-subtle timing problems. Turok 3 has slightly off time animations without it. It also cures errors in the DD Sim City game.

Hmm, that's interesting. I'll have to try testing Turok 3. Where exactly does this happen, where it's noticeable?

>Also, any N64 game with multiplayer must be be FAT-enabled for the purposes of netplay.

I understand this concern. I heard the AQZ netplay plugin doesn't even work in 2.2 though and it's closed source ;/ .

AmbientMalice:
>Hmm, that's interesting. I'll have to try testing Turok 3. Where exactly does this happen, where it's noticeable?

Start a new game. Walk over to the connecting walkway. A craft will crash through it, before disappearing. Without FAT, its afterburners can be seen through the road below. With FAT, the craft is properly concealed from view.

>Pretty sure with Azimer's XA 2 WIP 5, it sounds better overall with FAT off, in that game.

As I mentioned, I've been hearing crackling across the board with Azimer's using FAT. Resident Evil 2 is a good example. That game needs FAT. But with Azimer's, it now has crackling. Yet Jabo's audio is perfect. So I'm wondering whether the problem is on Azimer's end or PJ64's end.

>I heard the AQZ netplay plugin doesn't even work in 2.2 though and it's closed source

The curse of closed source. :( I thought AQZ stopped working because FAT was disabled across the board. But maybe something else broke.

AmbientMalice:
@LegendOfDragoon I'm thinking that Azimer's has some problems. Possibly regressions. The XA2 version sounds absolutely awful with Rogue Squadron regardless of settings.

>In game, the VI/s drops below 60 if using any sort of audio sync + FAT.

Why would you be using audio sync with fat? Using audio sync+FAT makes every game run at the wrong speed, AFAIK.

LegendOfDragoon:
> I'm thinking that Azimer's has some problems. Possibly regressions. The XA2 version sounds absolutely awful with Rogue Squadron regardless of settings.

It certainly does have regressions. The XA2 plugin was more likely to crash on 1964, last time i tried, compared to the version from the start of the repository. I still need to do further testing in Rogue Squadron to reduce crackling ;/ .

>Why would you be using audio sync with fat? Using audio sync+FAT makes every game run at the wrong speed, AFAIK.

When I say audio sync, i mean through the plugin either "Prevent Buffer Overruns" or Jabo's "Sync game to audio". If it makes the game run at the wrong speed, then that means there's something wrong with the audio. I rarely ever use the emulator's built-in audio sync.

>I've been having trouble finding games that are reliably "sounds okay without FAT, but FAT causes problems". You mentioned Body Harvest, which actually sounds pretty crappy if you listen closely to non-FAT audio, regardless of plugin. I can't get Resident Evil 2 to reliably act up. (That problem seems exclusive to Azimer's, anyway.)

Perhaps it's just a coincidence that many of the games I play, sound better with FAT off. Just tried OOT and that's another example. Body Harvest may sound bad, but the improvement from turning off FAT, was enough to help me confirm that Azimer's newer HLE code has improved for Body Harvest.

I'll see about doing more audio testing this week.

AmbientMalice:
>Just tried OOT and that's another example.

I just tested OOT.

Jabo has bad crackling without FAT. Minor crackling with FAT.
Azimer DS8 has significant crackling without FAT. Minor, intermittent pops with FAT.
Azimer XA2 sounds more or less the same with or without FAT.

>When I say audio sync, i mean through the plugin either "Prevent Buffer Overruns" or Jabo's "Sync game to audio".

I see what you mean. The reason I asked is Jabo's Sync Game to Audio causes speed problems in a number of games, whereas Azimer's does not.

LegendOfDragoon:
>Jabo has bad crackling without FAT. Minor crackling with FAT.
Azimer DS8 has significant crackling without FAT. Minor, intermittent pops with FAT.
Azimer XA2 sounds more or less the same with or without FAT. (Okayish.)

I usually just test Azimer XA2 and Jabo 1.7 these days. I still may try experimenting with 1964 Audio (it has potential). For me, Jabo's was good if using its audio sync & FAT disabled for OOT (just tested intro really). Azimer's XA2 crackles less for me with FAT disabled for OOT.

I do know that different hardware can play a role in people's experiences with audio. For instance, @theboy181 has less crackling than I do, on his hardware with shunyuan's plugin in SSB (FAT off, of course).

>I see what you mean. The reason I asked is Jabo's Sync Game to Audio causes speed problems in a number of games, whereas Azimer's does not.

Just a guess, but it's likely that Azimer's buffer size is bigger. With FAT, you need to use a larger buffer size. Which is one of the reasons I'm not a huge fan of FAT.

>Note - I tested all with LLE.

Same here. LLE ftw :smile: .

AmbientMalice:
>For me, Jabo's was good if using its audio sync & FAT disabled for OOT (just tested intro really)

What about the CPU stalls?

>I do know that different hardware can play a role in people's experiences with audio.

And there lies a massive problem. Look at Desmume and its "crackling for everyone, so choose one of several different sound sync methods" design.

LegendOfDragoon:
>What about the CPU stalls?

Yes it stalls, unfortunately.

>Is there any alternative to FAT in the long run to fix speed and general timing issues? (And also allow 100% perfect netplay for those who want it?) I know FAT has its own host of problems, but the current framework is pretty bad.

Good AI emulation fixes speed and timing issues, for the most part anyway. I'm mostly just uncertain about select few games (like Top Gear Rally). As for perfect netplay, FAT is a requirement.

The problem with audio, is that its results are affected by hardware. I don't think you can perfectly emulate it in real-time. It seems that different games use different formulas for audio, so maybe FAT can be good enough if you have multiple algorithms.

I still haven't tested Turok 3, and Sim City. I'll try testing those soon.

dsx-:
man, what did you guys break

FAT ON and AI=0 is supposed to have perfect audio in oot with jabo's plugin (at least it does in this emucr build)
in 2.2.0.3rc1 its crackle city

emucr build wins :P

cxd4:
>emucr build wins :P

That may probably be because they built it themselves, unlike u1!11@1123!@#@@#

cxd4:
re the rest:  didn't read

dsx-:
important point i forgot to mention is that i was using an emucr build from May
ill keep using it until the changes to later builds get fixed

dsx-:
never mind im an idiot, 2.2.0.3rc1 is ancient

dsx-:
anyway i dont get crackling in oot with fat on, ai=0, even with jabo's

LegendOfDragoon:
>anyway i dont get crackling in oot with fat on, ai=0, even with jabo's

Must be either due to either different hardware or different settings.

LegendOfDragoon:
@AmbientMalice 
>Azimer XA2 sounds more or less the same with or without FAT. (Okayish.)

How recent was the build that you tested? Just compiled latest XA2 source today and there are some serious regressions with FAT off. Rogue Squadron sounds messed up, more games have significant crackling, and WDC has no sound.

AmbientMalice:
>How recent was the build that you tested?

A few days.

Just compiled latest XA2 source today and there are some serious regressions with FAT off. Rogue Squadron sounds messed up...

I mentioned earlier that "The XA2 version sounds absolutely awful with Rogue Squadron regardless of settings."

>and WDC has no sound.

WDC has sound for me.

LegendOfDragoon:
>I mentioned earlier that "The XA2 version sounds absolutely awful with Rogue Squadron regardless of settings."

I'm getting weird results. I'll have to re-test FAT for Rogue, another day. Certainly WIP6 is significantly worse than original WIP 5, with FAT disabled, for XA2, for multiple games.

>WDC has sound for me.

I have no sound in WDC with FAT off, in XA2 WIP 6.

theboy181:
No sound here too..

AmbientMalice:
@LegendOfDragoon @theboy181 A very recent commit broke mingw support and broke all the audio, basically. This seems like a likely culprit. https://github.com/Azimer/AziAudio/commit/040884772521efa7095bc3f9381eb233e29e1ccb

Azimer:
I do not always follow Project64's issues or comments (there's a lot of noise at times).  If there's an issue again with AziAudio where I mistakenly forgot to modify the makefile or project for mingw or vs2008, please comment on my commit or open an issue and tell me to fix it.  I almost exclusively use vs2013 for all of my builds.  Thanks!! :D

Frank-74:
Messing around with settings with DK64 and found these get the rap in sync the best.

Counter Factor=2
AI count per byte=0
Fixed audio timing=off
Sync using audio=off

And set Jabo's to sync game to audio.

AI, fixed audio and sync using audio set to 0 would be best as the default.

AmbientMalice:
>And set Jabo's to sync game to audio.

That is a terrible idea, IMHO. For what it's worth, Jabo's audio is ALREADY currently forced to sync audio in the RDB, but Jabo's audio sync causes stalling and other issues. I was never happy with enabling it, but I did it as a temp measure. That was one of the reasons I was so supportive of scrapping Jabo's and moving to Azimer's with FAT.

>Counter Factor=2

It's important to note that CF was set (by me) to 1 in order to alleviate dong. (Polygon tearing for those not in on the meme.) It's especially needed if you're going to use Jabo's audio sync.

>AI count per byte=0
>Fixed audio timing=off

This is kinda meaningless. AI count per byte has no effect without FAT.

>Sync using audio=off

No real problem there. PJ64's sync using audio is kinda useless. I think it should probably be removed.

Frank-74:
> This is kinda meaningless. AI count per byte has no effect without FAT.

Wrong.  AI count is set dynamically whether FAT is on or off.

This is the line where AI count is set in Audio messages debugger output with FAT off and AI=0.

> : CAudio::LenChanged: Set Timer  AI_LEN_REG: 1472 m_CountsPerByte: 591

AmbientMalice:
>Wrong. AI=0 calculates the AI count per bytes value if FAT is on or off. Adjustment of the AI value can be done away with altogether. Its not needed as auto calculation is better.

I was under the impression that without fixed audio timing, aicountsperbyte is meaningless because the emulator isn't using it to control audio timing. Maybe I'm wrong.

>Limit FPS setting is responsible for causing the crackling in audio i've found.

That stands to reason, though. The games MUST be locked at their correct refresh rate. This means sound has to be buffered to compensate. If you disable the framerate lock, the game just sends the audio as fast as it can.

This is the same problem that PJ64 2.1 had -- by using PJ64 audio sync and FAT together, which was a ridiculously bad decision, almost all games ran at the wrong framerates, but the audio was smooth.

>Castlevania is a good example. AI=0, FAT off, sync using audio off, sync on in Jabo...

But the problem is that Jabo's audio sync is horrible. In multiple cases it gets game speed completely wrong, and it causes constant stalling. The audio is going to be smooth if you disable the FPS lock, but in most cases, Jabo's audio sync does a fairly dismal job of actually syncing audio. I mean no disrespect to Jabo. He was just working with what he had.

Maybe my ears are biased, but I just ran Castlevania 64 with FAT+AI=0 using Azimer's and the audio was just peachy. Even turning over the Prevent Buffer Overflow setting didn't make the audio crackly. Jabo's has some serious problems. "Correct" audio handling is going to break on Jabo's, IMO. That's just how the cookie crumbles.

LegendOfDragoon:
>I was under the impression that without fixed audio timing, aicountsperbyte is meaningless because the emulator isn't using it to control audio timing.

You are correct here.

>But the problem is that Jabo's audio sync is horrible. In multiple cases it gets game speed completely wrong, and it causes constant stalling. The audio is going to be smooth if you disable the FPS lock, but in most cases, Jabo's audio sync does a fairly dismal job of actually syncing audio. I mean no disrespect to Jabo. He was just working with what he had.

Imo, audio sync shouldn't be relied on, in general. I'm able to get good audio in most games w/o using it, with FAT = off.

>Maybe my ears are biased, but I just ran Castlevania 64 with FAT+AI=0 using Azimer's and the audio was just peachy.

It's not just you. It was pretty good on my end too.

Frank-74:
Azimer's plugins drift out of time, video/audio sync gets progressively worse with time.
Try FAT=0 with Goldeneye.  Goto the mission select screen and leave it there for 10 minutes or so, don't pause or open options, just let it run.  Then after 10 minutes, press B then A, to go back and forward screens.  The audio on pressing A/B (page turn/door close) will now be delayed over a second behind the video.  This never happens with Jabo's dsound, no matter if sync audio is on or off.

AmbientMalice:
>Azimer's plugins drift out of time, video/audio sync gets progressively worse with time.
>Try FAT=0 with Goldeneye. Goto the mission select screen and leave it there for 10 minutes or so, don't pause or open options, just let it run. Then after 10 minutes, press B then A, to go back and forward screens. The audio on pressing A/B (page turn/door close) will now be delayed over a second behind the video.
The thing is -- bugs with Azimer's can be fixed. Bugs in Jabo's can't. I am aware that Azimer's sometimes goes out of sync. I figure this sort of thing can be fixed. Doesn't Azimer's have two different buffer modes? But I will test GE like you said.

Frank-74:
Also, the latest WIP6 is terrible with FAT off, the audio constantly speeds up and slows down in GE.  DS8 WIP5 is better for FAT off.

Here, with FAT on, AI=0, Goldeneye gets throttled down to 40-45 VI's in the intro when bond walks onto the screen.  FAT off and its constant 60VI's, but with speed problem in WIP6.  My older WIP5 build doesn't have the speed problem.

DS8 has two buffer modes, its defined in DirectSoundDriver.cpp.  #define STREAM_DMA.  I've always commented it out, as standard buffer is better.  //#define STREAM_DMA.

Frank-74:
@AmbientMalice 
I think I found the sync issue last night in the DS8 plugin.  Seems that force old audio sync is what causes the delay, no matter if you enable/disable it in the options.  PBO makes no difference.  The delay happens quicker if you turn on force old audio sync, takes a LOT longer with it off.  Just press A/B a few times and you'll notice the delay getting longer with force old audio sync on.

I'll have a go at fixing it later tonight.

purplemarshmallow:
Without FAT the emulator is really unstable if you disable the frame limiter. 
In Turok Dinosaur Hunter if I disable the frame limiter I get: Fatal Error: Stopping emulation
FAT fixes these problems with disabled frame limiter I think this should be merged

oddMLan:
I also vote this commit as a candidate to get merged before the release of Project64 2.3.
FAT brings almost only pluses and little to no cons.

The reason it was disabled before was that it made graphics and framerate inconsistent. With the improvements Azimer made to it, that is no longer relevant.

I propose as global settings
FAT **On**
AI = **0**
Sync using audio **Off**

I manually set these for the games I play, and I found it improves the audio a lot, like, absolutely no crackling or popping of any kind.

Just a nitpick I have, I noticed that with FAT enabled, the VI counter no longer gets rounded.
I mean, you see numbers such as 60.04 VI/s, 59.97 VI/S, 57.94 VI/s, etc.
A rounding function should be implemented when having FAT enabled so we have numbers like 60 VI/s, 60 VI/s and 58 VI/s. The reason a rounding was first implemented in Project64 I'm not aware of it, but it does make look the UI look better. Enabling FAT would defeat that original purpose, whatever was that. So that's why I suggest enabling or implementing VI/s rounding with FAT enabled.

Frank-74:
> I noticed that with FAT enabled, the VI counter no longer gets rounded.
I mean, you see numbers such as 60.04 VI/s, 59.97 VI/S, 57.94 VI/s, etc.

Only happens if you turn off Limit FPS (F4), and have the audio plugin doing the sync.

oddMLan:
My bad, I had Sync using Audio enabled for some reason.
FAT + Sync using Audio = VI counter fluctuation
just FAT = no VI counter fluctuation

AmbientMalice:
There's still the problem of Azimer's desycing after extended periods. Jabo's doesn't suffer from this issue, however.

oddMLan:
Then I think in the meantime it would be wise to still keep Jabo's Audio plugin as the default.
Wouldn't hurt to bundle Azimer's audio as an alt plugin, though.

Frank-74:
@oddMLan 
I've just made a branch of the last WIP5 version of Azimer's 0.7.  Undefined the bugged STREAM_DMA option in DS8, and fixed the nosound on switching plugins to XA2.
https://github.com/Frank-74/AziAudio/tree/Last-WIP5

It builds with XA2 by default in VS.  To build DS8, you need to add 'XAUDIO_LIBRARIES_UNAVAILABLE' to the Preprocessor definitions in AziAudio > Properties > C++ > Preprocessor.

WIP6 is half done adding support for shared buffer type between API's or something.  Its not finished and goes out of sync/has speed problems with FAT off in PJ64.

cxd4:
I'm not sure whether I'd call it "bugged", so much as just "incomplete".

Azimer said he still has to work more things on to complete this experiment.  He knows more about the RCP AI internals than any of us here, so there's no reason to think his implementation is at fault, let alone finished.  So there is plenty of time for that to change.  Shunyuan just copied the typical AI code written in the "normal" way and got lucky.

oddMLan:
Thanks for that @Frank-74 , I've compiled both versions now

LegendOfDragoon:
>Undefined the bugged STREAM_DMA option in DS8, and fixed the nosound on switching plugins to XA2.

STREAM_DMA works good with Top Gear Rally and WDC (prolly stunt racer too). IIRC Azimer wants to eventually make that an option, so that you can simply enable it for games that benefit from it. I think I may try doing some more experimenting with audio. @Frank-74 would you be interested in testing and giving feedback, if I get around to tweaking code?

AmbientMalice:
@LegendOfDragoon How does STREAM_DMA work with Ridge Racer 64 using FAT? Because currently Ridge Racer 64 hitches every second or so using Jabo's and garbles every second or so using Azimer's. Interestingly, Ridge Racer is a game that runs far too slow if Audio Sync is used.

Frank-74:
@LegendOfDragoon Yes I'll test your tweaked code.
@AmbientMalice Ridge Racer is fine here with Jabo and Azimer's 0.7, FAT on or off.  Make sure to set CF to 1.

LegendOfDragoon:
>How does STREAM_DMA work with Ridge Racer 64 using FAT?

Not sure if STREAM_DMA makes a difference with FAT enabled, but it ran good with Ridge Racer 64 with CF = 1 as @Frank-74 recommended.

>Yes I'll test your tweaked code.

Ok, i improved STREAM_DMA for FAT=off. https://github.com/LegendOfDragoon/AziAudio/tree/Experiment

@Frank-74 Let me know your experience with it, with FAT = off. I was going to do more stuff like try to fix save state issue with the XA2 version, but it requires too many changes ;/ .

Frank-74:
@LegendOfDragoon Seems good to me, tested with several games.

Frank-74:
The improved DS8 STREAM_DMA with FAT off works well with Cruis'n World.  Fixes the too fast audio issue.  https://github.com/Azimer/AziAudio/issues/4.  XA2 still has the speed problem.

LegendOfDragoon:
>@LegendOfDragoon Seems good to me, tested with several games. The improved DS8 STREAM_DMA with FAT off works well with Cruis'n World. Fixes the too fast audio issue. Azimer/AziAudio#4. XA2 still has the speed problem.

Thanks for the feedback. Good to know that it's good on not just my setup :smile: . How's Goldeneye with it?

AmbientMalice:
Closed in order to rework as series of smaller PRs.

PR #536, opened 2015-07-11
Make Glide64 default video plugin.

AmbientMalice:
This is a rather tricky PR request. I don't expect it to be merged straight away, if at all, since changing the plugin setting is only the first step. In my humble opinion, Glide64 should be the default plugin. And it should be treated as the compatibility baseline, especially in the RDB. So if @project64 approves of this change, I would like to remove all the game descriptions such as "bad graphics in menus - use Glide64". -- since the default would now be Glide64.

In the absence of a truly decent LLE plugin like z64 or the software accurate plugin or whatever, it may be necessary, or at least prudent, to set Jabo's as the default plugin on a per-game basis for those games that require LLE graphics.

### DIFF
--- a/Source/Project64/Settings/Settings Class.cpp
+++ b/Source/Project64/Settings/Settings Class.cpp
@@ -311,7 +311,7 @@ void CSettings::AddHowToHandleSetting ()
 
 	//Plugin
 	AddHandler(Plugin_RSP_Current,   new CSettingTypeApplication("Plugin","RSP Dll",       "RSP\\RSP 1.7.dll"));
-	AddHandler(Plugin_GFX_Current,   new CSettingTypeApplication("Plugin","Graphics Dll",  "GFX\\Jabo_Direct3D8.dll"));
+	AddHandler(Plugin_GFX_Current,   new CSettingTypeApplication("Plugin","Graphics Dll",  "GFX\\PJ64Glide64.dll"));
 	AddHandler(Plugin_AUDIO_Current, new CSettingTypeApplication("Plugin","Audio Dll",     "Audio\\Jabo_Dsound.dll"));
 	AddHandler(Plugin_CONT_Current,  new CSettingTypeApplication("Plugin","Controller Dll","Input\\PJ64_NRage.dll"));
 


cxd4:
Usually I find Jabo's plugin gives more stable graphics, supports more things than Glide64.  Though that may just be the limited range of games I grew up playing.

project64:
While I would like to merge this I do not know if glide64 is better in general more for speed and compatibility.

AmbientMalice:
Glide64 is somewhat slower, but has much higher compatibility. (I'm hesitant to say "accuracy" since Glide64 has a shitload of hacks under the hood.)

The biggest advantage is framebuffer support. Jabo's is too slow. Banjo Kazooie/Tooie need per-frame framebuffering for transitions and even some puzzles. With Glide64, per-frame FB is too slow in some games. But it can be easily turned on or off with ALT B.

Resident Evil 2 needs Glide64. In fact, Glide64 and angrylion's software are the only plugins that play it properly. The game is hideous with Jabo's.

Rayman 2 has missing transitions and worse, wrong coloured lums on Jabo's. The list goes on for many games.

There are some games where Glide64 is lacking. For example, Toy Story 2 - the weapon charge bar doesn't fill. GLideN64 fixes this with a special depth mode. My guess is Jabo's uses a cheap and dirty depth hack to fix the UI elements being out-of-order. Glide64 also a rainbow-colour texture corruption bug that occasionally turns up. Perfect Dark and Winback come to mind. But in Perfect Dark's case, Jabo's has no lighting. And it has no lens flares. And the sky is often missing. (Glide64 *does* have a problem where the combat simulator needs FB disabled.) In such cases, Glide64 comes off much better.

But overall, Glide64 is far more compatible with a slight speed trade-off.

cxd4:
You should test more often the "Software rendering" feature in Jabo's plugins; the Direct3D6 1.5.2 plugin has a similar feature to that for debugging RSP issues versus Direct3D issues.

cxd4:
As for speed, Glide64 last I heard as of the GLES/Mupen64Plus portable fork makes over 10 times the number of OpenGL calls than glN64 or RiceVideo.  It's slower due to the layers and amounts of OpenGL calls, but not exactly as a matter of what happens outside the API or just not being DirectX.

AmbientMalice:
>You should test more often the "Software rendering" feature in Jabo's plugins

I test it often. Too buggy to be useful, unfortunately. It fixes the camera detection in Snap, but breaks a lot more. Also, it unfortunately doesn't solve the problem of Jabo's unusably slow framebuffer.

cxd4:
It's not supposed to be on at the same time as frame buffer options since software rendering already implies that almost certainly, so that's a redundant combination of settings

Software rendering has never broken anything for me other than all 3-D triangles in LLE graphics, which suggests that only the information processed by high-level dlists is sourced for the rendering.

Lithium64:
I agree on make Glide64 the default video plugin it has the best compatibility, it is slower but can be optimized, this plugin has many memory leaks that can be fixed.

LegendOfDragoon:
Glide64 is has better HLE compatibility, but significantly worse speed.

>it is slower but can be optimized

I doubt anyone will bother significantly optimizing Glide64.

AmbientMalice:
>I doubt anyone will bother significantly optimizing Glide64.

Have you taken a look at libretro's fork? https://github.com/libretro/mupen64plus-libretro They've been tinkering with Glide64 and Glitch64 over the past year or so. But I dunno whether the improvements amount to much.

This caught my eye - https://github.com/libretro/mupen64plus-libretro/commit/de4150f9cc2da3cb78b6da6c52ff255c1efb968c A vertex buffer added to the wrapper? Does PJ64's have one? I don't think so. Does it have any real benefits?

>I'd be more in favor of using Glide64 as default plugin, if it was ported to OpenGL instead of relying on some wrapper.

That would be ideal. TBH, I'm puzzled as to why Gonetz decided to use gl64 instead of porting Glide64 to OpenGL, considering what a nightmare gl64 has proven to be.

I'm still enamoured by the idea of integrating z64 into Glide64 to create a fast, versatile HLE/LLE plugin. Gonetz integrated z64 into GLideN64, but he did a dismal job because it is worse than z64 in almost every way.

mudlord:
What would be nicer is a new plugin from scratch.
But sadly its impossible to have nice things and so we must constantly drink from the poison well of bitterness and hate.

AmbientMalice:
>What would be nicer is a new plugin from scratch.

Agreed 100%. Beautiful, clean code. Preferably LLE as much as possible. Avoid per-game hacks as much as possible. (N64 games lie. They lie about their ucodes. They lie about their texture formats. They lie about their resolutions. It's no wonder HLE graphics emulation is so bloated.)

>But sadly its impossible to have nice things and so we must constantly drink from the poison well of bitterness and hate.

The reverse engineered ucodes that underpin HLE emulation need to go die in a fire. They've been a poison pill in N64 emulators since the turn of the century. N64 graphics emulation has been afflicted by a fatal "oh, well, it's good enough so no need to reinvent the wheel" mentality.

LegendOfDragoon:
>Have you taken a look at libretro's fork? https://github.com/libretro/mupen64plus-libretro They've been tinkering with Glide64 and Glitch64 over the past year or so. But I dunno whether the improvements amount to much.

Briefly. If I understand correctly, they actually did port Glide to OpenGL. And they ported it to C :smile: . I just couldn't be bothered porting that code to zilmar spec. I'd be happy if someone else decided to port it though.

>What would be nicer is a new plugin from scratch.
But sadly its impossible to have nice things and so we must constantly drink from the poison well of bitterness and hate.

Agreed. At this point, I'm better off writing a new plugin from scratch, than maintaining an existing one. I personally like to nitpick and reinvent the wheel. I have benefited greatly from doing so, with N64 emulation :smile: .

cxd4:
> TBH, I'm puzzled as to why Gonetz decided to use gl64 instead of porting Glide64 to OpenGL, considering what a nightmare gl64 has proven to be.

Well why do you say that?  How is glN64 codebase any more of a nightmare than Glide64 codebase?  That's probably why Gonetz chose it in the first place--he knew it was done in cleaner C++ than the C++ used in Glide64.  (In fact I think that was just what he complemented on EmuTalk if I remember.)

> This caught my eye - libretro/mupen64plus-libretro@de4150f A vertex buffer added to the wrapper? Does PJ64's have one? I don't think so. Does it have any real benefits?

That's because in libretro they use OpenGL ES 2+ for portability to things like iOS and Android.  Those same features ironically require OpenGL 2.0 or later to do what you linked in a non-ES counterpart, so somebody with an OpenGL 1.5 video card would not necessarily be able to run OpenGL ES code.

My personal preference is to default to compatible vertex buffers--for example, you should never (for any reason) say `glVertex4f(x, y, z, w)` when you could say, for size as well as performance, `glVertex4fv(&coords[0])`.  Also, it's virtually impossible to find even a theoretical implementation of OpenGL that doesn't support client-side vertex arrays--`glVertexArray` and `glDrawArrays` (or `glDrawElements` to optimize shared and re-used vertices in the GPU cache), which may for extreme cases be noticeably inferior in speed to server-side vertex arrays which require OpenGL ES 2+ or OpenGL 2.0+, or the relevant extensions.  I happen to have taken the time to have implemented both options (client- and server-side vertex arrays) in the angrylion OpenGL fork, but I have already done a few minor rewrites to Glitch64 in this repository to cache the vertex data in arrays in at least some form.

I've said it already, but the number of wasted OpenGL calls is the issue, not the type of calls or sole implementation of Glitch64 within glide.  Either RiceVideo or glN64 OpenGL is tremendously faster for that reason without even using the vertex caching you just linked to.

cxd4:
Which reminds me I'd like to spend my time bitching further yet again that the Khronos bridgeway between primitive and modern OpenGL functions is retarded.

In deprecated OpenGL it was glVertexPointer, glColorPointer, ....
Now it's glVertexAttribPointer(GL_VERTEX_ARRAY, ...), glVertexAttribPointer(GL_COLOR_ARRAY, ...).

Language-level differences like this show that they practically did the opposite of making any effort to compatibly introduce server- versus client-side vertex arrays into the current layout of OpenGL as an API.  So rather than get rid of 5 function entry points, they moved 5 to the "crap stack" which every sane video driver maker should feel obligated to support anyway AND added another unnecessary API point.

LegendOfDragoon:
>The reverse engineered ucodes that underpin HLE emulation need to go die in a fire. They've been a poison pill in N64 emulators since the turn of the century.

Honestly, I don't think HLE is all that bad :smile: . Not only is it fun to code HLE software, but it has its advantages too.

@cxd4 Do you know a good profiling tool for OpenGL? I'd love to start profiling Glide64, Rice Video, and glN64.

AmbientMalice:
>Honestly, I don't think HLE is all that bad.

To be clear, neither do I. The problem with N64 HLE graphics is there are very important missing effects that will likely never be fixed because reverse engineering and debugging ucodes is so much work.

LegendOfDragoon:
>To be clear, neither do I. The problem with N64 HLE graphics is there are very important missing effects that will likely never be fixed because reverse engineering and debugging ucodes is so much work.

The biggest problem imo is HW acceleration. You're right that reverse engineering & debugging ucodes is a lot of work, but so is optimizing LLE :smile: .

myownfriend:
I agree with ghost that a new plugin should be made from scratch. Though, I never understood why emulators opted to use plug-ins anyway. In theory, all of them are working toward the same goal so why not pool everyone's efforts into one really optimized backend that's as fast, accurate, and feature-rich as can be.

AmbientMalice:
>I agree with ghost that a new plugin should be made from scratch.

If wishes were horses, then beggars would ride.

>Though, I never understood why emulators opted to use plug-ins anyway. In theory, all of them are working toward the same goal so why not pool everyone's efforts into one really optimized backend that's as fast, accurate, and feature-rich as can be.

This isn't precisely true. Project 64 was originally supposed to be a cycle accurate N64 emulator. Mupen and 1964 were off doing their own thing. It made sense for them to conform to a shared plugin standard, especially since open source sharing wasn't really a "thing" back in 1999.

myownfriend:
Fair point but now that open source sharing is now a thing, wouldn't it make sense to move away from plug-ins? Having to maintain 3 major graphics plug-ins seems like it would be more work in the long run.

AmbientMalice:
>Having to maintain 3 major graphics plug-ins seems like it would be more work in the long run.

Yes and no. Practically speaking, the current N64 video plugin landscape consists of the following, excluding redundant stuff like Jabo and Rice:

1: Glide64 - a dead-end plugin that is kept around because it "kinda works" better than the competition when it comes to speed vs accuracy. Abandoned in 2012. Has some bad bugs.

2: Ziggy's LLE - a dead plugin untouched since 2007. Provides the best hardware accelerated LLE graphics. Code integrated into GLideN64.

3: GLideN64 - actively, if irregularly, developed plugin that combines code from gl64 (2004), Glide64 (2012), and Ziggy's LLE (2007.) Supports LLE & HLE, plus various N64 features unsupported by other hardware accelerated plugins - mipmapping, noise, Video Interface. Plugin suffers performance issues and some crippling bugs. Until these are fixed, Glide64 remains the top dog.

4: HatCat's Pixel Accurate Software Plugin - based on Angrylion code/MESS RDP code. Extremely slow. Updated now and again. LLE plugin.

Realistically, the only current options are to work on GLideN64 to make it acceptable, or create a NEW plugin.

cxd4:
>I agree with ghost that a new plugin should be made from scratch.

That "ghost" btw refers to a deleted GitHub account (specifically mudlord).

Everybody wants emulators/programs written from scratch, either inside or outside of emulators or N64.  It just sounds cooler that way--especially to end users who don't care about the open-source world of picking up where others left off.  There are a lot of accusations of "stolen code" though but those are really traceable more at the recent events starting 2012, UltraHLE as well but not the beginning since things like the plugin system, like many of the newcomer developers seem to suspect.

As for Re: plugins, explained it a bunch of times, but I notice on your profile it says you're involved with Dolphin so I don't think there's a point to me re-explaining it again.  They were fine with plugins, too, for like 99% of the total timespan that Dolphin has existed in.  People need to stop drawing their own conclusions about why they really got rid of plugins.  Now, if you want to use something else as an example--like RetroArch--about the whole "central backend" philosophy to ignore having any use for plugins, then that would make more sense.  But much of Dolphin has been developers spiting people involved with N64 or anyone not a part of their social little gang anyway, so if they couldn't use N64 as an example then I'm sure their explanation for dropping plugins would have been rather different.

LegendOfDragoon:
>Fair point but now that open source sharing is now a thing, wouldn't it make sense to move away from plug-ins? Having to maintain 3 major graphics plug-ins seems like it would be more work in the long run.

It's not like the same people are working on 3 different graphics plugins. In general, different people are working on different plugins.

>4: HatCat's Pixel Accurate Software Plugin - based on Angrylion code/MESS RDP code. Extremely slow. Updated now and again. LLE plugin.

If you prefer having filters on, Angrylion's actual plugin is good to use. Plus it has relatively recent improvements, like a bug fix for WDC.

>Realistically, the only current options are to work on GLideN64 to make it acceptable, or create a NEW plugin.

After studying and optimizing Angrylion's code, I'm starting to think a software plugin might be a good alternative. It's certainly a very good code base to learn from.

cxd4:
Nothing about the filters is changed between angrylion's "actual plugin" and the GL fork.

LegendOfDragoon:
>Nothing about the filters is changed between angrylion's "actual plugin" and the GL fork.

After taking a closer look, there's less of a difference in the vi filter code than I previously thought. The speed difference was mostly just the compiler.

cxd4:
Whatever.  No need to say you're wrong as though it's become surprising or something.

Just interested in more fundamentally meaningful projects now so have unintentionally put off continuing the collaboration with angrylion.  I also didn't change any of his MSVC-2002-based NOINLINE/FORCEINLINE compiler-specific speed hints for the actual VI filter functions that get called in the main loop, so none of that stuff is changed apart from maybe a couple instances of playing with SSE way back when.

myownfriend:
I wasn't suggesting that a centralized backend or a new plug-in be made from scratch because I thought it sounded cool or to spite anybody. I was just thinking it could be easier than trying to just merge multiple plug-ins, but I didn't know about GlideN64 until AmbientMalice mentioned it.

Also, from a end-user point of view I just generally prefer it when I don't have to try out different plug-ins in an emulator. That's why I brought it up, not because I consider myself to be a member of some Dolphin clique.

cxd4:
Wasn't trying to imply that you were doing either of those.

Just saying their reasons for abolishing plugins aren't along the lines of yours for doing it here.  So far it's something I've had to explain to absolutely everyone who has either used Dolphin as an example or has in any way been involved with it, but you could well be the first counterexample.  If so I'm relieved.

Re: centralized backend being easier than merging multiple plugins into one, actually RetroArch has made efforts to do both of those.  The centralized backend makes it possible to abstract API-specific code away from the plugins, find out what they have in common and make it into one plugin, -ish, so actually it isn't choosing between those two things, it's applyign both of them.

Re: prefer not to have to try out different plug-ins in emulators, I keep less of them installed when I only feel the need to test some so I don't have any need to try multiple plugins.

Do I think the plugin-based approach is superior?  No, I contemplate often a plugin-"free" approach as well.  But plugins by themselves is little to blame with the issues N64 hardware brings--social issues some people will say different, but often times it makes said people look better when they have something to blame.

vgturtle127:
I thought we were going to switch to GlideN64?

AmbientMalice:
>I thought we were going to switch to GlideN64?

In the long term, GLideN64 is arguably the best video plugin. But it currently has too many issues to be an official plugin, IMO. However, it is improving quite a bit.

It's an iffy trade-off. Switching to GLideN64 now would break a lot of games, but it might encourage people to find solutions. As a foundation, GLideN64 is vastly superior to Glide64 and infinitely superior to Jabo's.

vgturtle127:
@AmbientMalice *sigh*

Nintendo 64 emulation is so inaccurate and buggy it depresses me... :disappointed:

LasagnaPie:
People still using Voodoo Graphic cards ?

vgturtle127:
Since Glide64 and GLideN64 are the only real options to switch to, I would assume since Glide64 is recommended in the Project64 notes a lot that we should switch to that until GLideN64 has better compatibility? I don't know, I haven't finished testing but both Glide64 and GLideN64 have pretty good compatibility...

@LasagnaPie I'd hope not. Realistically, you shouldn't be running an OS older than Windows 7, and that would narrow the hardware you could have pretty significantly. At least in terms of legacy hardware.

As for hardware, I would think everybody by now would have to upgrade to at least decent 200x hardware, which means I'm not sure why anybody is still bothering to fight for things like XP support and pre-DX9 stuff. I hope they don't browse the internet on such old software, that would put you at risk just using the web browser let alone browsing the web normally.

AmbientMalice:
@project64 Would it be possible to revisit this particular issue? PJ64 using Jabo for default is causing a lot of issues. I see people on the internet asking whether any emulator can run Paper Mario. PJ64 can run Paper Mario. Jabo's plugin flickers with it. Same goes for heaps of other games. If you're okay with switching to Glide64 for default, the RDB can be updated and heaps of outdated status entries can be replaced.

On a related note, if PJ64 can get framebuffer notification support in the near future, there should no longer be performance issues with framebuffer support using Glide64 for Banjo Kazooie, assuming Glide64 works with PJ64's method. It's really bad that PJ64 can't even run a game like Banjo Kazooie "out of the box".

PR #542, opened 2015-07-16
Minor changes in English main file

MELERIX:
missing changes from:
https://github.com/project64/project64/commit/73b55a2f9f8f292140e9027903da9290d127d394

and a minor improvement.

### DIFF
--- a/Source/Glide64/Combine.cpp
+++ b/Source/Glide64/Combine.cpp
@@ -15901,7 +15901,11 @@ void CombineBlender ()
       }
       A_BLEND (GR_BLEND_SRC_ALPHA, GR_BLEND_ONE);
       break;
-
+      
+      case 0x5000: /* Vigilante 8 explosions */
+      A_BLEND (GR_BLEND_ONE_MINUS_SRC_ALPHA, GR_BLEND_SRC_ALPHA);
+      break;
+            
     default:
       A_BLEND (GR_BLEND_SRC_ALPHA, GR_BLEND_ONE_MINUS_SRC_ALPHA);
     }


cxd4:
"Field order:" or "Sort order:" makes more sense than "Field**s**" order", with the 's'.

I think the original "Sort fields in this order:" was more readable/correct, but it's not a big deal.

MELERIX:
uhm, then should be without "s" ?

cxd4:
That's what I meant, yeah.

Or "Sort fields in this order:" or blah.  The original looked fine IMO but if you're just trying to sync the English language files to changes to the Spanish ones then yeah I understand.

MELERIX:
well, the change is mostly to avoid troubles with the translations and gui size.

cxd4:
If "field**s**" should be plural with the "s" in it, then it would be easier to read "Order of fields".

"Apples order?"
"Tiles order?"
"Houses order?"

The "s" just doesn't seem to imply the same thing.  Order OF apples would make sense, but without the "of", in its most compact form it should say "Apple order".

MELERIX:
ok, changed ;)

cxd4:
Yep, looks good now.

cxd4:
Regarding the merge conflicts, they can be fixed by doing:
```batch
cd project64
git checkout master
git pull https://github.com/project64/project64.git
git checkout lang
git merge master
```
Then Git terminal will merge `master` with `lang` (your pull request's branch name) except it will highlight the conflicting lines with a series of `>>>>>>>>` characters.  You can then manipulate those to resolve the merge conflicts yourself as you see fit (required) then finish committing the merge.

Alternatively, re-submitting the pull request works.  `git rebase` possibly does too.

project64:
I should remove all references to beta code out of the source .. MSG_INVALID_EXE has no place in the .exe any more

PR #544, opened 2015-07-18
cleaner RdramChanged algorithm (not biased to only 4 vs. 8 MB)

cxd4:
It's just confusing to see arbitrary values like 0x400000 or 0x800000 scattered through if/else.

Should it be `if (old_size == 0x400000)`, `if (old_size != 0x800000)`, `if (old_size > 0x400000)` ... none of those are clear to what's meant for the function to do.

To be more intelligent the function should really only be caring to VirtualFree/VirtualAlloc based on this:
```c
if (old < new) {
    /* re-allocate RDRAM[old] extended up to RDRAM[new] */
} else if (old == new) {
    return;
} else {
    /* re-allocate RDRAM[old] freed down to RDRAM[new] */
}
```

### DIFF
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
@@ -5386,14 +5386,12 @@ void CMipsMemoryVM::RdramChanged ( CMipsMemoryVM * _this )
 			WriteTrace(TraceError,__FUNCTION__ ": failed to allocate extended memory");
 			g_Notify->FatalError(GS(MSG_MEM_ALLOC_ERROR));
 		}
-		_this->m_AllocatedRdramSize = 0x800000;
 	}
 	else
 	{
 		VirtualFree(_this->m_RDRAM + 0x400000, 0x400000,MEM_DECOMMIT);
-		_this->m_AllocatedRdramSize = 0x400000;
 	}
-
+	_this->m_AllocatedRdramSize = new_size;
 }
 
 void CMipsMemoryVM::ChangeSpStatus()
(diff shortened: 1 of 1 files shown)


Azimer:
The only thing that makes me curious about this is what if there are bad rdb values?  Should you allow the end user to hang themselves with the bad value or have it corrected?  Perhaps that is somewhere else in the code.

cxd4:
Well actually @Azimer, that was part of why I felt like making these changes.

In my 64-bit fork of Project64 1.4, I added options to use more than just 4 or 8 MB of RDRAM.  A couple games were accessing the limiters for up to 64 or even (in Aleck64 cases) 512 MB of RDRAM.  Since the 32-bit Windows code for MemoryFilter was not portable to 64-bit, my options were a) let the ROMs crash on reading past 8 MB of RAM b) implement address range checking in software for LW, LB, LH ... c) add arbitrary options to use more than 8 MB of RDRAM.  Out of curiosity and personal interest in what some games did, I felt like playing with c) for the sake of learning.

But yes, ideally you shouldn't be able to allocate more than 16 MB of RDRAM.

And according to AIO/RPGMaster, it already is possible to edit the CFG file (but not the RDB file--only the CFG file lets you do this) to use more than 8 MB of RDRAM.  He noticed a corruption of the allocated RDRAM while loading saved states or switching ROMs, and that was the other thing this pull request was meant to fix.

LegendOfDragoon:
@Azimer, currently the only valid sizes for RDRAM in the RDB are 4 MB and 8 MB. Here's where it's located if you want to look at & possibly change. https://github.com/project64/project64/blob/master/Source/Project64/Settings/SettingType/SettingsType-RDBRamSize.cpp#L37

cxd4:
Nah, not going to change.  I would agree mostly with Azimer that settings besides 4 or 8 MB are mostly "bad values" and don't care to implement them in the RDB file either.

What RPGMaster was talking about, was the CFG file in what I said about it letting you set values besides 4 or 8 MB.  So that has the bug already; the aim of this pull request was to dissolve its corruption.

LegendOfDragoon:
I was thinking of maybe adding a 16MB option for the RDB.

Issue #549, opened 2015-07-28
Make CMemoryLabel::AsciiToHex marginally more efficient


JMarlin:
Kind of a silly issue, but I saw a super minor thing I could improve upon and thought I'd jump on it.

Converting from a monolithic switch to a slightly smarter range check can reduce a worst case of 22 comparisons down to 5.

JMarlin:
Reopening due to incorrect case conversion math

cxd4:
whoops forgot about this though I think I remember your changes being merged, unless there was something else you left the issue open for?

JMarlin:
Nope, if everything looks good to you then this is clear to be closed.

cxd4:
Okay.  Anything else to apply this issue to in other sources @project64 or is this the end?

JMarlin:
I haven't looked at it in a while, but I believe this same method was duplicated elsewhere in the codebase and perhaps the whole thing could use a good search-and-replace, but I don't know when I'd have time to get around to it so at the moment, on my end, I'm going to be calling this the end.

JMarlin:
Shit. Did not mean to hit close

cxd4:
I think you're right but forgot exactly the other locations.

The RSP plugin and 3 or 4 other copies of the same function you improved throughout the Project64 core project.  Maybe you can find them by right-clicking the `AsciiToHex` function and doing "Find All References" in Visual Studio.

JMarlin:
Not using VS, but I'm sure I can grep it just the same

cxd4:
Ah, cool!

Yeah I've been trying to get PJ to compile without Visual Studio for some time now (or at least without Visual Studio Professional, to start out with).  Of course on a couple occasions I just commit changes on Linux anyway, but it's nice to actually be able to compile, run and test without having Windows installed.

tony971:
I've been able to compile with community for months.

cxd4:
That just required specifying a "Microsoft account".

Then after that they use those to track where your wife and kids live along with what sorts of other "free" products you're interested in so they can sell you teh r*mz.

PR #550, opened 2015-07-28
#549 changed switch/case table to cascaded range checks

JMarlin:


cxd4:
I think I remember attempting some commits to use a look-up table ...
```c
static const char hexfigs[] = {
    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
    'A', 'B', 'C', 'D', 'E', 'F',
};
```
But in this case it had to be a 128-byte lut; I didn't know how to install it to the C++ classes.

So then I tried to replace all the calls with the standard library `strtoul`, as that should be able to simplify this code too so that the custom function wasn't needed.  I can't remember what happened with that PR though...though I think for the RSP it ended up being merged.

JMarlin:
Thanks! It's a dumb little thing, but this is my first accepted contribution to an OSS project and I definitely fist-pumped the air the second I read this email.

cxd4:
Cool, then congratulations on your first OSS contribution.  It's a fair start.

Source/Project64/N64 System/Mips/Memory Labels Class.cpp
```diff
@@ -23,33 +23,25 @@ DWORD CMemoryLabel::AsciiToHex (char * HexValue)
...
-		case 'f': Value += 15; break;
-		default: 
-			Value = (Value >> 4);
-			Count = Finish;
+			Value += Current - '0';
+		}
+		else
+		{
+			if(Current < 'A')
+				Current += 'A' - 'a';
```
cxd4: There was one question I had been wondering here.

In ASCII, `'A' - 'a'` evaluates to a negative number, because `('A' < 'a')`.
So I think instead of converting uppercase to lowercase, this converts lowercase to non-printables.

But better algorithm in principle overall anyway so good job.  I'll let anyone else decide if it's an issue.

JMarlin:
Dammit, you're right. Reopened issue to fix that.

JMarlin:
Fixed by pull request #552

PR #553, opened 2015-08-01
more rdb updates

dsx-:
is good


cxd4:
The overall changes here seem all fine currently except for I think the comment on _Mario no Photopi_ saying "needs [third-party] RSP plugin", when the default one that comes with PJ64 doesn't really have any issues with it either.

Was that the only thing delaying being able to merge these changes Mr. zilmar?

dsx-:
default rsp cant do mario no photopi with hle, yours can

needs rsp plugin

cxd4:
Well, mine can't either unless you hex-edit the conf to force the semaphore lock on anyway even when there is HLE.  But generally the feature is only safely available through LLE, which is why my next release has a configuration system to make the feature unavailable in HLE.  You can force it on anyway but it's not a very safe idea for some games.

PR #555, opened 2015-08-01
Fix PI_DMA_READ

Nekokabu:
fix PI_DMA_READ like a PI_DMA_WRITE.

### DIFF
--- a/Source/Project64/N64 System/Mips/Dma.cpp
+++ b/Source/Project64/N64 System/Mips/Dma.cpp
@@ -35,12 +35,18 @@ void CDMA::OnFirstDMA()
 void CDMA::PI_DMA_READ()
 {
 //	PI_STATUS_REG |= PI_STATUS_DMA_BUSY;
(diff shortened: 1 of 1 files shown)


cxd4:
Good ideas, thanks for offering these changes.

I never thought to made PI_DMA_READ consistent with PI_DMA_WRITE because I cannot do hardware tests.  I simply did the fix to PI_DMA_WRITE based on what some games needed that 1964 had figured out; I don't have any way to prove that I should do the same changes to PI_DMA_READ and make that the same.

Nekokabu:
I believe to have forgotten to add 1. I have confirmed it in the memory view.
That part is the added part 64DD cnv support.

cxd4:
Ah, then you may be right.

If that change was intended as a bug fix, then I would have made it to its own commit.
It's always nice to be able to cite some hardware tests or a ROM that the fix proves.

Thanks again.

Issue #290, opened 2015-03-20
Fushigi no Dungeon - Fuurai no Shiren 2 - Oni Shuurai! Shiren Jou!


theboy181:
not supported in PJ64, but is loading in Mupen64.

unsupported OPcode?

theboy181:
Ganbare Goemon - Mononoke Sugoroku too

AmbientMalice:
Crash is caused by 32bit memory being enabled. Will test Ganbare Goemon - Mononoke Sugoroku.

https://github.com/project64/project64/pull/291

AmbientMalice:
Goemon - Mononoke Sugoroku crash was also caused by 32 bit mode.

https://github.com/project64/project64/pull/292

theboy181:
closing assuming that the RDB gets updated.

PR #562, opened 2015-08-17
Update Fuurai no Shiren 2 RDB

Nekokabu:


AmbientMalice:
You disabled 64 bit mode. Are you sure that was a good idea? It was having OpCode crashes with 32 bit mode back in March. https://github.com/project64/project64/issues/290

Nekokabu:
I deleted it, by mistake:)
thanks!

cxd4:
same thing with the jabo settings?
```
-Clear Frame=2
-Culling=1
-Self Texture=1
```
Were those also deleted on accident, or were they deleted because you found Jabo's Direct3D seems to look better on the game with those settings removed?

PR #565, opened 2015-08-19
reduced the compiler errors if building w/ VS2008 Express

cxd4:
Let me know if there is something I did incorrectly or overlooked with this change.

The idea is to start slight adjustments to be flexible to the possibility of building on some other platforms besides Microsoft Windows (or at least with the full desktop features), as well as potentially be able to compile Project64 with just the lightweight Express editions of Visual Studio installed.

### DIFF
--- a/Source/Project64/N64 System/Mips/Memory Class.h
+++ b/Source/Project64/N64 System/Mips/Memory Class.h
@@ -10,6 +10,15 @@
 ****************************************************************************/
 #pragma once
 
+/*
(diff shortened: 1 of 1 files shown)


cxd4:
Ya, not one of my favorite changes to make.

As it turns out, there is a long discussion by Microsoft about it in `<objbase.h>`:
```c
#define __STRUCT__ struct
#define interface __STRUCT__
```

PR #567, opened 2015-08-20
Make sure main.cpp does include <tchar.h>.

cxd4:
Fixes more VS2008 Express compiler errors, this time in `main.cpp`, because the main source file uses the `_T()` macro a few times.

The errors don't happen on non-Express VS because there you have access to WTL and ATL, so:
* `afxres.h` includes `User Interface.h`,
* which includes `WTL App.h`,
* which includes `<atlbase.h>`,
* which includes `<atlcomcli.h>`,
* which includes `<atlcore.h>`,
* which **finally** includes `<tchar.h>`.

But to port to other platforms, or compile in VS2008 Express, `main.cpp` needs to be independent of WTL or such things, when it's really just the functions and features from `<tchar.h>` that it needs.

### DIFF
--- a/Source/Project64/main.cpp
+++ b/Source/Project64/main.cpp
@@ -1,4 +1,5 @@
 #include "stdafx.h"
+#include <tchar.h>
 #include <Tlhelp32.h>
 
 CTraceFileLog * LogFile = NULL;


project64:
I do not want #include <tchar.h> in.

the reason is the code base should only be utf8 never going to be compiling with unicode ... all usages of _T() should be removed and adding tchar.h means they can creep in easily.

cxd4:
>all usages of _T() should be removed

That works, too.  I did not know you wanted them removed.

Currently you have that here:
https://github.com/project64/project64/blob/master/Source/Project64/main.cpp#L25
... and here ...
https://github.com/project64/project64/blob/master/Source/Project64/main.cpp#L100

So you mean you just want them deleted so tchar.h has no purpose?

project64:
yes better to remove the _T so
LogFilePath.SetNameExtension("Project64.log");

this code is never going to be compiled as unicode so no need to set up the text strings for it.

cxd4:
good idea, sounds easy enough

PR #570, opened 2015-08-22
Fix Blurry Fonts on High-DPI Windows 10

exhalatio:


exhalatio:
I'll add the high-DPI of resources and fixed the font size on the screens.

exhalatio:
@JunielKatarn That sound like a good idea.

JunielKatarn:
Yeah, Visual Studio always does this dumb "add conditionals for everything" by default.
It's good for actual coding, but tends to need a hand when it comes to defining project configurations.

cxd4:
Kind of looks somewhat complex though (well coming from me that means little).

It isn't as if I'm qualified to comment on correctness with this type of thing, but I was hoping there would be a solution that didn't involve tripling the BMP files with extra BMP files.  This also adds to the amount of Windows-specific code PJ64 has so therefore adds to the ratio of code which isn't directly portable, weighted against the code that makes instant porting more effortless.

Maybe there's a way to improve the scaling algorithm GDI uses or whatever to be flexible to different versions of Windows without having to add more lines of BMP statements?  BTW I have no idea what I'm talking about.

exhalatio:
Yeah, it seems to be scaled one BMP file by using the StretchBlt instead of BitBit. I tried scaling StretchBlt. It looks ugly. I seem to need to be more research.
Portable Device also seems to be switched with a multiple of BMP files in DPI.

Left: StretchBlt (Incomplete...) Right: HighDPI BMP
scaling

cxd4:
Could always create the UI graphical elements with GL functions instead of GDI ones.

There would still be stuff like window/mouse/keyboard interaction that would still have to be Windows-centric.

DaMan69:
I wouldn't lose sleep over the prettiness of a barely seen screen.

toehead2001:
Regarding the About dialog box. I think it would be simpler to just use standard window, rather than have all those bitmaps at different sizes. Something like this:
about project64 1

exhalatio:
Standard dialog is great.
I have a problem fonts is blurred on Main UI. so, I hope added <B>PerMonitorHighDPIAware</B> in Source / Project64 / Project64.vcxproj

Last commit Windows 10
win10

Fixed
win10fixdpi

toehead2001:
I'd say close this pull, and open a new one with just the `<EnableDpiAwareness>PerMonitorHighDPIAware</EnableDpiAwareness>` in the .vcxproj file

The first-run Language Selector will still need to be updated like the About box was, but that can be done in a separate pull request.

exhalatio:
I'll make new pull request the DPI.
Would you updated the first-run Language Selector?

PR #573, opened 2015-08-26
Propose a breakpoint in TranslateVaddr #ifdef _WIN64.

cxd4:
Since this only seems to really affect the ability to have a recompiler, I don't really care that much about this now.  I was under the initial concern that it affected the completeness of being able to run the interpreter in 64-bit (seems to not ever get reached).

However it was a part of my original pull request:  https://github.com/project64/project64/pull/530
So for that reason I have it split to this one.

### DIFF
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
@@ -528,6 +528,15 @@ bool CMipsMemoryVM::TranslateVaddr ( DWORD VAddr, DWORD &PAddr) const
 	{
 		return false;
 	}
+
+	if (sizeof(void *) > sizeof(DWORD))
+	{
+/*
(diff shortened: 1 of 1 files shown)


Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
```diff
@@ -528,6 +528,15 @@ bool CMipsMemoryVM::TranslateVaddr ( DWORD VAddr, DWORD &PAddr) const
 	{
 		return false;
 	}
+
+	if (sizeof(void *) > sizeof(DWORD))
```
project64: why not just an #ifdef

cxd4:
Either one works.

Your macro preference could in some scenarios be insufficient to prove whether `sizeof(void *)` transcends the size of the data type which `PAddr` was declared under.  I'm not sure one can use the pre-processor alone to reliably predict that every time, as hints like `_WIN64` are just text.

PR #586, opened 2015-09-03
Fix Filename in compressed savestate

Nekokabu:


project64:
I am probably just not thinking clearly so not seeing it, what does this fix

PR #590, opened 2015-09-07
PIF RAM mutual DMA x64 breakpoints replaced with fast C

cxd4:
Not my favorite commit to make, since pointer typecasts of this type break strict aliasing rules and may make the functionality of Project64 vulnerable to bugs in theoretical, future ports to less popular systems.  (I could have avoided these casts but figured the asm was written to force win32 performance so didn't feel like bothering with the minor chance of what would happen if the better C code didn't optimize.)

Sorry about the C extern function, but until instructed otherwise by zilmar in this pull request, that is the only way I know of to get this job done.  I don't know how to install it/define it inside one of the C++ classes or whether that's even what people want, though I will follow any necessary instructions.

Anyway, results--using the _Basic CFB Plugin_ from 2000:
cd64mem_x64
neon64_x64
sp_crap_x64
bomberman_x64

### DIFF
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
@@ -40,7 +40,7 @@ CMipsMemoryVM::CMipsMemoryVM( CMipsMemory_CallBack * CallBack, bool SavesReadOnl
 	m_IMEM       = NULL;
 }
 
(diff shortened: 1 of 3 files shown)


cxd4:
Most games work regardless of whether MemoryFilter is implemented for Win64, by the way.  It isn't MemoryFilter that's concerning half the commercial stuff; it's COP1 in the interpreter using asm.

Issue #103, opened 2015-02-19
broken RSP MWC2 decoding of signed offsets


cxd4:
An instruction such as:
```asm
SSV     $v21[6], -24($23)
```
... is incorrectly getting disassembled with a hexadecimal offset:
![hex_is_bad_mk](https://cloud.githubusercontent.com/assets/1396303/6269453/408c43d0-b821-11e4-8d7c-de81af722a0f.png)
(Saying "T7" instead of "$23" is fine though if you declare an assembly language macro to name it T7.)

FFFFFFE8_16 is not a valid RSP offset, either as a 16-bit for primary ops or a 7-bit offset for SWC2.

Valid range of offsets is (-64 <= offset < +64).

If we had wanted to write this as hex you would say (-0x40 <= offset < +0x40), but either one is still much easier to type in something like Windows Calculator than "FFFFFFE8" if you're just trying to add the addresses.  (Typing "- 24" decimal isn't really any less convenient either, and decimal is the standard base for arithmetic analysis.)

PR #592, opened 2015-09-07
[RSP] bad DMEM offsets when (offset < 0)

cxd4:
Here is an old issue I forgot to close which we can fix right away:
https://github.com/project64/project64/issues/103

In this case I have changed the LWC2 and SWC2 instruction formatting string from:
```c
"%s\t$v%d[%d], 0x%04X(%s)"
```
to
```c
"%s\t$v%d[%d], %c0x%03X(%s)",
```

I changed from %04X to %03X because only DMEM offsets 0x000:0xFFF are reachable.

However strictly speaking the signed 7-bit offset can only range between -64:+63 decimal, so it could really be %02X and still be consistent.  Which is why I'm not even convinced that we really need signed hexadecimals for especially this when it could be signed decimals.  But it is a bug that the signed RSP offset value is being printed as stuff like 0xFFFFFFFF instead of just -0x001, as that is the point being addressed here.

### DIFF
--- a/Source/RSP/RSP Command.c
+++ b/Source/RSP/RSP Command.c
@@ -26,6 +26,8 @@
 
 #include <windows.h>
 #include <stdio.h>
+#include <stdlib.h>
+
 #include "opcode.h"
 #include "RSP.h"
 #include "CPU.h"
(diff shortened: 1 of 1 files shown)


cxd4:
I was starting to wonder if maybe it should be `%s ? "-" : ""`, not `%c ? '-' : '+'`.
Since + is only mathematically correct for positive numbers, not 0.

Tho it's legal C technically as a unary op so whatever.  I guess if you're fine with it then I am.

Issue #32, opened 2015-01-26
Mario Kart 64 MP speed issues


theboy181:
Not sure if this can be emulated correctly, but MK64 has issues with emulating the correct speed while playing Mario Kart multiplayer. Most noticeable in DK jungle with 3 or more players. I think that Nintendo has coded the game to increase some multiplier to achieve reasonable results in frame rate when playing MP on the real HW.  This is only emulated correctly on the Wii VC.

cxd4:
I got that working fairly well by tweaking the VI refresh rate setting in the RDB, but I have no idea what I set it to anymore.  The last time I played with it to get it working, was when some guy said he offered $500 to whoever got it working, so I felt like getting it working to see what his reaction was, then when I finished answering all his PMs about how to install the fix he just ran off and never came back. :D

But yeah try lowering it from 1500 to something like, 1100 maybe 600, it'll make a lot of other things slower as well but DK Jungle 3-player won't be on crack either.

theboy181:
This creates only a work around and has other effects during other pets of the gameplay. 

I think the real fix is going to be finding out what nintendo has actually done and emulate those features. I have tested this on many emulators and no change. The real HW seems to be ok though. 

Tbh I Havnt checked CEN64 since the core updates.   I wonder.

Predator82Germany:
My N64 had this problem too but don´t know which pal version of the game it was

cxd4:
Typically non-cycle-accurate has to rely on said workarounds, but whatever, technically it is an issue.

It definitely has been asked about a lot, for all the emulators, no matter what timing they do.

Predator82Germany:
That means it´s fixable anyhow in face of the big-n issue?

cxd4:
wut

cxd4:
> I think the real fix will be finding out what Nintendo has actually done with this game,

No I think the real fix will be what I said.
Set VI refresh rate manually when not automatic, because non-cycle-accurate emulators have no "real fix", and "finding out what Nintendo has actually done with this game" is more of a stereotype on your end for something that's rather self-explanatory:  Nintendo had timer issues on the actual system so needed to adjust to it in this particular course.  What they did is clear; doesn't mean they'd know how to fix this issue from an emulator's perspective.

theboy181:
:(  

Sounds like there is very little hope then. I don't see how fixing the timing is going to help here. The game acts similar to OOT regarding Links movement speed frame rate. 

Can you explain any further?

[...]

PR #605, opened 2015-09-16
Fix multiplayer timings in Mario Kart 64 (cheat)

tony971:
As explained [here](http://tasvideos.org/GameResources/N64/MarioKart64/Patches.html), these codes fix the speedups that occur in multiplayer Mario Kart 64. This cheat should be enabled by default but I don't know how to do that.


tony971:
Fixes #32

PR #607, opened 2015-09-17
Move Mario Kart multiplayer timing fix to rdb

tony971:


AmbientMalice:
Just my two cents, but is this 100% desirable with no possible unwanted side effects? Because sticking hacks in the RDB means the average user will have no idea they're enabled.

tony971:
Yes, it is 100% desirable with no side effects. The problem came about in emulators because testers for the game noticed that on certain maps in multiplayer, the game ran in slow motion. Turns out, MK64 makes its internal calculations at 60 fps but runs at 30 fps of in-game time. The timings were based off of the idea that you will have one frame of input lag. With original hardware, adding more players would introduce more frames of input lag on certain maps. To counteract this slow motion, the developers added in a scaling factor to increase how much in-game time to use per input lag frame. This would make the game run at full speed again, despite the additional lag frames. Modern computers won't have additional lag frames so these scaling factors only make the game run at super speed. The cheat code basically forces the scaling factor to be 2, like it would be for single player.

tony971:
Yes, it is 100% desirable with no side effects. The problem came about in emulators because testers for the game noticed that on certain maps in multiplayer, the game ran in slow motion. Turns out, MK64 makes its internal calculations at 60 fps but runs at 30 fps of in-game time. The timings were based off of the idea that you will have one frame of input lag. With original hardware, adding more players would introduce more frames of input lag on certain maps. To counteract this slow motion, the developers added in a scaling factor to increase how much in-game time to use per input lag frame. This would make the game run at full speed again, despite the additional lag frames. Modern computers won't have additional lag frames so these scaling factors only make the game run at super speed. The cheat code basically forces the scaling factor to be 2, like it would be for single player.

tony971:
To;dr: I guess there would be regressions if your computer was enough of a toaster to need more than one frame to process input from multiple players. That doesn't apply to computers capable of emulation, though.

cxd4:
dunno, but nice avy

loganmc10:
Isn't this cheat only for the U version of the ROM? Doesn't the cheat need to be modified for the E and J versions?

loganmc10:
By the way, some mupen64plus users are reporting that sometimes (randomly) in multiplayer mode, an item will become unusable (you'll pick it up but can never use it). It only happens in multiplayer mode and the belief is that it is happening as a result of this cheat.

Has anyone using Project64 come across this issue?

tony971:
You can find all the info about the cheat [here](http://tasvideos.org/GameResources/N64/MarioKart64/Patches.html). Basically, the game devs noticed that as they added more players, there were more input lag frames. In order to combat this, they added speed multipliers to these maps when they were expected to have increased delays. Computers don't suffer from the increased input lag frames, so the speed multipliers just made the maps run quicker. This cheat nulls the speed multipliers.

tony971:
So the game added their own version of frame skipping. The cheat removes it.

loganmc10:
I understand what the cheat does, and it works for the U version of the ROM. But it doesn't work for the E version of the ROM. Cheats are region specific, they need to be offset depending on the region of the ROM. The cheat on that website was made for the U rom.

Look at the cheats in Project64.cht, you'll notice that the E version and U version of games have different codes for the same cheats.

tony971:
Ahh I didn't realize that the offsets were different across regions. This probably should only be applied to the U versions, then.

loganmc10:
Yeah there is even a tool/guide to converting the cheats:

http://cheats.emulation64.com/converting.htm

I tried converting this cheat based on the offsets of others cheats for MK64 but it didn't work, someone that knows more about how this all works might know what needs to be modified.

tony971:
Well in the meantime, I have a feeling disabling the cheat for our European/Japanese friends might get rid of the unusable item reports. I haven't had the issue in 4P mode, but my cheat is applied to the correct portion of the game.

Howl1n:
Thanks for this great cheat, really would like to use it on retroarch (rpi/recalbox). However I dont know which config file to change, any ideas?

PR #611, opened 2015-09-18
Have error-checking when drawing RSP command window.

cxd4:
According to MSDN, `DrawText` needs a signed `int` for the string length parameter.

That's why the 64-bit compilation of Project64 RSP plugin issues several warnings--about "loss of precision" when implicitly converting from `size_t` to a 32-bit integer and also about signed/unsigned mismatch if we were to compile this with warning level /W4.

So instead of passing a `size_t` where we need an `int`, I find it convenient that the standard library `sprintf` function returns `int` which fits in perfectly to the parameters being read here.
http://www.cplusplus.com/reference/cstdio/sprintf/

Related very closely to this pull request as well, for more reasons on why I did that:
https://github.com/project64/project64/pull/610

### DIFF
--- a/Source/RSP/RSP Command.c
+++ b/Source/RSP/RSP Command.c
@@ -247,6 +247,7 @@ void DumpRSPData (void)
 void DrawRSPCommand ( LPARAM lParam )
 {
 	char Command[150], Offset[30], Instruction[30], Arguments[40];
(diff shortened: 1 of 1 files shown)


cxd4:
Screenshot just to show current status after doing this change:

new_rsp_window

Issue #172, opened 2015-03-01
problems in translations files with UTF-8 encoding


MELERIX:
translations files with UTF-8 encoding are not displayed properly.

Lithium64:
I forgot to say to Zilmar, but they need to be encoded as ANSI to function properly.

MELERIX:
I see, thanks.

I'll fix my translation later with ANSI encoding and few other changes.

project64:
which particular language are you testing with and is looking wrong, is it in every string or only in certain parts of the UI ?

MELERIX:
I tested with Spanish and Portuguese, but of course there are more language files affected.

an easy way to test, for example: open PJ64 -> open File Menu -> click in Language

and you will see this: http://i.imgur.com/d2s0Aea.jpg

project64:
OK done two commits (https://github.com/project64/project64/commit/b6359831676ca51a80dc3510f75271553baedea9 and https://github.com/project64/project64/commit/71cc6def7b35f0d533d2b8a108834823834ac288), How is it looking for you now? Is there still issues that need fixing?

MELERIX:
oh, I'll test in few minutes.

btw, can you commit this please ? https://github.com/project64/project64/pull/207 patch is updated now.

MELERIX:
now is working fine ;)

translations in UTF-8 are displayed properly now.

thank you Zilmar.

MELERIX:
well, I reopened this again, because now I've discovere some new issues.

for example in a clean start of Project64, when you select roms folder and it start to scan the roms, the status bar will show some weird characters:

![Text](http://i.imgur.com/qqCWWnL.jpg)

project64:
ok I will have a look tonight, there could be other issues as well .. I am loading the strings as utf8, but all the windows functions for 8 bit treat it as ansi which does not work, so now I convert them all to utf16 once loaded and calling the W functions of windows .. so this is a case where I am most likely passing an 8 byte char array instead of 16 byte char array.

Lithium64:
@project64 
![bug4](https://cloud.githubusercontent.com/assets/10705488/6501103/ecdf26da-c2f2-11e4-939e-bc8d84fa718b.png)

MELERIX:
few more issues detected...

![Text](http://i.imgur.com/ANWoNt4.png)

![Text](http://i.imgur.com/uuybuGc.png)

![Text](http://i.imgur.com/fFHBuTz.png)

[...]

PR #615, opened 2015-09-20
Updates encoding of Lang/French.pj.Lang

Watilin:
Addresses #172 

I double-checked all accented chars in the file, but couldn't compile to see if anything's been broken -- can't install Visual Studio at the moment. It *normally* shoudn't break, but if someone is willing to test the french version.

**Note:** in Github website, the file will appear as containing diamonds with question marks. This is normal: the site interprets contents as UTF-8 while the file is actually ANSI.

Also, I'm willing to fix spelling mistakes and improve typography, but I'm not sure I could put this in the same PR, as this seems to me like something completely different from an encoding issue.

I'll wait for your comments ;)


project64:
the file can not be ansi it has to be utf-8 file

Watilin:
Ok, this also answers the question I was asking to myself about why the project wasn't using UTF-8 for all langs :P
Actually the issue I was adressing relates to Jabo's plugin, so this PR is pointless. I'm closing it.

Watilin:
Just one dumb question, and I'm sorry if it's already been asked: all translation files begin with a byte order mark. Is this on purpose? I've been told that BOM is only necessary when carrying a stream over network.

project64:
yes the BOM is necessary with out it the file is treated and ansi file instead of utf8 file. Project64 will also now not load any language file that does not have the BOM

cxd4:
It also is necessary for The Duck Song.  https://www.youtube.com/watch?v=MtN1YnoL46Q

Watilin:
Nooo it's that song again, I thought I could forget it  >_<

PR #620, opened 2015-09-23
Pif Ram: Get rid of pointer casts

lioncash:
Same codegen and doesn't have undefined behavior as far as the standard cares.

### DIFF
--- a/Source/Project64/N64 System/Mips/Pif Ram.cpp
+++ b/Source/Project64/N64 System/Mips/Pif Ram.cpp
@@ -173,13 +173,13 @@ void CPifRam::PifRamWrite()
 				{
 					ResponseValue = (ResponseValue << 8) | ((Response[(z - 1)*2] << 4) + Response[(z - 1)*2+1]);
 				}
-				*(QWORD *)&m_PifRam[48] = ResponseValue;
+				std::memcpy(&m_PifRam[48], &ResponseValue, sizeof(QWORD));
 				ResponseValue = 0;
 				for (int z = 7; z > 0; z--)
 				{
 					ResponseValue = (ResponseValue << 8) | ((Response[((z + 8) - 1)*2] << 4) + Response[((z + 8) - 1)*2+1]);
 				}
-				*(QWORD *)&m_PifRam[56] = ResponseValue;
+				std::memcpy(&m_PifRam[56], &ResponseValue, sizeof(QWORD));
 			}
 			break;
 		case 0x08: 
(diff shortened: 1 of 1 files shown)


cxd4:
```asm
;.intel_syntax noprefix
;# ...
	mov     eax, DWORD PTR [rax]
	bswap   eax
	mov     DWORD PTR [rdx], eax
;# ...
```
So does `memcpy` from string.h.
```c
#include <stdint.h>
#include <string.h>

extern unsigned char PifRamPos[64];
extern unsigned char DRAM[0x00FFFFFFul + 1];

static uint32_t swap32by8(uint32_t word);

int main(void)
{
    uint32_t PIF_RAM_word;

    memcpy(&PIF_RAM_word, &PifRamPos[0], sizeof(uint32_t));
    PIF_RAM_word = swap32by8(PIF_RAM_word);
    memcpy(&DRAM[0x000000], &PIF_RAM_word, sizeof(uint32_t));

    return 0;
}
```

However the undefined behavior was left more as a reminder to remove the external linkage on `swap32by8` (memcpy() as well for MSVC without /Oi when compiling) which mostly is just a slow macro-instruction, which I think I will do next after this is merged.

lioncash:
The undefined behavior I was referring to was casting to a stricter aligned type and dereferencing it.

cxd4:
Right.  I was not happy with the type cast either.

Really there were only 2 reasons I did not write memcpy from the get-go:
* To make it easier for zilmar to know it was equal C to the inline asm he used.
* Because even `memcpy` would not have had an use if I had avoided externally linking my `swap32by8` from the virtual memory unit in the first place.

So I like that you did this, but in the future when the Intel-centric endianness stuff is taken care of, my future idea was to eventually apply point 2 above with either
```c
#define WORD_BYTE_ENDIAN_ADJUSTMENT     3
#define BES(i)          ((i) ^ WORD_BYTE_ENDIAN_ADJUSTMENT)

for (i = 0; i < 64; i += 4) {
    RDRAM[i + 0] = PifRamPos[i + BES(0)];
    RDRAM[i + 1] = PifRamPos[i + BES(1)];
    RDRAM[i + 2] = PifRamPos[i + BES(2)];
    RDRAM[i + 3] = PifRamPos[i + BES(3)];
}
```
... or, for cacheability and size preferences ...
```c
for (i = 0; i < 64; i++)
    RDRAM[BES(i)] = PifRamPos[i];
```

But yes, ultimately if we can get everything in the accurate MIPS-native endian, then really just one single `memcpy` would replace the whole loop.

PR #623, opened 2015-09-23
Add support for Visual Studio 2015, xdebug appears to no longer be av…

death-droid:
…ailable in Visual Studio 2015. So for now memtest must be disabled for Debug builds to operate correctly.


JunielKatarn:
Hi. I had played around with this in the past.
Is xdebug even used? As far as I can tell, you can safely build and run without including that header (VS2013, VS2015).

Issue #618, opened 2015-09-22
Improving French Translation


Watilin:
I want to improve the french translation, especially the overall consistency, and I'd like to debate about which accurate terms to choose and to stick to. Here is a non exhaustive list of such terms.

(By the way, let me know if there is a better place to talk about this kind of stuff.)

In **bold**, what seems to be the current consensus. Discuss if you disagree ;)

English Term | Possible Translations | Example .Lang Entries
-------------|-----------------------|----------------------
Save State | état, **sauvegarde d'état**, sauvegarde rapide | 125, 127, 129
Slot | slot (keep english term), **emplacement** | 191, 220
Input | input (keep english term), entrée, contrôle, **contrôleur** | 424, 2007
Directory | dossier, **répertoire** | 402, 485
Core | **noyau**, cœur (depends on context) | 501, 502, 542
Cheat | **cheat** (keep english term), triche, code triche | 11, 130
Reset | **réinitialiser** (config menus), relancer, **redémarrer** (actual emulation) | 121, 340, 720, 1100
Recursion | (see below) | 485

Also, there are some technical or vague terms I'm almost sure the current translation can be improved, but since I have no idea what these terms mean, I can't find a better translation.

**Current consensus** here seems to be don't translate.

Technical Term | Example Entry
---------------|--------------
Advanced Block Linking | 505
PI DMA | 511
TLB unmapping | 514
Fast SP | 530
Display lists | 701
LUT | 2026
Byte-swapping image | 2036

Here are the terms I'd really like someone to explain them to me.

Term | Entry
-----------|------
Quantity digit | 1016
Counter factor | 525
Failed to load word | 2009

About 485, “Use Directory Recursion”: I cannot find any satisfying translation to this, so I propose to alter the meaning and use “Utiliser également les sous-dossiers” (or “sous-répertoires”) which translates back to “Also use sub folders”. This is way more understandable in my opinion.

Now comes the spaaace problem. As you may know, the french language takes more space than the english one. Words are often longer, and sentences often use more words. I already noticed buttons and dialogs where the end of the text is clipped. Other translations may suffer this problem as well. Is there any planned work on fixing this? Just asking.

[...]

PR #624, opened 2015-09-23
Improves French Translation

Watilin:
There goes the PR I promised in #618.

One thing related to typography that I didn't add is nobreak spaces, because they're dificult to spot in source code, and because their handling by common windowing libs is not good enough to be worth it.

Of course this is a draft and I'm counting on you guys to help me achieving a really high-quality translation :)


Lang/French.pj.Lang
```diff
@@ -316,59 +316,59 @@
 
 //Save Type
 #620#  "Par défaut"
-#621#  "4-kbit EEPROM"
-#622#  "16-kbit EEPROM"
-#623#  "32-kbyte SRAM"
+#621#  "4 kbit EEPROM"
+#622#  "16 kbit EEPROM"
+#623#  "32 ko SRAM"
```
cxd4: This is the only change I didn't know about, because in English I know the hyphen (`'-'`) is required.

I don't know if that's ever appropriate for French, though, but Google Translate suggested "32-kbyte SRAM" to "32 Ko de SRAM", so I guess you're probably right.

cxd4:
Looks great to me otherwise, as far as I can tell from capitalization/terminology and what I know at least about English anyway.

Lang/French.pj.Lang
baptiste0602: He's right : "de" doesn't mean anything in this situation because it's a technical specification.
But I think, we should capitalize the "k" in "Ko".

Lang/French.pj.Lang
```diff
@@ -90,94 +90,94 @@
...
+#318#  "Forcer le feedback"
+#319#  "Format de fichier"
 
 //Select ROM
-#320#  "Sélectionner le Répertoire ROM en cours"
+#320#  "Sélectionner le répertoire de ROM actuel"
 
 //Messages
-#340#  "Mauvaise ROM ? Utilisez GoodN64 & vérifier la mise à jour de RDB"
+#340#  "Mauvaise ROM ? Utilisez GoodN64 et vérifiez la mise à jour de RDB"
```
baptiste0602: I think "Mauvaise ROM ? Utilisez GoodN64 et vérifiez que le fichier RDB est à jour" is better here.

Watilin:
Aw, you shouldn't have merged so early, there was still work to do.
Now I think the simpliest is that I commit on top of my `lang-fr` branch and then make another PR.

I'd pick baptiste's suggestions, except about capitalizing the "K" in "ko". As confusing as it looks, this is a small "k" in "ko" but a capital "M" in "Mo". Sauce: https://fr.wikipedia.org/wiki/Octet

cxd4:
>Sauce: https://fr.wikipedia.org/wiki/Octet

Looks like a mistake/typo to me.  I have no clue why they would capitalize the metric prefixes "G" for giga-, "M" for mega-, "T" for tera but only a lowercase "k" for kilo-??

Also, why does it have to be Mo instead of MB or Mb?  That page says the o means "octet" instead of byte.  What if we ported Project64 into compiling for a system with 9-bit bytes? :P Then it would no longer be an octet.

baptiste0602:
Yes, I think the lowercase "k" is a typo...

With a "french" pronunciation, "bit" and "byte" is the same. Maybe what's why we translate it to "octet". I think we can keep Mb and MB.

cxd4:
I was thinking it's because "byte" is a made-up word by some American who wanted to refer to memory-addressable units without confusing "bite", as in to eat.  "Octet" is more correct but only refers to 8-bit bytes, hence my wondering that.

I do know that at least in English, Mb = mega **bit**, and MB = mega **byte**.  Hate it when I intentionally lowercase the 'b' and people think I'm sloppily referring to bytes.

Watilin:
I know for sure that this is a lower case "k" because it's the *kilo* prefix from the international unit system, just like *km* for kilometer and *kg* for kilogram. More on this can be found here: https://fr.wikipedia.org/wiki/Kilo

> What if we ported Project64 into compiling for a system with 9-bit bytes? :P Then it would no longer be an octet.

It would be a [nonet](https://en.wikipedia.org/wiki/Nonet) :P

cxd4:
Well it would be a nonet on that system, but an octet when compiling for x86.

So we would need a French translation of a term agnostic to the number of bits in a byte.

cxd4:
Good point about kilometer being "km" and mega-/giga-/etc. meter being "Mm" and "Gm".  That's true; the "k" in "km" must be lowercase.  I forgot about that example, so maybe you're right.  I've never seen anyone say "kB" in English, though.  It should be KB and KiB.

Watilin:
> So we would need a French translation of a term agnostic to the number of bits in a byte.

Sometimes during my studies, « mot » (word) was used to indicate data units of different sizes, including the 8-bit byte. But I don't think this is common usage.

I trust french users would still understand « octet » even for non 8-bit units, although I must admit that it would introduce ambiguity between 8-bit and, to stick with the example, 9-bit bytes.

Issue #556, opened 2015-08-01
Windows 10 BSOD.


AmbientMalice:
So I'm getting win32kfull.sys BSOD crashes when I try to boot games with Project 64. I'm running Windows 10 with an Nvidia GTX 750ti and an Intel Core Quad CPU. The OS is fully updated. (Not that I have a choice.)

Breadwinka:
Run in compatibility mode for windows 8 or 7 usually fixes this issue, You may still get some BSOD from time to time.

LegendOfDragoon:
@AmbientMalice can you try testing older versions of Project64? Such as 1.7 and 1.6?

AmbientMalice:
@LegendOfDragoon I'll test later. I highly doubt they'll be an issue since PJ64 causing BSODs is a 2.x trait.

LegendOfDragoon:
Even if 1.6 is safe, it may have started somewhere in 1.7.

Breadwinka:
@LegendOfDragoon doesn't happen in 1.6 or 1.6.1

Breadwinka:
@LegendOfDragoon Just tried 1.7 and no BSOD. But if I use 2.x i get bsod almost everytime I chose a game.

LegendOfDragoon:
@Breadwinka thanks for the input. Have you tried the absolute earliest version of 2.0? Also, how about a late version of 1.7, like 1.7.51.51?

Breadwinka:
@LegendOfDragoon ill try and find a 1.7.51.51 version and a earlier 2.0 as well. Sucks pj64 site is down atm.

Breadwinka:
@LegendOfDragoon 2.1 = BSOD

project64:
@Breadwinka  is it all plugins, or just jabo's gfx plugin. Since there is no drivers pj64 can not cause a blue screen directly. The only thing it can do is making different calls to the gfx driver which can cause a blue screen. Try using opengl, or jabo's 1.6 plugin with 2.x. Most likely it needs a driver update.

Breadwinka:
![image](https://cloud.githubusercontent.com/assets/13487931/9032408/4cf44f04-398c-11e5-814f-fb83856acc8f.png)

those were the plugins i was using when the BSOD but also happens with stock plugins too. ill try with 1.6 plugins shortly.

Breadwinka:
[...]

PR #625, opened 2015-09-23
Potential Windows 10 BSOD FIX

LuigiBlood:
That's not even a joke.
I tried my hardest to force a BSOD either via CLI or the ROM Browser on my Windows 10 laptop (which had that problem).

It just works now. Moving the code there somehow fixes it.

It usually takes less than 10 tentatives to BSOD my laptop, and I've tried a 100 times here.

If the placement is too problematic about "g_Settings->SaveBool(GameRunning_LoadingInProgress, false);", my other suggestion is just to put to what it was and then remove RefreshMenu from there, OR we put both lines of code inside StartEmulation2().

Either way you can still get this pull request in and find a way later.


Breadwinka:
+1 to this as this has fixed https://github.com/project64/project64/issues/556 for me as well.

LegendOfDragoon:
@LuigiBlood good lookin out :smile: .

PR #627, opened 2015-09-24
Make solution files work with all versions of Visual Studio (2010+)

JunielKatarn:
Follow up on merge request https://github.com/project64/project64/pull/623 .
Simplified VCXPROJ files in order to make them 2013 and 2015 compatible.


JunielKatarn:
Hi @death-droid. Would you take a look?
I centralized the PlatformToolset setting into the file Platform.props file. That way, the compiler collection can be set in one place.

At this point, Project 64 can be edited and built with both VS2013 and VS2015.

JunielKatarn:
Also fixes WinXP's "Not a Win32 application" error with recent Visual Studio versions (2012+).

death-droid:
Seems good to me @project64 this would be helpful to merge in.

JunielKatarn:
Good news.
Latest update makes editing and building with any version of Visual Studio starting with 2010 work out of the box (no tweaking required).

Note: VS2008 SLN and project files need to be kept apart because they have a different format.

JunielKatarn:
Great news!
The latest commit in this PR fixes an issue on Windows XP, which blocked PJ64 from loading the ROM list (and therefore, being usable at all).

In short, Project64 should now work on XP when build from Visual Studio 2010 or later.

PropertySheets/Platform.props
```diff
@@ -1,34 +1,39 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup Label="Configuration">
+    <PlatformToolset>v140_xp</PlatformToolset>
```
cxd4: (in response to issue https://github.com/project64/project64/issues/649)

So far I think I might need to change this line to v120_xp, just to avoid consuming the necessary time involved with installing VS2015 when I already have 2013 on my hard disk (and probably uninstalling 2013 to save disk space).

PR #632, opened 2015-09-25
`if` is not a function.

cxd4:
It's more of an operator.

`function(expression)` is correct.  `operator(expression)` is misleading and confusing.

It should be `if (expression)`, not `if(expression)`.

### DIFF
--- a/Source/RSP/Recompiler Ops.c
+++ b/Source/RSP/Recompiler Ops.c
@@ -449,7 +449,7 @@ void Compile_ADDI ( void ) {
 	if (RSPOpC.rt == 0) return;
 
 	if (RSPOpC.rt == RSPOpC.rs) {
-		if(Immediate != 0) {
+		if (Immediate != 0) {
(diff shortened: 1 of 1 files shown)


Watilin:
There could be a project's style guide. Not sure if a `STYLEGUIDE.md` file has a special meaning for GitHub but I've seen [such a file in another project](https://github.com/OpenUserJs/OpenUserJS.org/blob/master/STYLEGUIDE.md).
Could also be on gh-pages or in the wiki.

cxd4:
Yeah, gh-pages or GitHub wiki would be cool.

If this were my repository, I would just ask everyone to contribute in 4-space tabs, not tabbed indents, but I would have to expect this stylistic rule to be violated all the time anyway as both Visual Studio and Notepad++ stubbornly indent with tabs by default every time, without some rigorous configuration to turn that feature off.

But if(exp) vs. if (exp) I thought should be mostly self-explanatory.

Watilin:
Well... https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328

PR #636, opened 2015-09-26
Banjo-Kazooie (U) v1.1 works in x64 now.

cxd4:
For some reason, the USA 1.0 ROM doesn't need this instruction implemented, but 1.1 does.

Otherwise BK 1.1 fails with the interpreter breakpoint upon booting before displaying any gfx.
Thanks to @tony971 for making me aware of this.


cxd4:
x64_bk

tony971:
Also fixes Top Gear Rally

PR #637, opened 2015-09-26
Fix flags for debug/validate logging.

JunielKatarn:
Trace logging was not working due to wrong flag setting. This small change fixes it.

Question: is there a way to set the VALIDATE_DEBUG flag at runtime (console argument, configuration setting)?

### DIFF
--- a/Source/Project64/main.cpp
+++ b/Source/Project64/main.cpp
@@ -26,7 +26,7 @@ void InitializeLog ( void)
 
 	LogFile = new CTraceFileLog(LogFilePath, g_Settings->LoadDword(Debugger_AppLogFlush) != 0, Log_New,500);
 #ifdef VALIDATE_DEBUG
-	LogFile->SetTraceLevel((TraceLevel)(g_Settings->LoadDword(Debugger_AppLogLevel) | TraceValidate));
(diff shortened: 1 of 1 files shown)


cxd4:
If I understood your question correctly (if not you already knew this), it's:

Configuration Properties :: C/C++ :: Preprocessor :: Preprocessor Definitions

The console argument to the MSVC compiler executable should be
`/D "<C_MACRO_TO_#define_GOES_HERE>"`

Otherwise the only way I could think of would be to #define it inside of `afxres.h`.

JunielKatarn:
I'm afraid not. I meant setting it when the program is already running.
But thanks for the reply.

On Sat, Sep 26, 2015 at 4:57 PM, cxd4 <notifications@github.com> wrote:

> If I understood your question correctly (if not you already knew this),
> it's:
>
> Configuration Properties :: C/C++ :: Preprocessor :: Preprocessor
> Definitions
>
> The console argument to the MSVC compiler executable should be
> /D "<C_MACRO_TO_#define_GOES_HERE>"
>
> Otherwise the only way I could think of would be to #define it inside of
> afxres.h.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/project64/project64/pull/637#issuecomment-143506096>.
>



-- 
- Julio

cxd4:
Right, I simply deleted my post because I realized we were thinking of totally different things.

Issue #638, opened 2015-09-27
(x64) Games try to use more than 8mb of memory


tony971:
Affects the following games:
- Army Men - Sarge's Heroes
- Dazaemon 3D
- F-ZERO X
- Glover 2 (Beta)
- Hybrid Heaven
- Mario no Photopi
- NBA Live 2000
- Roadsters Trophy
- South Park
- Supercross 2000
- Vigilante 8
- War Gods
- Xena Warrior Princess - Talisman of Fate

tony971:
Also affects Dazaemon 3D

cxd4:
I had tested _Army Men:  Sarge's Heroes_ a while back.

The problem is that setting this game to use 16 MB or more of DRAM (even so high as 512 MB of RAM) will leak Project64 into either an infinite loop, TLB status crashes, address error exceptions or other problems, depending on how much RAM you ask for exactly in the CFG file.

I'm thinking it might be because the game only checks a small switch between 4 or 8 MB RDRAM.  If you set it to something higher than 8 it will just have logic like if (size == 8 MB) limit = 0x7FFFFF; else limit = LIMIT_4MB_DRAM;.  At the same time, some leak in the game code is accessing beyond 8 MB of DRAM (either because of Project64 bug or just because the game is badly designed).

Or maybe the RSP plugin and RDP plugin are not both in complete agreement on what the memory limit is.  This is why the plugin specs should have passed MAX_DRAM_ADDR to the plugins, not just a pointer to DRAM.

Otherwise the long and short of it is to check the addressing range in the R4300 ops, or to simply fall back on MemoryFilter working in 64-bit Windows builds of Project64.  Though I think the latter is easier.

theboy181:
Have you tired setting higher memory settings for these games in the per game CFG file?

project64:
closing old issue, if this is still an issue it can be reopened

PR #639, opened 2015-09-27
MemoryFilter starts to compile.

cxd4:
Since tony reported issue https://github.com/project64/project64/issues/638, I decided it can be partially understandable to use MemoryFilter on 64-bit Windows.  So I guess I don't care not to implement it.

It basically compiles just fine now, but you need to have the LB, LW, SW etc. functions not be "DWORD"-centric for the physical address and use more portable MIPS code instead of x86/WINAPI stuff.  That remaining business can be tended to in a later PR.

### DIFF
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
@@ -1987,7 +1987,21 @@ void CMipsMemoryVM::ResetMemoryStack()
 
 int CMipsMemoryVM::MemoryFilter( DWORD dwExptCode, void * lpExceptionPointer ) 
 {
-#ifdef _M_IX86
+#if defined(_M_IX86) && defined(_WIN32)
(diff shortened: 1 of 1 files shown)


tony971:
Oh I was just about comment on the build failing after this commit but @JunielKatarn beat me to it

cxd4:
The issue is that the Windows LLP64 ABI does

#if (32-bit Window, not 64-bit)
typedef unsigned int size_t;
typedef unsigned long DWORD;

So the types really are equivalent, but C grammatical differences between `int` and `long` cause VC++ (not MSVC) to think there is no relationship to compare them satisfactorily.  A minor detail I'll have to rearrange in the next PR.

Yes Ctrl+Shift+B is easy, but I often have to do a clean 100% rebuild because of testing across VC2008 and VC2015, which fight with each other's build results, and if I re-clone the repository that will also force me to do a clean rebuild.  So I felt I lacked the time to properly test this on 32-bit as well.

PR #640, opened 2015-09-27
Use 8mb of memory when needed

tony971:
Fixes the following games in x64:

* Banjo Kazooie (pending #636 )
* Banjo Tooie
* Bassmasters 2000
* Forsaken 64
* The Legend of Zelda - Ocarina of Time
* Namco Museum 64
* NASCAR 2000
* NASCAR 99
* NHL 99
* Pokemon Stadium 2
* Power Rangers - Lightspeed Rescue
* SD Hiryuu no Ken Densetsu
* Sydney 2000 (Unreleased)
* Tom Clancy's Rainbow Six
* Yoshi's Story

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -532,6 +532,7 @@ Status=Compatible
 Plugin Note=[video] (see GameFAQ)
 Culling=1
 Self Texture=1
+RDRAM Size=8
 
 [514B6900-B4B19881-C:4A]
 Good Name=Banjo to Kazooie no Daibouken 2 (J)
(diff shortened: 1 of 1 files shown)


AmbientMalice:
Just a thought -- in the long run, doing the 16MB RAM global with 4MB set for games that need it might be way more simple than setting 8MB on a per-game basis for every game that randomly craps itself in x64.

tony971:
Well there are at least 13 more games that have memory issues (see #638 ) but these ones can be fixed by setting 8mb in the rdb. The rest should* be fixed once MemoryFilter is ported to x64.

tony971:
If they didn't get it, there would be a break point in Memory Virtual Mem.cpp

tony971:
If they didn't get it, there would be a break point in Memory Virtual Mem.cpp

cxd4:
@ExtremeDude2 it's a breakpoint in 64-bit builds of Project64, because 32-bit uses MemoryFilter.

ExtremeDude2:
onice

PR #650, opened 2015-10-01
Remove external linkage of swap32by8()--now static.

cxd4:
Well actually this PR does at least two things.
* keeps `swap32by8()` (which I introduced a while back) to a single translation unit with `static` linkage, instead of making it `extern` within the headers to access it across more than one module
* adds some constants and functions/macro functions to help port platform endianness

I'd say the second is more important.

I don't have a clue what platforms like Android might do for byte order, but on x86 we have all kinds of code saying `x ^ 3` all the time.  I think this is not good in the long term for releasing Project64 with a big-endian memory spec either on platforms that have it for their native endian, or just simply to test plugins of the zilmar spec with `PluginInfo -> MemorySwapped = FALSE;`

Sorry about macro functions, don't know how those work with MSVC's debugger for people who use that more often.  In theory we could convert the macro functions to C external functions without macros.

### DIFF
--- a/Source/Project64/N64 System/Mips/Pif Ram.cpp
+++ b/Source/Project64/N64 System/Mips/Pif Ram.cpp
@@ -297,19 +297,14 @@ void CPifRam::SI_DMA_READ()
 			{
 				continue;
 			}
-			RDRAM[RdramPos ^3] = m_PifRam[count];
+			RDRAM[RdramPos ^ 3] = m_PifRam[count];
 		}
 	}
 	else
 	{
-		for (size_t i = 0; i < 64; i += 4)
+		for (size_t i = 0; i < 64; i++)
 		{
-			unsigned __int32 pif_ram_dword;
-			std::memcpy(&pif_ram_dword, &PifRamPos[i], sizeof(unsigned __int32));
-
-			pif_ram_dword = swap32by8(pif_ram_dword);
-
-			std::memcpy(&RDRAM[SI_DRAM_ADDR_REG + i], &pif_ram_dword, sizeof(unsigned __int32));
+			RDRAM[(SI_DRAM_ADDR_REG + i) ^ 3] = PifRamPos[i];
 		}
 	}
 	
@@ -393,14 +388,9 @@ void CPifRam::SI_DMA_WRITE()
 	}
 	else
 	{
-		for (size_t i = 0; i < 64; i += 4)
+		for (size_t i = 0; i < 64; i++)
 		{
-			unsigned __int32 rdram_dword;
-			std::memcpy(&rdram_dword, &RDRAM[SI_DRAM_ADDR_REG + i], sizeof(unsigned __int32));
-
-			rdram_dword = swap32by8(rdram_dword);
-
-			std::memcpy(&PifRamPos[i], &rdram_dword, sizeof(unsigned __int32));
+			PifRamPos[i] = RDRAM[(SI_DRAM_ADDR_REG + i) ^ 3];
 		}
 	}
 	
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.cpp
@@ -40,9 +40,9 @@ CMipsMemoryVM::CMipsMemoryVM( CMipsMemory_CallBack * CallBack, bool SavesReadOnl
 	m_IMEM       = NULL;
 }
 
-unsigned long swap32by8(unsigned long word)
+static uint32_t swap32by8(uint32_t word)
 {
-    const unsigned long swapped =
+    const uint32_t swapped =
 #if defined(_MSC_VER)
         _byteswap_ulong(word)
 #elif defined(__GNUC__)
@@ -282,7 +282,7 @@ bool CMipsMemoryVM::LB_VAddr(DWORD VAddr, BYTE& Value)
 		return false;
 	}
 
-	Value = *(BYTE*)(m_TLB_ReadMap[VAddr >> 12] + (VAddr ^ 3));
+	Value = *(uint8_t *)(m_TLB_ReadMap[VAddr >> 12] + (VAddr ^ 3));
 	return true;
 }
 
@@ -293,7 +293,7 @@ bool CMipsMemoryVM::LH_VAddr(DWORD VAddr, WORD& Value)
 		return false;
 	}
 
-	Value = *(WORD*)(m_TLB_ReadMap[VAddr >> 12] + (VAddr ^ 2));
+	Value = *(uint16_t *)(m_TLB_ReadMap[VAddr >> 12] + (VAddr ^ 2));
 	return true;
 }
 
@@ -344,7 +344,7 @@ bool CMipsMemoryVM::LB_PAddr(DWORD PAddr, BYTE& Value)
 {
 	if (PAddr < RdramSize())
 	{
-		Value = *(BYTE*)(m_RDRAM + (PAddr ^ 3));
+		Value = *(uint8_t *)(m_RDRAM + (PAddr ^ 3));
 		return true;
 	}
 
@@ -361,7 +361,7 @@ bool CMipsMemoryVM::LH_PAddr(DWORD PAddr, WORD& Value)
 {
 	if (PAddr < RdramSize())
 	{
-		Value = *(WORD*)(m_RDRAM + (PAddr ^ 2));
+		Value = *(uint16_t *)(m_RDRAM + (PAddr ^ 2));
 		return true;
 	}
 
@@ -416,7 +416,7 @@ bool CMipsMemoryVM::SB_VAddr(DWORD VAddr, BYTE Value)
 		return false;
 	}
 
-	*(BYTE*)(m_TLB_WriteMap[VAddr >> 12] + (VAddr ^ 3)) = Value;
+	*(uint8_t *)(m_TLB_WriteMap[VAddr >> 12] + (VAddr ^ 3)) = Value;
 	return true;
 }
 
@@ -427,7 +427,7 @@ bool CMipsMemoryVM::SH_VAddr(DWORD VAddr, WORD Value)
 		return false;
 	}
 
-	*(WORD*)(m_TLB_WriteMap[VAddr >> 12] + (VAddr ^ 2)) = Value;
+	*(uint16_t *)(m_TLB_WriteMap[VAddr >> 12] + (VAddr ^ 2)) = Value;
 	return true;
 }
 
@@ -469,7 +469,7 @@ bool CMipsMemoryVM::SB_PAddr(DWORD PAddr, BYTE Value)
 {
 	if (PAddr < RdramSize())
 	{
-		*(BYTE*)(m_RDRAM + (PAddr ^ 3)) = Value;
+		*(uint8_t *)(m_RDRAM + (PAddr ^ 3)) = Value;
 		return true;
 	}
 
@@ -486,7 +486,7 @@ bool CMipsMemoryVM::SH_PAddr(DWORD PAddr, WORD Value)
 {
 	if (PAddr < RdramSize())
 	{
-		*(WORD*)(m_RDRAM + (PAddr ^ 2)) = Value;
+		*(uint16_t *)(m_RDRAM + (PAddr ^ 2)) = Value;
 		return true;
 	}
 
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.h
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.h
@@ -31,24 +31,15 @@
 #endif
 
 /*
- * This will help us gradually be able to port Project64 for big-endian CPUs
- * later.  Currently it is written to assume 32-bit little-endian, like so:
+ * To do:  Have address translation functions here?
+ * `return` either the translated address or the mask to XOR by?
+ *
+ * This will help us gradually be able to port Project64 for big-endian CPUs.
+ * Currently it is written to assume 32-bit little-endian, like so:
  *
  * 0xAABBCCDD EEFFGGHH --> 0xDDCCBBAA HHGGFFEE
  *   GPR bits[63..0]         b1b2b3b4 b5b6b7b8
  */
-#define BYTE_SWAP_MASK          3u
-#define HALF_SWAP_MASK          ((BYTE_SWAP_MASK) & ~1u)
-#define WORD_SWAP_MASK          ((BYTE_SWAP_MASK) & ~3u)
-
-// Convert MIPS server-native byte order to client-native endian.
-#define BES(offset)             ((offset) ^ (BYTE_SWAP_MASK))
-
-// Convert MIPS server-native halfword order to client-native int16_t order.
-#define HES(offset)             ((offset) ^ (HALF_SWAP_MASK))
-
-// Convert MIPS server-native word order to client-native int32_t order.
-#define WES(offset)             ((offset) ^ (WORD_SWAP_MASK))
 
 class CMipsMemoryVM :
 	public CMipsMemory,
--- a/Source/Project64/N64 System/Mips/Memory Virtual Mem.h
+++ b/Source/Project64/N64 System/Mips/Memory Virtual Mem.h
@@ -30,7 +30,25 @@
 #define Eip     Rip
 #endif
 
-extern unsigned long swap32by8(unsigned long word);
+/*
+ * This will help us gradually be able to port Project64 for big-endian CPUs
+ * later.  Currently it is written to assume 32-bit little-endian, like so:
+ *
+ * 0xAABBCCDD EEFFGGHH --> 0xDDCCBBAA HHGGFFEE
+ *   GPR bits[63..0]         b1b2b3b4 b5b6b7b8
+ */
+#define BYTE_SWAP_MASK          3u
+#define HALF_SWAP_MASK          ((BYTE_SWAP_MASK) & ~1u)
+#define WORD_SWAP_MASK          ((BYTE_SWAP_MASK) & ~3u)
+
+// Convert MIPS server-native byte order to client-native endian.
+#define BES(offset)             ((offset) ^ (BYTE_SWAP_MASK))
+
+// Convert MIPS server-native halfword order to client-native int16_t order.
+#define HES(offset)             ((offset) ^ (HALF_SWAP_MASK))
+
+// Convert MIPS server-native word order to client-native int32_t order.
+#define WES(offset)             ((offset) ^ (WORD_SWAP_MASK))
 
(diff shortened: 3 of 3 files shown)


cxd4:
And yes I compiled and ran both 32- and 64-bit builds of the EXE this time.  Loads Mario64.

Source/Project64/N64 System/Mips/Pif Ram.cpp
```diff
@@ -297,19 +297,16 @@ void CPifRam::SI_DMA_READ()
...
 	}
 	else
 	{
-		for (size_t i = 0; i < 64; i += 4)
-		{
-			unsigned __int32 pif_ram_dword;
-			std::memcpy(&pif_ram_dword, &PifRamPos[i], sizeof(unsigned __int32));
-
-			pif_ram_dword = swap32by8(pif_ram_dword);
+		register size_t i;
```
lioncash: There's no particular reason to leave this in a scope it's not used in. By the way, have you benchmarked or checked that the assembly produced is actually better with the register keyword. I can bet it does not alter semantics at all (it's also deprecated).

cxd4: >There's no particular reason to leave this in an outer scope it's not used in.

That assumes the view that "scope" should exist in between.

If you suppose that in between the body of a loop (such as `for`) and just outside the loop, should be a scope of its own (the control and heading of the loop), then it is your opinion that the declaration should go to the loop header.

In particular C99 and C++ should allow just as you say, but I find that there's also no particular reason for me to assume that this slight increase in readability to humans varies directly with ease of parse-ability (and readability) to compilers.  (The same logical issues with direct language parsing having constituted some of the options held back before C99.)  I actually find that the C89 version looks less open to interpretation and less ambiguous, IMO, and easier for me to read in my honest opinion, if not also easier for the compiler to parse faster.

But if it looks just that much harder for you or @project64 to read, I will change it back.

> By the way, have you benchmarked or checked that the assembly produced is actually better with the register keyword.  I can bet it does not alter semantics at all

Then you propose that my purpose for specifying `register` is the pretense that it necessarily makes the generated code any different or run faster?

Like any hint for any language, I'm aware that it makes no automatic difference in the end result.  It is just a compiler hint--one which often goes ignored (especially in this case, as the compiler's job is to optimize which should be registers, which should be on the stack, and which should be in-lined or even exist at all).  (And it **usually** has all the information it needs to correctly decide such, for most basic algorithms.)

But unless you can find a reason to suggest that there's anything to lose from my approach, and not just that "there's almost never anything to gain", then I'm really not sure what positive difference it makes to change that.

cxd4: >There's no particular reason to leave this in an outer scope it's not used in.

That assumes the view that "scope" should exist in between.

If you suppose that in between the body of a loop (such as `for`) and just outside the loop, should be a scope of its own (the control and heading of the loop), then it is your opinion that the declaration should go to the loop header.

In particular C99 and C++ should allow just as you say, but I actually find that the C89 version looks less open to personal interpretation and less ambiguous, IMO, and easier for me to read in my honest opinion, if not also easier for the compiler to parse faster.  (Said issues with direct language parse-ability to compilers having constituted many of the other hold-backs of optional features not present before C99.)

But if it looks just that much harder for you or @project64 to read, I will change it back.

>By the way, have you benchmarked or checked that the assembly produced is actually better with the register keyword.  I can bet it does not alter semantics at all

Then you propose that my purpose for specifying `register` is that I think otherwise?

Like any hint for any language, I'm aware that it makes no automatic difference in the end result.  It is just a compiler hint--one which often goes ignored (especially in this case, as the compiler's job is to optimize which should be registers, which should be on the stack, and which should be in-lined or even exist at all).  (And it **usually** has all the information it needs to correctly decide such.)

But unless you can find a reason to suggest that there's anything to lose from my approach, and not just that "there's _almost_ never anything to gain", then I'm really not sure what positive difference it makes to change that.

Another minor use for `register` is that it prevents taking the address of something by mistake.

cxd4: I also suspect that, for this function, the type of `i` has no reason to be different between each of these `for` loops in the cleaned-up end result, so while your reasoning is that `i` should not be declared outside of the relevant local scope, I am sure that the declaration can be re-used for multiple `for` loops for easier maintainability over changing what the loop variable type is in a single declaration, not multiple repeated super-local declarations.

lioncash: > but I find that there's also no particular reason for me to assume that this slight increase in readability to humans varies directly with ease of parse-ability (and readability) to compilers.

The reason I suggested it is not only for readability (if that was the case, I wouldn't have suggested it at all). Why expose a variable to unintentional modification in a scope it is not used in (this applies to any other case, not just this one in particular; yes I know nothing modifies it past this, but that's not the point I'm getting at)? There is no point (IMO) with regards to software robustness/safety aside from the usually heard, "This is how it is in C89 or <insert dated standard from other language here>, therefore I do this habit".

> The same logical issues with direct language parsing having constituted some of the options held back before C99

Right, however, I'm quite sure parsing methods have gotten better since prior to 1999.

> if not also easier for the compiler to parse faster.

This is something I'll need a source on that supports the claim (preferably a modern one).

> unless you can find a reason to suggest that there's anything to lose from my approach

Aside from it being deprecated?

cxd4: Yes my method does theoretically expose the variable to unintentional modification.

However for some functions it also can make maintaining the loop variable type easier.  If it was `int i` or `size_t i` each time, you could have to weigh your security argument against one of maintainability of centralizing the type declaration instead of repeating it within each/multiple `for` loops.  Repeating the same exact text/keywords over and over also can make debugging more tedious or risk drawbacks (probably not security ones though, of the form you're citing).

>Right, however I'm quite sure parsing methods have gotten better since prior to 1999.

OK, but I doubt that parsing methods "getting better" implies which style parses faster.

>This is something I'll need a source on that supports the claim (preferably a modern one).

It's not something I suspect directly off the words of others.

I'm simply observing the logic that normal local declarations are more consistent.
```c
// Would you prefer:
int i, j;
int x = 0, y = 0, z = 0;
for (i = 0; i < LIMIT; i++)
    for (j = 0; j < LIMIT; j++)
        { /* statements depending on x, y and z */ }

// or should it be as you say, to maximize the locality of relevant scope with:
```c
for (int i = 0; i < LIMIT; i++)
    for (int j = 0, x = 0, y = 0, z = 0; i < LIMIT; i++)
```
Who will say?  Maybe the latter can compile faster, but I have no intention of writing it.

>Aside from it being deprecated?

Although, this reasoning is rather recursive.

We can't have something "deprecated" without having a reason for doing it.  You can't cite deprecation as the reason for avoiding something because that's recursive reasoning; there has to be a reason *why* it was deprecated.

So what exactly does "deprecated" mean in this case?  Human nature permits anything to be deprecated no matter how useless or even how use**ful** it may have been.

lioncash: > Yes my method does theoretically expose the variable to unintentional modification.
> However for some………

[Unfortunately the ISO C++ Standard Committee does not agree](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#-p3-express-intent), [nor does CERT (even for C)](https://www.securecoding.cert.org/confluence/display/c/DCL19-C.+Minimize+the+scope+of+variables+and+functions), [nor does Google](https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Local_Variables), etc, etc. I could go on and on listing examples. If such a practice was more beneficial in the long run, I'm sure your preference would be the one accepted by the standardization committee.

>  Maybe the latter can compile faster

No matter which way you format your code, the difference is negligible. If it wasn't, you wouldn't have a multitude of coding styles for the language (and it would be pretty embarrassing for compiler writers, to be honest).

> Although, this reasoning is rather recursive

You did not ask why, you asked what. If you want the why on that, see [here](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4193.html#809) and [here](https://isocpp.org/files/papers/n4340.html) (the introduction provides your answer in this one, with more detail provided further in).

cxd4: You may also have noticed a common fundamental thing between those sources.
They all think more like human beings, less like computers.

Sometimes what they see as more readable, is to small parsers also.  Other times, not really.

>You did not ask why, you asked what.

Wrong, I said:  _unless you can find a reason to suggest that there's anything to lose_

To which you replied by re-stating that C++ deprecated it. That's recursive.

If you had asked me for a reason why there's anything to lose from doing C99's `for (int i = ...` as you stated, and I replied "Aside from it being deprecated?" (assuming some crazy, hypothetical C89-compatible standard chose to deprecate it, whether you personally liked it or not), would you not find that reasoning to be recursive?

*What* is not important.  The question always (and clearly the one I asked) is why.

Should a specification make any wrong or inaccurate turns, then more important is the ability to just not use that spec.  If it maintains backwards compatibility, conforming to it is easier.  Otherwise, how would you react.

>[Unfortunately the ISO C++ Standard Committee does not agree](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#-p3-express-intent)

You mean [this](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-recycle)?

Then yes, they do disagree with what I just said.

Unfortunately for them, their "reason" is poorly developed and has little explanation.
I would think that proper "recycling" would look more like this:
```c
int i;

for (i =   0; i <  200; i++) { x[i] = f(i); }
for (i = 400; i <  600; i++) { y[i] = g(i); }
for (i = 800; i < 1000; i++) { z[i] = h(i); }
```
It is fact that debugging this code is easier by simply changing the type of `i`, in this form.

>[nor does CRT (even for C)](https://www.securecoding.cert.org/confluence/display/c/DCL19-C.+Minimize+the+scope+of+variables+and+functions)

I think that declaring a variable at all, anywhere in your code, exposes it to hazards.

Sometimes maintenance hazards, sometimes "security" hazards of form you cite.
Just depends which type of hazard you prefer minimizing the risk of.

Off topic:  Here is another CERT page which I found was somewhat questionable:
https://www.securecoding.cert.org/confluence/display/c/FIO19-C.+Do+not+use+fseek%28%29+and+ftell%28%29+to+compute+the+size+of+a+regular+file

They dictate that the POSIX `fstat()` should be used in place of `fseek()` because of the fact that fseek() to the end of a file is undefined for **non-POSIX** implementations.  c wut they did thar?

>nor does Google outside of object construction where it makes sense, it's also an advisory rule in the MISRA-C guidelines to narrow scope where possible, etc, etc. I could go on and on listing examples.

I am sure you could and that most of them are perfectly credible sources.

But I have not and never will see C programming (cannot say for C++, though) as a question of what other human beings think, but as a question of how I am sure computer logic thinks.  I don't believe for a second that `for (size_t i = ...` is any less ambiguous or open to multiple possibilities of interpretation (or even personal language implementation) than just declaring the variable first.

>If such a practice was more beneficial in the long run, I'm sure your preference would be the one accepted by the standardization committee.

As far as standards committees are concerned, "beneficial" needs extra pragmatism.

They are more interested in the future because they have the power to control it.  You might find this theory is paranoid, but it's really just human nature.  To them, portability means very little if it's only because of backwards compatibility against something they would rather have set their own way.

>If you want the why on that, see [here](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4193.html#809) (section 809) and [here](https://isocpp.org/files/papers/n4340.html)

OK, so based on those links, their reasoning is that the rare plausibility of having a unique use for C's version of `register` could, in theory, be replaced by a more concrete and much more useful tool for design/optimization in a future version of the C++ specification.

Certainly possible, but I am glad this "deprecation" hasn't removed the old future yet until they can indeed confirm something to replace it.

Besides, I can take a page from your book and propose that not using the `register` keyword opens up part of your code to accidentally "taking the address" of a variable.  In ANSI C, taking the address of a variable whose declaration contains `register` class is invalid, as you cannot fetch a memory pointer for a CPU register.  (Similarly, you probably use `const` to prevent accidentally writing to a variable you find no plausibility in writing to.)

lioncash: >You may also have noticed a common fundamental thing between those sources.
They all think more like human beings, less like computers.

Yeah, they're what maintain and write the code; that's kind of how the relationship works.

> It is fact that debugging this code is easier by simply changing the type of i, in this form.

You can obsolete the need to change it by using [auto](http://en.cppreference.com/w/cpp/language/auto) in many circumstances where this would actually matter in practice, among other things (if modification of loop variables is frequent, outside of circumstances where the design as a whole is changed, either local or in another translation unit that necessitates the change, there's a clear design issue that isn't being thought out). You don't need to concern yourself with the type.

> If you had asked me for a reason why there's anything to lose from doing C99's for (int i = ... as you stated, and I replied by re-iterating that it is deprecated (assuming some crazy, hypothetical C89-compatible standard chose to deprecate it, whether you personally liked it or not), would you not find that reasoning to be recursive?

If you replied it was deprecated, and it was indeed the case (I would look for sources first to back it up or ask you to provide it), I would move code over to the non-deprecated form once confirmed; basic software maintenance 101.

> I don't believe for a second that for (size_t i = ... is any less ambiguous or open to multiple possibilities of interpretation (or even personal language implementation) than just declaring the variable first.

But it is and can be. Scoping rules exist. If they didn't, I would agree.

>They are more interested in the future because they have the power to control it. You might find this theory is paranoid, but it's really just human nature. To them, portability means little if it's only because of backwards compatibility against something they would rather have their own way.

literally_any_newer_language_spec.txt. 

> Besides, I can take a page from your book and propose that not using the register keyword opens up part of your code to accidentally "taking the address" of a variable... In ANSI C

[Sure, but this isn't a purely C codebase, so it's not enforced in this context](http://ideone.com/Wqs3lN).

cxd4: >Sure, but this isn't enforced in C++ (in C, go nuts if you like).

OK.  Well the reason why that's unfortunate, is because Project64 is written in C++. :)

But that means I should take your advice to heart, so I'll remove the "register" in a bit.

>If you replied it was deprecated, and it was indeed the case (I would look for sources first to back it up or ask you to provide it), I would move code over to the non-deprecated form once confirmed; 

Would you, now?  I almost find that impressive.

So if I'm reading what you said right (and assuming you read my post right):  if this crazy hypothetical future spec would deprecate the `for (type i =` style, you would adjust by moving over to the non-deprecated C89-ish counter style simply because it isn't deprecated.

That means that in spite of all these scope security and readability defenses for the C99 style you have cited, ultimately the deciding factor for you is what does the modernistic, newer spec have deprecated/not deprecated.

Well, no insult, I can genuinely applaud you for being so tolerant of basing correct "software maintenance" as such a simple matter of what other people (namely these official "committees") think.  It isn't that I can't respect this, but I personally see no point to being controlled singularly by the opinions of others--except of course when you are working for a company or other's project that persists the use of such a spec, then it's a no-brainer.  But for your own code and your own projects?  You could just as easily set the saner compiler to use an older version of the spec.

Like I said:  with "deprecation" always comes a reason, and that reasoning is all that counts.
Especially if they deprecate something that contradicts what they added to C99.

>basic software maintenance 101.

Good theory but I have an even easier one:  Respect backwards compatibility.

Then there is no cause for having to mass-change/maintain updates like that.

>literally_any_newer_language_spec.txt.

If they disregard backwards compat., then yes.  (See also:  OpenGL deprecation)

>But it is and can be. Scoping rules exist.

And always have existed--just as they have in C89.  However, whether the loop heading (as opposed to the body) constitutes its own existing "scope" is again a matter of one's own (and the individual language's) personal interpretation.  Maybe that could be another reason why ANSI C does not support what you say.

>You don't need to concern yourself with the type.

If the type is irrelevant, then I would think that to be one more reason why `for (i = 0; i < N; i++)` is easier to read than `for (arbitrary_type_here i = 0; i < N; i++)`.

Never did recall seeing "int" or "size_t" in any Calculus textbooks about sequences.
Hmm...could that be because the type is irrelevant?

>You can obsolete the need to change it by using auto in many circumstances where this would actually matter in practice, [...]

Possibly, but in my example `i` is also the parameter to 3 different functions.  For whatever reason (optimization, problems with porting, or even fixing design problems like you said--maybe my example is too "unique") it just is faster to debug/analyze to switch back-and-forth between by changing one central declaration, not 3+ declarations.  Otherwise, you could use a custom typedef (like the semi-custom ones in C99's stdint.h).

>Yeah, they're what maintain and write the code, it's a pretty nice one-sided relationship :^).

Kudos to them, then.  I am glad it's not one-sided for me, at least.

I have had few, temporary portability problems when compiling C89 code on modern C/C++ compilers.  (I do not always care to do things like write "register" though.)

cxd4:
I am sure that @project64 does not find it amusing to see me sparking (supposedly, to him at least) unnecessary friction with other contributors, so as thanks for this fun debate I will get to applying the critique of @Lioncash in a bit.  (Besides, I have a feeling zilmar prefers the C++ method.)

cxd4:
Done.

The `for` loop counter declarations were moved back to the modern C++ style they were before.

Source/Project64/N64 System/Mips/Pif Ram.cpp
lioncash: In closing, I'd like to address some things.

> Would you, now? I almost find that impressive.

You presented me with a, 'between a rock and a hard place' scenario. While I would not agree with a standard decision on that from a design standpoint., it's more a matter of, "In [x] years, will my program still be considered valid code on newer standards?". Compatibility works both ways. Moving to a new language is a consideration, but at that point that begs the questions of whether or not that's possible (assuming constraints, etc), or worth the time.

> And always have existed--just as they have in C89. However, whether the loop heading (as opposed to the body) constitutes its own existing "scope" is again a matter of one's own (and the individual language's) personal interpretation. Maybe that could be another reason why ANSI C does not support what you say.

You can likely get a definitive answer to this by e-mailing someone related to the ISO C Working Group (there are contact links on the working group's site).

> Good theory but I have an even easier one: Respect backwards compatibility.

The same can be said of forward compatibility as well.

That's all I wanted to address in particular.

project64:
@cxd4 the real issue with this pull request I have is I am trying to remove macros were ever possible, sometimes they can not be helped and they help a lot, but where possible I want to remove them and stop them being added in.

cxd4:
It's an interesting dilemma to avoid macros for this case, but I'm sure I'll find a way to improvise shortly.

cxd4:
On second thought @project64 I do believe I'll leave the byte order changes to you in the future.

In the meantime I've removed them from distracting this pull request.  Anything else left?

PR #654, opened 2015-10-02
Disable Data Execution Prevention for Project64.exe (Visual Studio 2010+)

JunielKatarn:
See forum post http://forum.pj64-emu.com/showpost.php?p=63666&postcount=19
(Pointed out by @Frank-74)

Builds based on VCXPROJ (VS2010 and later) don't currently disable DEP, making the program unable to unload old plugins (1.6 DLLs).

This simple setting change fixes the problem.
Built and tested on VS2015.

### DIFF
--- a/Source/Project64/Project64.vcxproj
+++ b/Source/Project64/Project64.vcxproj
@@ -36,6 +36,7 @@
       <SubSystem>Windows</SubSystem>
       <MinimumRequiredVersion>5.01</MinimumRequiredVersion>
       <StackReserveSize>1</StackReserveSize>
+      <DataExecutionPrevention>false</DataExecutionPrevention>
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>


AmbientMalice:
Is allowing PJ64 to load old plugins a good idea? IIRC, older plugins tended to use the registry as their dumping grounds, clashing with modern PJ64's more sensible, portable approach.

JunielKatarn:
I'm not sure, but this behavior would be consistent with Zilmar's builds.
I would vote for disabling, but for the time being, I just mean to keep it
consistent.

On Fri, Oct 2, 2015 at 12:44 AM, AmbientMalice <notifications@github.com>
wrote:

> Is allowing PJ64 to load old plugins a good idea? IIRC, older plugins
> tended to use the registry as their dumping grounds, clashing with modern
> PJ64's more sensible, portable approach.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/project64/project64/pull/654#issuecomment-144946281>.
>



-- 
- Julio

AmbientMalice:
Fair enough. (Accidentally deleted old post.) So long as it doesn't cause any side effects, it's probably okay.

PR #655, opened 2015-10-02
Disable global soft-depth and enable per-game.

AmbientMalice:
This will cause a significant reduction in CPU usage for a variety of games. Will need further testing to identify and games that slipped through.

### DIFF
--- a/Config/Glide64.rdb
+++ b/Config/Glide64.rdb
@@ -182,6 +182,7 @@ Internal Name=1080 SNOWBOARDING
 alt_tex_size=1
 depthmode=0
 fb_clear=1
+fb_render=1
 optimize_texrect=1
 swapmode=2
 
@@ -190,6 +191,7 @@ Good Name=1080 Snowboarding (JU) (M2)
 Internal Name=1080 SNOWBOARDING
 depthmode=0
 fb_clear=1
+fb_render=1
 optimize_texrect=1
 swapmode=2
 
@@ -834,6 +836,7 @@ Internal Name=DONKEY KONG 64
 depth_bias=64
 depthmode=1
 fb_clear=1
+fb_render=1
 lodmode=1
 
 [053C89A7-A5064302-C:4A]
@@ -842,6 +845,7 @@ Internal Name=DONKEY KONG 64
 depth_bias=64
 depthmode=1
 fb_clear=1
+fb_render=1
 lodmode=1
 
 [EC58EABF-AD7C7169-C:45]
@@ -850,6 +854,7 @@ Internal Name=DONKEY KONG 64
 depth_bias=64
 depthmode=1
 fb_clear=1
+fb_render=1
 lodmode=1
 
 [0DD4ABAB-B5A2A91E-C:45]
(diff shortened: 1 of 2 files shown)


cxd4:
Only one concern.  Wouldn't it make sense to make default ROM settings for testing unknown or new ROMs centered around compatibility and full support, just in case of the unknown?

Or was it just that you had default to disable it for speed just because there were so few ROMs that actually needed it, compared to those that don't, for a smaller INI file?

Lithium64:
In my opinion this isn't necessary, because even intel atom powered netbooks can run all games on fullspeed.

Frank-74:
My Atom netbook cannot run all games on fullspeed with software depth buffer enabled.  I get many slowdowns.

Lithium64:
@Frank-74 Do you manually put your amount of VRAM on the gui settings?

Glide64 automatic detection of VRAM doesn't work properly.

Frank-74:
@Lithium64 Yes, I have 1024MB in the VRAM gui settings.  Perfect Dark intro slows down to 30VI's at the start, and hovers around 40-45, brief moments of 60VI's, at 640x480 resolution.  To be honest, turning off software depth buffer doesn't help that much.  Only Jabo's 1.7 can run the intro at full speed.  And with corona effects/lights if I go into options and toggle Software Rendering on then off.

Frank-74:
Just updated to latest Intel HD Graphics with Intel's driver update utility.  Now PD slows down to 20VI's on the intro.  So, this is the last time I'm ever going to update my drivers to the latest version.  99.999999% of the time, your better off with older drivers.  Shit, now this tablet is 'fing useless, and I can't install older drivers.

AmbientMalice:
Software depth buffer is a CPU overhead. Combined with other overheads, it can be the difference between smooth framerates and chugging on a low-end CPU. Games are far more likely to need FB to RAM, which Glide64 has disabled by default. Most N64 games do not need software depth buffer.

AmbientMalice:
>Shit, now this tablet is 'fing useless, and I can't install older drivers.

Why?

Frank-74:
> Why?

Intel update utility totally deleted my old old drivers so I can't rollback.  I should've used device manager>update drivers instead.  Can't find older intel hd graphics drivers for win8.1 with bing.  Version I have now is 10.18.10.4226.  My original version was .37.. something, not 42...

Frank-74:
I think having an onboard GPU with crap software OGL is the bottleneck.  Not the CPU.  No Glide64 or GLideN64 for me at the moment.

cxd4:
With the things glide does, yeah.

If you can't find or recover an older version of your particular Intel card's drivers, I guess you can only wait to see if they ever put out a newer release to fix it.  Hopefully it's just a matter of Intel **accidentally* breaking these things and just being in Microsoft's pocket, rather than trying to make the whole ecosystem worse.

Otherwise, you'll be lucky if you can get a build of mesa with the Wayland accelerator build in to it.  Several Linux distros just use that by default instead of proprietary drivers released by Intel, NVIDIA or whoever else, so it would be easiest to just run Glide on Linux than it is to try compiling that whole thing.

PR #659, opened 2015-10-05
(Glide64) Fix Bug's Life\Toy Story 2 depth issues.

AmbientMalice:
I was reading https://github.com/gonetz/GLideN64/issues/743 and stumbled across the hack needed to fix the issue(s) affecting these two titles. Toy Story 2's laser charge bar not rendering may seem minor, but it was a niggle nonetheless.

Kudos to @oddMLan for pointing out the functionality of the zmode_compare_less hack.


oddMLan:
Oh yeah, I forgot that hack wasn't included in the current Glide64.rdb, nice catch

PR #661, opened 2015-10-09
Set CN64System thread priority to TIME_CRITICAL

Frank-74:
Changes CN64System::StartEmulationThread priority from Normal to Time Critical.

### DIFF
--- a/Source/Project64/N64 System/N64 Class.cpp
+++ b/Source/Project64/N64 System/N64 Class.cpp
@@ -395,6 +395,7 @@ void  CN64System::StartEmulation   ( bool NewThread )
 
 void CN64System::StartEmulationThread (  ThreadInfo * Info ) 
 {
+	SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_TIME_CRITICAL );
 	CoInitialize(NULL);
 	
 	EmulationStarting(*Info->ThreadHandle,Info->ThreadID);


AmbientMalice:
Just a thought - isn't "Time Critical" considered dangerous because it hogs resources and can badly jam up PCs if something goes wrong?

LegendOfDragoon:
>Just a thought - isn't "Time Critical" considered dangerous because it hogs resources and can badly jam up PCs if something goes wrong?

I think it might be dangerous.

AmbientMalice:
What would happen if https://github.com/project64/project64/blob/315027b7fb36fa1389eb4cf25a60c3e0af8b4b8a/Source/Project64/Plugins/Audio%20Plugin.cpp#L182 were set to "Normal" instead of Time Critical? My understanding is that in most scenarios, programs should use the same level of priority for threads and let Windows sort things out. Setting to above normal is generally considered okay, though.

LegendOfDragoon:
Audio threads are often set to higher priority.

AmbientMalice:
>Audio threads are often set to higher priority.

Yea, but TIME CRITICAL could be a potential problem.

Frank-74:
I set it to Time Critical because the Audio was set at that.

I'll close this PR then and try setting every thread to the same level.  Any lower than highest priority though, and I get lag from windows processes, like wiggling the mouse is enough to slow the video visibly when in full screen.  I noticed this testing Cruisin' World on the rotating logo.

LegendOfDragoon:
@Frank-74 I understand why you set it to time critical, I am just not sure it's best for everyone to use it. Maybe making it an option in the RDB would be better, while keeping normal the default value.

>Yea, but TIME CRITICAL could be a potential problem.

Good point. Maybe this should be looked into.

Frank-74:
I've read up a bit on threads, and time critical just works a bit different to the other settings.  The parent thread class determines the base level priority of the other thread settings.  Any setting other than time critical adjusts the priority on demand.  Time critical keeps the priority static/constant, at level 15.

Realtime priority is what can cause problems.  It has a base level of 31.

I've noticed that CAudioPlugin::AudioThread is only ever active with Jabo's DSound, and then uses only 0.02 CPU.  When using Azimer's, CAudioPlugin::AudioThread is never used.
azisoundthread

Frank-74:
I'll work on making it an RDB setting.

Puzzled as to why Azimer plugins don't use the CAudioPlugin::AudioThread.

LegendOfDragoon:
>I've read up a bit on threads, and time critical just works a bit different to the other settings. The parent thread class determines the base level priority of the other thread settings. Any setting other than time critical adjusts the priority on demand. Time critical keeps the priority static/constant, at level 15.

Good to know.

>Puzzled as to why Azimer plugins don't use the CAudioPlugin::AudioThread.

AiUpdate is optional, so you can either create your own thread and use that, or rely on AiUpdate.

>I've noticed that CAudioPlugin::AudioThread is only ever active with Jabo's DSound, and then uses only 0.02 CPU. When using Azimer's, CAudioPlugin::AudioThread is never used.

Zilmar's Audio and 1964 Audio also use it.

Frank-74:
@LegendOfDragoon What may be a problem is dsound.dll!DirectSoundCaptureEnumerateW.  It sets itself to 14 which has a base and thread priority of 22.  Can't find a way to set it to Time Critical (15).  I force it in Process Hacker, but it keeps resetting itself to 22.

Frank-74:
I've made it an RDB setting, but just realized it doesn't need to be set for individual games.  Need to change it to an application setting so that "EmulationTimeCritical=1" in the Project64.cfg [default] section switches it on globally.

AmbientMalice:
@Frank-74 How many CPU cores does your PC have?

Frank-74:
@AmbientMalice 4 CPU cores.  Intel(R) Atom(TM) CPU  Z3735F @ 1.33GHz.  2 Gb of RAM.

Frank-74:
I've just done a PR that makes it an application option.  To turn it on you need to add EmulationTimeCritical=1 under the [default] section in Project64.cfg.
https://github.com/project64/project64/pull/664

AmbientMalice:
The reason I ask is because Time Critical is extremely dangerous on single core CPUs -- in fact, the tendency to cause lockups is one reason why thread priority tampering is so frowned upon.

Frank-74:
@AmbientMalice 
Well I think its 10 years since the last time a single core CPU was manufactured.

I don't like Microsoft tampering with the priority of the emulated N64 Cpu, recompiler etc.  I've watched the thread priority jump between 10 and 14 while its running.  And Microsoft threads always take priority away from PJ64.  Its why wiggling the mouse slows the render window down.  Time Critical locks it to 15 and speed doesn't alter at all.

Its an optional setting anyway, to help performance on low powered netbooks/laptops/pc's.

JunielKatarn:
I own an Intel Compute Stick. Very low-end, and PJ64 lags a lot. I'll give it another try after this change gets merged.

PR #681, opened 2015-10-22
Remove SDL dependency from TexCache unit (Win32 libs fix).

cxd4:
Okay, prior to making this pull request I still cannot compile Glide64 x64 (Debug Win32 works but not Release Win64 for some reason), due to this single linker error only:
```
error LNK2019: unresolved external symbol SDL_Delay referenced in function
"void __cdecl TexCache(void)" (?TexCache@@YAXXZ)
```

The dependency issue with libraries like SDL is a bit further of an inconvenience on Windows than it is with other platforms (which sometimes may include SDL with the operating system, so not as much reason to avoid it), but in this case I would like to begin by removing SDL from `TexCache.cpp` so that I can compile it at all to begin with before commencing any further tests.

### DIFF
--- a/Source/Glide64/TexCache.cpp
+++ b/Source/Glide64/TexCache.cpp
@@ -539,11 +539,10 @@ void TexCache ()
       if (ghq_dmptex_toggle_key) {
         DisplayLoadProgress(L"Texture dump - ON\n");
         ClearCache();
-        SDL_Delay(1000);
       } else {
         DisplayLoadProgress(L"Texture dump - OFF\n");
-        SDL_Delay(1000);
       }
+      SDL_Delay(1000);
     }
   }
 #endif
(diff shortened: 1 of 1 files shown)


cxd4:
Seems that was all I needed to change for 64-bit PJGlide64 to work.
lolnope

cxd4:
If anyone has doubts about calling kernel32 `Sleep()` instead of `SDL_Delay`,
then please refer to the contents of `SDL2-2.0.3\src\timer\windows\SDL_systimer.c`:
```c
void
SDL_Delay(Uint32 ms)
{
    Sleep(ms);
}
```
See also:
http://stackoverflow.com/questions/23258650/sleep1-and-sdl-delay1-takes-15-ms
http://www.gamedev.net/topic/620796-sdl-delay-directx-and-opengl/

which confirm that `Sleep` is the Win32 equivalent.

PR #684, opened 2015-10-25
Fix manifest file generation (themed window controls again)

oddMLan:
With these changes we'll get themed window buttons and controls again. This was broken way back where an option to build as x64 was added into the solution. This is a more universal approach in contrast to the pull requests I made way back in February. We'll let Visual Studio generate one and include Common Controls 6.0 which is required to get a prettier UI.

Should work for both 32-bit and 64-bit releases (tested both).


oddMLan:
Damn, I think I messed up the pull request, is the first commit supposed to be there? That was when I tried to sync my fork with the upstream. If that's an inconvenience, I can close this PR and make a new one.

cxd4:
The first commit could have been avoided, but the overall changes look clean.
https://github.com/project64/project64/pull/684/files

I don't know anything about manifest though.

oddMLan:
Ok since I messed up adding the first gigantic commit, I'll close this PR and make one when I have the time

PR #688, opened 2015-10-26
Fix linker error due to missing LanguageSelector files in VXCPROJ

JunielKatarn:
Building with VS2010 or later, there is a linker error (not finding CLanguageSelector.Select()).

Adding LanguageSelector.h and .cpp to Project64.vcxproj fixes this.


tony971:
This gets x64 building completely in VS2015.

PR #697, opened 2015-10-27
Fix settings menu

### LINKED ISSUE #689: (VS2015) Opening settings menu crashes emulator
Happens with all build configurations. Release builds simply close the window with no warning. Debug x32 gives the following error:

    Debug Assertion Failed!
    
    Program: ..\Project64.exe
    File: C:\Program Files (x86)\Microsoft Visual Studio 14.0\atlmfc\include\atlwin.h
    Line: 3191
    
    Expression: 0

JunielKatarn:
Fixes #689
See https://github.com/project64/project64/blame/master/Source/Project64/Plugins/Plugin%20List.cpp#L15
After that change, the (stdstr &) cast throws an exception.

This commit fixes the menu for both Win32 and x64 platforms.

### DIFF
--- a/Source/Project64/Plugins/Plugin List.cpp
+++ b/Source/Project64/Plugins/Plugin List.cpp
@@ -98,7 +98,7 @@ void CPluginList::AddPluginFromDir ( CPath Dir)
 			}
 
 			Plugin.FullPath = Dir;
-			Plugin.FileName = ((stdstr &)Dir).substr(((stdstr &)m_PluginDir).length());
+			Plugin.FileName = Dir.GetDirectory().substr(m_PluginDir.GetDirectory().length());
 
 			if (GetProcAddress(hLib,"DllAbout") != NULL) 
 			{


tony971:
This allows the settings menu to open but plugins are not initialized when loading a ROM. Every ROM load will give the same two error messages:

    Cannot open a ROM because plugins have not successfully initialized.

followed by

    CN64System::EmulationStarting: Exception caught
    File: N64 System\N64 Class.cpp
    Line: 267

Switching plugins also doesn't do anything. While trying to switch from x64 Glide to x64 HatCat/Angrylion, hitting the apply button reverted the selection to x64 Glide.

cxd4:
>Also, switching plugins also doesn't do anything. While trying to switch from x64 Glide to x64 HatCat/Angrylion, hitting the apply button reverted the selection to x64 Glide.

This is a regular problem I often get with either x86, x64 or ghost processes of PJ64 freezing the config file, so I doubt it's specific to these changes.

tony971:
I'm not saying that this change caused the problems. Just that the settings menu isn't entirely fixed by it.

cxd4:
Wasn't talking about the menu; you were talking about the plugin "Apply" button reverting it.

That means the settings file is being locked by another/ghost process; that can't be fixed by code.

tony971:
It didn't happen until recently. If code broke it, code can fix it.

cxd4:
>It didn't happen until recently.

Wrong.  It's **been** happening ever since 1.7.0.50.

Code didn't broke it; you broke it by leaving a ghost process locking the configuration settings file.  You only just now noticed this "code defect" because you only just now started to attempt overwriting read-only file permissions, which is not something the operating system will let you do.

tony971:
I started with freshly compiled everything and no saved settings. This is the first time I've experienced the issue. There were no custom settings I used and nothing residual about my setup. The files weren't placed in program files and I have full permissions. Doing this same process in the past never resulted in not being able to change the plugins. This is a new issue.

cxd4:
It's not about having full permissions dude.

I don't use program files either and go figure I use an admin account.
Still doesn't change the fact that this "issue" has been there since the beginning.  It's a nub fail.

tony971:
So this time I tried something new. I started off with freshly compiled everything (again), put it in an isolated test folder (again), let new configuration folders be generated by first time use (again) but this time I selected your angrylion fork as the initial plugin (no graphics plugin was listed before). It still switched to Glide on application.

cxd4:
Besides what I said, it's possible for a commit to also write wrong/no values to the CFG.

Any idea when this settings crash regression was introduced @JunielKatarn?

JunielKatarn:
I'm afraid not. My recent changes don't tamper with CFG files, or plugin loading.
All I can say is I'm also experiencing them (both x86 and x64). No stall processes, no "dirty" config dirs.

tony971:
nub fail.

cxd4:
I just meant the ATL assert that crashed the EXE.

PR #698, opened 2015-10-27
[N-Rage] Disabled broken pre-processor code.

cxd4:
I have no idea what this was for...

This "hack for GNU C compilers" was not present in N-Rage 1.83 original sources.
http://members.chello.at/n-rage/dinput8/downloads.html

So either rabiddeity or squall-leonhart or somebody added this after N-Rage 2.00.

N-Rage almost compiles now on MinGW 64-bit already.  This is one of the only obstacles there is, so unless somebody can speak for what this macro code might be intended to be used for...I'd say it's part of the broken code unique to N-Rage 2.3 where we should have used N-Rage 2.1 codebase instead.

### DIFF
--- a/Source/nragev20/settings.h
+++ b/Source/nragev20/settings.h
@@ -26,7 +26,7 @@
 
 #define DIRECTINPUT_VERSION 0x0800
 
-	// hacks for GNU C compilers
+// hacks for GNU C compilers // No thank you....
 #ifdef __GNUC__
 
 #ifndef WINVER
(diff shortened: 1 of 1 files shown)


cxd4:
If I may confer with the N-Rage porting efforts of @SwooshyCueb, he had to make the same change.
https://github.com/SwooshyCueb/project64/commit/20c2cc1450cb1c9a89e9e516326d1ed1bd773e54#diff-9e8646b30fa99e81d7212f579fd3584dR40

Too bad squall or whoever isn't here to defend whatever introduced these lines of code.

PR #701, opened 2015-10-27
[N-Rage] fixed broken FormatMemPak in 64-bit

cxd4:
Maybe in MSVC this isn't a compiler error, but in GCC's implementation of the C++ standard it is.

If `sizeof(void *) > sizeof(ULONG)`, then `(ULONG)pointer_to_any_type` is an error.

Even if it's not a compiler error for Visual Studio compilers, I'm sure it fixes a broken feature in 64-bit builds of N-Rage for pj64.

### DIFF
--- a/Source/nragev20/PakIO.cpp
+++ b/Source/nragev20/PakIO.cpp
@@ -800,17 +800,19 @@ inline WORD CountBlocks( const unsigned char * bMemPakBinary, LPBYTE aNoteSizes 
 
 void FormatMemPak( LPBYTE aMemPak )
 {
-	FillMemory( aMemPak, 0x100, 0xFF );
+	size_t iRand, n;
 
+	FillMemory(aMemPak, 0x100, 0xFF);
 	aMemPak[0] = 0x81;
 
 	// generate a valid code( i hope i can calculate it one day)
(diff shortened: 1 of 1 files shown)


cxd4:
```
project64\Source\Script\MinGW\..\..\nragev20\PakIO.cpp:
In function 'void FormatMemPak(LPBYTE)':
project64\Source\Script\MinGW\..\..\nragev20\PakIO.cpp:811:24: error:
cast from 'LPBYTE {aka unsigned char*}' to 'ULONG {aka long unsigned int}' loses precision [-fpermissive]
  int iRand = (( (ULONG)aMemPak / 4 + (ULONG)g_strEmuInfo.hMainWindow / 4 ) % (sizeof(aValidCodes)/8) );
                        ^
project64\Source\Script\MinGW\..\..\nragev20\PakIO.cpp:811:58: error:
cast from 'HWND' to 'ULONG {aka long unsigned int}' loses precision [-fpermissive]
  int iRand = (( (ULONG)aMemPak / 4 + (ULONG)g_strEmuInfo.hMainWindow / 4 ) % (sizeof(aValidCodes)/8) );
                                                          ^
```

PR #705, opened 2015-10-27
[Glide64] Add missing <gccpriv.h> dependency for GCC.

cxd4:
Not sure why the original Glide64 package didn't include this header (perhaps because Gonetz never tested on MinGW but only native Linux GCC?), but it is indeed a part of the WX/MSC package and is needed to compile Glide64 on MinGW.

Without it we have this error for 12/14 of the C++ source files:
```
In file included from $project64\Source\Script\MinGW\..\..\3rd Party\wx\include/wx/defs.h:21:0,
                 from $project64\Source\Script\MinGW\..\..\3rd Party\wx\include/wx/wx.h:15,
                 from $project64\Source\Script\MinGW\..\..\Glide64\winlnxdefs.h:45,
                 from $project64\Source\Script\MinGW\..\..\Glide64\Gfx_1.3.h:65,
                 from $project64\Source\Script\MinGW\..\..\Glide64\Main.cpp:40:
$project64\Source\Script\MinGW\..\..\3rd Party\wx\include/wx/platform.h:577:32:
fatal error: wx/msw/gccpriv.h: No such file or directory
 #    include "wx/msw/gccpriv.h"
                                ^
```

However after https://github.com/project64/project64/pull/676 which got Glide64 compiling as 64-bit, much of Glide64 works on MinGW straight away.  Only a handful of tweaks are needed to compile PJGlide64 without using MSVC, such as this one.  I will get to the rest in time.

### DIFF
--- a/Source/3rd Party/wx/include/wx/msw/gccpriv.h
+++ b/Source/3rd Party/wx/include/wx/msw/gccpriv.h
@@ -1,0 +1,127 @@
+/*
+ Name:        wx/msw/gccpriv.h
+ Purpose:     MinGW/Cygwin definitions
+ Author:      Vadim Zeitlin
+ Modified by:
+ Created:
+ RCS-ID:      $Id: gccpriv.h 36155 2005-11-10 16:16:05Z ABX $
+ Copyright:   (c) Vadim Zeitlin
+ Licence:     wxWindows Licence
+*/
+
+/* THIS IS A C FILE, DON'T USE C++ FEATURES (IN PARTICULAR COMMENTS) IN IT */
+
+#ifndef _WX_MSW_GCCPRIV_H_
+#define _WX_MSW_GCCPRIV_H_
+
+#if defined(__MINGW32__) && !defined(__GNUWIN32__)
+    #define __GNUWIN32__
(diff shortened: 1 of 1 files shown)


JunielKatarn:
I understand we need to be able to use other compiler collections for non-Windows platforms.
However, do we know for sure there will be no performance/functionality hits on the emulator if we don't use MSVC for Windows?

cxd4:
Me, personally, I do know for sure, yes, because many other projects (some of which mine) optimize the best with non-MSVC compilers (not necessarily gcc/mingw) as proven overtime by several benchmarks, asm analyses and other tests, as well as some notes I've taken over time of some absurd optimization strategies that MSVC uses with things like branch prediction.

But, all that being besides the point.
I will not attempt to encourage anyone to use non-MSVC if they just want a Windows build.
I myself use MSVC for many of my projects, due to linker advantages over using MinGW.

Functionality hits?  Maybe, because MinGW is infamous for breaking post-ANSI features and other things.  I see it as a challenge to further increase the portability of Project64's coding, so any said functionality hits ought to be fixed simply within the code.

PR #709, opened 2015-10-28
Add PJGlide and NRage MinGW compile scripts?

cxd4:
Hmm...I don't want to pollute the repository (hence the directory choice of Scripts/MinGW) space/size so can understand if this pull request should be rejected, but I didn't feel like selfishly keeping the scripts to myself either in case other people wanted to use them / help port the emulator to Linux etc.

The scripts aren't really big, but they are sort of experimental.  They do however compile both Glide64 and N-Rage without warnings or errors.  (I can think of a few people who would complain that I am not using GNU `Makefile` files instead of batch scripts....)  I do have a handful of linker namespace issues to tend to, but that can come later.  So I figure if people don't have the 10-20 GB of Visual Studio version X installed but wanted to test compile their changes it could be useful.

So any comments on the new scripts / should we not have them in the repo etc.?

### DIFF
--- a/Source/Script/MinGW/glide64.cmd
+++ b/Source/Script/MinGW/glide64.cmd
@@ -1,0 +1,81 @@
+@ECHO OFF
+TITLE MinGW Compiler Suite Invocation
+
+set src=%CD%\..\..\Glide64
+set obj=%CD%\Glide64
+
+if not exist %obj% (
+mkdir %obj%
+)
+
+    set MinGW=C:\MinGW
+REM set MinGW=C:\msys64\mingw64\x86_64-w64-mingw32\..
+
+set FLAGS_x86=^
+ -I"%src%\.."^
+ -I"%src%\..\3rd Party\wx\include"^
+ -I"%src%\..\3rd Party\wx\lib\vc_lib\msw"^
+ -I"%src%\..\Glitch64\inc"^
+ -S^
+ -masm=intel^
+ -march=native^
+ -Os
+
+set C_FLAGS=%FLAGS_x86%
+
+cd %MinGW%\bin
+set CC=%MinGW%\bin\g++.exe
+set AS=%MinGW%\bin\as.exe
+
+ECHO Compiling Glide64 plugin sources...
+%CC% -o %obj%\Main.asm                  %src%\Main.cpp %C_FLAGS%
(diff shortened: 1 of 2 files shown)


cxd4:
Also I haven't implemented [pre-compiled headers](https://gcc.gnu.org/onlinedocs/gcc/Precompiled-Headers.html) yet, so both projects compile slowly.  N-Rage isn't so bad...Glide however I could definitely do with some pre-compiled headers in the future.

aschwant:
I think they look fine. What is the reason you are not going with makefiles and msys? A makefile would reduce the repeated compile and assemble commands.

cxd4:
My lightweight, simplified installation of MinGW doesn't include support for `make` or makefiles.

While I'm no fan of their syntax (no matter how small or how big), they are a necessary instrument for convincing other collaborators to want to join and help further the project.  I can only encourage others to add `Makefile` files into the repository's Glide64, N-Rage, RSP and Project64 source directories, as this is more modern than shell scripting, but I myself will not do this.  My scripts can help do this, though.

I do have a full, automated 32-bit-only MinGW install which has support for makefiles, but in practice I prefer to use batch scripts because I can follow them much more directly and have not really found anything to gain from using `Makefile` syntax (either CMake, GNU makefile, SGI makefile, Microsoft make...insert X style of makefile here; I just don't find any reason to care because scripts have done everything flexibly enough for me so far without requiring the extra dependencies.)

>A makefile would reduce the repeated compile and assemble commands.

I agree the repetition of my compile commands is starting to bother me.

I'm beginning to become annoyed with my copy-paste "Main.cpp", `gcc main.asm main.cpp`, `as.exe main.o main.asm`, `ld.exe glide.dll main.o ...` etc.  This repetitiveness does make the pattern easy to follow at least so I can still read and maintain them better, though updating may be tedious.

The good news is someone took a similar batch script I did with AziAudio and modernized it:
https://github.com/Azimer/AziAudio/blob/master/Build/mingw-build.cmd

This proves that the tedious copypasta repetition in batch files can be reduced.

The bad news is that it's so full of Win32-shell-specific features, I can no longer read it.

If we were to decide to merge these scripts in this PR I would appreciate it if somebody had a way to simplify my copypasta of the file names using their knowledge of the Windows command shell, but keep it easy to follow please.

Issue #704, opened 2015-10-27
Cannot change plugins


tony971:
Regardless of which plugins you choose, PJ64 will decide to use a single set.

aschwant:
This is fixed with #713.

Issue #703, opened 2015-10-27
Plugins not initialized


tony971:
Not sure if it's only with executables built with VS2015, but plugins fail to initialize on ROM load.

JunielKatarn:
Same here.
Tested with a clean build, right from master branch, latest commit (as of this writing).

This bug was introduced within the last 48 hours.

JunielKatarn:
I just verified it also desn't work with VS2008.
It's safe to assume it's not VS version-dependent.

cxd4:
lol.
Which is why I wonder When did all of these regressions pop up out of the blue??

But even if I spend my time isolating and fixing it, zilmar is busy restructuring much of the entire Project64 core-UI modulation.  I just figured it's safer to wait for that first then start touching the core again.

JunielKatarn:
I'm sure it shouldn't be too hard to track, but I'm busy at the moment.
If no one fixes it before, I might take a look over the weekend. (Any volunteers? XD)

cxd4:
If time goes on then one of us fixes it in the end.

tony why don't you play around with `git checkout #######` and re-build from each GitHub commit hash number and isolate it?  As you seem to have a lot of time for testing.

JunielKatarn:
@tony971: I would start with the latest commit available 3 days ago (meaning, pushed to the main repo, NOT commit date).

Then, do a binary search approach. Shouldn't take that long :P

Frank-74:
Maybe test before/after this commit 3 days ago?
https://github.com/project64/project64/commit/0144305c6a50e88158c09c76a56126ae8f5f3449

aschwant:
This is fixed with #713.

PR #713, opened 2015-10-30
Fixed incorrect file name causing plugins to fail to load

aschwant:
This is a fix for #703 and #704. `Plugin.FileName` was being set to an invalid string in 74355d34d8bddc4f93a94bce32bb44482ab0114f.


cxd4:
Good job!

JunielKatarn:
After this fix, Project64 builds and runs both x86 and x64 (Using @cxd4 no sound plugin) for Windows :)

PR #718, opened 2015-10-30
[RSP] Execute recompiler CPU without SEH for non-MSVC.

cxd4:
SEH is the last of the remaining compiler errors.

There are no more compiler errors with MinGW after this pull request, just linker errors.

I moved the `memset()` on a whim that it shouldn't be able to generate exceptions within the __try __except block, mostly so I could copy-paste only 3 lines of code instead of 4 repeated lines.


cxd4:
>[RSP] This memset() should be able to cause exceptions.

I meant to say "shouldn't" be able to cause exceptions.
Not sure how to rename commits.  Oh well.

PR #720, opened 2015-10-30
Fix manifest file generation (themed window controls again)

oddMLan:
(Redid PR: Copied description from previous PR #684 )

With these changes we'll get themed window buttons and controls again. This was broken way back where an option to build as x64 was added into the solution. This is a more universal approach in contrast to the pull requests I made way back in February. We'll let Visual Studio generate one and include Common Controls 6.0 which is required to get a prettier UI.

Should work for both 32-bit and 64-bit releases (tested both).


oddMLan:
**WARNING:** It makes Project64 look really slick on Windows 10

Before these changes
before

After
after

jjjustinleblanc:
i sure hope this gets merged soon.. may not be "important" per se, but god damn the legacy interface is ugly

cxd4:
haha.  I'm not picky one way or the other how the GUI is done, but the changes themselves look like cleaner code if the manifest file is being voided out anyway.  So I wouldn't know, looks just fine to me.

PR #729, opened 2015-11-06
[Common] Use size-neutral types.

cxd4:
Somewhat (but not really) related to https://github.com/project64/project64/pull/728 .

Here, again, `int32_t` is a completely random and arbitrary preference of parameter type, as the relevant functions have no pertinence to the necessity of having 32-bit precision (let alone exactly that).

What happens if you're compiling for a system whose `Sleep()` function equivalent, for example, only accepts a 64-bit (or a 21-bit, or 16-bit) unsigned integer for the timeout counter in milliseconds?  This choice of `uint32_t` is still partly biased to the Windows `Sleep()` function and is not really portable.

A better choice would be the neutral type `unsigned long`, but `size_t` would be understandable.

### DIFF
--- a/Source/Common/SyncEvent.cpp
+++ b/Source/Common/SyncEvent.cpp
@@ -15,9 +15,9 @@ void SyncEvent::Trigger()
     SetEvent(m_Event);
 }
 
-bool SyncEvent::IsTriggered(int32_t iWaitTime)
+bool SyncEvent::IsTriggered(size_t iWaitTime)
 {
-    return (WAIT_OBJECT_0 == WaitForSingleObject(m_Event,iWaitTime));
+    return (WAIT_OBJECT_0 == WaitForSingleObject(m_Event, iWaitTime));
 }
 
 void SyncEvent::Reset()
--- a/Source/Common/SyncEvent.h
+++ b/Source/Common/SyncEvent.h
@@ -9,7 +9,7 @@ public:
     ~SyncEvent(void);
 
     void Trigger (void);
-    bool IsTriggered (int32_t iWaitTime = 0);
+    bool IsTriggered(size_t iWaitTime = 0);
     void Reset();
     void * GetHandle();
 
--- a/Source/Common/Util.cpp
+++ b/Source/Common/Util.cpp
@@ -2,7 +2,7 @@
 #include "Util.h"
 #include <Tlhelp32.h>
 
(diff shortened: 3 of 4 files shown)


project64:
in which case i will use a #define to handle the different os, I prefer standard types where possible over say unsigned long, now I do not mind size_t caus it is not fixed between 32bit and 64bit and want the value changed, not so much here.

cxd4:
>now I do not mind size_t caus it is not fixed between 32bit and 64bit and want the value changed, not so much here.

What does that mean?  You want the value changed "not so much here", or that you would rather use `size_t` over standard types such as `unsigned long`?

>in which case i will use a #define to handle the different os

If the different os has a Sleep() function, then just call Sleep() without macros like #ifdef WIN32 Sleep(uint32_t) #elif ... Sleep(uint64_t) #elif ... Sleep(uint16_t) ... it should just be a standard type like unsigned int/long that is not fixed between 32-bit and 64-bit, like uint32_t is.

project64:
well I doubt sleep value is ever over a 32bit value anyway ... I was more refering to size_t works better in for 32/64 builds since it will act different

cxd4:
Which is why, in this case, I'm not sure whether you prefer `size_t` or `unsigned long`.

Either one acts different in 32- and 64-bit builds (as `size_t` usually typedef'd to an `unsigned long` anyway, except on Windows), so either one is an improvement over `uint32_t`, which has very little to do with milliseconds or Sleep interval precisions because 32 bits is irrelevant to Physics class.

The only reason I can think of for making it a `uint32_t` is if the MIPS hardware has a 32-bit limit on what the timer for sleeping can be or something, even though it has 64-bit registers.  Other than that, why uint32?

>well I doubt sleep value is ever over a 32bit value anyway

For the sake of not arguing this, I'll assume nobody will care to allow > 32-bit time units / inputs.

If it's *under* a 32-bit value, you would still have to #ifdef inside the function to decide how to convert the uint32_t into a uint16_t or whatever anyway.  This would be no different if the source was a 32-, 64-, 48-, 57-, etc.-bit `unsigned long` of whatever width.  The difference is that `sizeof(unsigned long) >= sizeof(uint32_t)`, so the `long` has a greater chance of not losing information than the uint32_t does.

cxd4:
*burp*

PR #730, opened 2015-11-08
Remove Sram Size

Nekokabu:


Lang/Norwegian.pj.Lang
```diff
@@ -430,4 +430,4 @@ Skift plugin og restart spill nå?"
 #2043#  "Det kode navnet er allerede i bruk"
 #2044#  "Du har nådd det maksimumme antall koder for denne ROM'en"
 
-//The End
\ No newline at end of file
+//The End
```
cxd4: Thank you!!  Finally somebody here who understands that lines end with line breaks.

PR #733, opened 2015-11-09
Set default toolset to MSVC 14

JunielKatarn:
Sets the default tool set to the latest freely available Visual Studio version (2015).
Besides technical reasons, Microsoft no longer provides direct download links to version 2013 Community (see https://www.visualstudio.com/products/visual-studio-community-vs).

If you prefer to override this with your preferred/installed version, do either:

* Command line:
  * `MSBuild.exe Project64.sln /p:PlatformToolset [aditional optional parameters...]`
* Globally overriding the value for your user (.props file):
  * Edit file `C:\Users\YourUser>\AppData\Local\Microsoft\MSBuild\v4.0\Microsoft.Cpp.Win32.user.props`, adding:<br/>
```
	<PropertyGroup Label="Configuration">
		<PlatformToolset>v120_xp</PlatformToolset><!-- Or desired tool set -->
	</PropertyGroup>
```


cxd4:
Sounds good to me.  I'll update x64.user.props as well assuming the same rules apply for Win64.

If and when I update to VS2015 or what comes after that, I'll just have to be sure to remember to delete my changes to these files in the future so that VC2015+ compiles will work right.  It at least beats me having to `git pull` and edit Platform.props every time I pull from the repository when that file got updated.

PR #734, opened 2015-11-09
Use a standard window for the About dialog

toehead2001:
about


tony971:
About Windows 10

Looks great on Windows 10.

cxd4:
Any thoughts @exhalatio (or zilmar) ?  Does this help solve the issue you were thinking of?

Maybe other improvements missing to add on to this if it got merged?

Source/Project64/User Interface/Gui Class.cpp
```diff
@@ -1019,24 +983,18 @@ DWORD CALLBACK AboutBoxProc(HWND hWnd, DWORD uMsg, DWORD wParam, DWORD lParam)
...
 	case WM_INITDIALOG:
 	{
-		enum { ROUND_EDGE = 15 };
-
-		DWORD dwStyle = GetWindowLong(hWnd, GWL_STYLE);
-		dwStyle &= ~(WS_CAPTION | WS_SIZEBOX);
-		SetWindowLong(hWnd, GWL_STYLE, dwStyle);
+		//Title
+		LONG_PTR originalWndProc = GetWindowLongPtrW(hWnd, GWLP_WNDPROC);
+		SetWindowLongPtrW(hWnd, GWLP_WNDPROC, (LONG_PTR)DefWindowProcW);
```
project64: you should not set the wndproc create the dialog with the W version

MELERIX:
> Looks great on Windows 10.

that is because Windows 10 RTM does not have colours in the title bar and the borders of windows, but now in Windows 10 Insider Preview they have added colours for the title bar and the borders of windows.

Source/Project64/User Interface/Gui Class.cpp
toehead2001: You mean change it to this?

`LONG_PTR originalWndProc = GetWindowLongPtr(hWnd, GWLP_WNDPROC);`
`SetWindowLongPtr(hWnd, GWLP_WNDPROC, (LONG_PTR)DefWindowProc);`

project64: I mean in CMainGui::AboutBox
it should be

DialogBoxParamW(GetModuleHandle(NULL), MAKEINTRESOURCEW(IDD_About), m_hMainWindow, (DLGPROC)AboutBoxProc,(LPARAM)this);

toehead2001: oh... That makes so much more sense. Lack of grammar can be so ambiguous :blush:

toehead2001:
Does the pull look ok now?

Source/Project64/User Interface/Gui Class.cpp
cxd4: glad I'm not the only one bad at reading his grammar, lol

project64:
looks fine

PR #736, opened 2015-11-09
CTLB:  Replace `int' with `size_t'.

cxd4:
To be awfully pedantic, `int` is not the best choice of type here.

1.  `sizeof()` operator yields the native source type `size_t`, which is not signed and larger than `int`.
2.  CPU memory limits such as the *n*th index into an array[] for `array[n]` are indicated by the standard type `size_t`, which defines the maximum index into such an array.

tl;dr:  `size_t` is better than `int` is better than `int32_t`, in order of least likely to break code. :)

### DIFF
--- a/Source/Project64/N64 System/Mips/TLB class.cpp
+++ b/Source/Project64/N64 System/Mips/TLB class.cpp
@@ -303,7 +303,8 @@ void CTLB::RecordDifference( CLog &LogFile, const CTLB& rTLB)
 
 bool CTLB::operator == (const CTLB& rTLB) const
 {
-    for (size_t i = 0, n = sizeof(m_tlb)/sizeof(m_tlb[0]); i < n; i++)
+    const size_t n = sizeof(m_tlb) / sizeof(TLB_ENTRY);
+    for (size_t i = 0; i < n; i++)
     {
         if (m_tlb[i].EntryDefined != rTLB.m_tlb[i].EntryDefined)
         {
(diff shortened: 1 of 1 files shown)


cxd4:
Another change I did was that I rewrote
`n = sizeof(m_tlb) / sizeof(m_tlb[0])`, to `... / sizeof(TLB_ARRAY)`, because I think the type of the array is more useful to see than just repeating the array name with `m_tlb[0]`.

And no I'm not exposing the `n` variable to accidental modification by moving it up to an outer scope from the inner scope in which it's actually used, because I have declared it `const`. :3 So not a variable.

project64:
the reason I do m_tlb[0] is that if the type changes from TLB_ENTRY to something else, if TLB_ENTRY type still exists the code will compile with out error, and if TLB_ENTRY is the same size of the new structure no issue, but if I say create a new structure add a new variable and assign m_tlb to the structure then this code will be wrong and easily left in. So while this is fine it is error prone

cxd4:
That's true and I agree completely.

I knew that already at the time I made the change, but felt that showing the text sizeof(TLB_ENTRY) would help me look up the type more easily without guessing / opening Visual Studio and finding the header file that declared the array with X type.

As you said, `sizeof(m_tlb) / sizeof(m_tlb[0])` is less error-pone.  Technically my argument would be that it makes changing the array's name harder, as it re-specifies an extra copy of the "m_tlb" name and makes that slightly harder to maintain.  But I think what you said matters more overall so I will push that change now.

cxd4:
Normally I would solve the "name" maintenance problem with a macro:

```c
#define SIZE_IN_ELEMENTS(array) (sizeof(array) / sizeof(array[0])
```

so that I would not have to re-state the TLB buffer's name twice with an extra copy.

I supposed that while we're evading macros and pre-processor voodoo as much as possible, we may as well do what macros cannot do and say sizeof(array) / sizeof(type_of_array) which #define cannot handle.  But whatever I like either method.

PR #739, opened 2015-11-10
"fixed" implementation of Controller 1.1 specifications

cxd4:
Thanks very much to zilmar's suggestion, I've modified the PJ N-Rage plugin (which, unbeknownst to me, actually was set to use 1.1 instead of 1.0) to demonstrate the "proper" compliance to the official, publicly released version of the plugin specifications for 1.1.

Note that by "proper" I don't necessarily mean "better", but it is stable and it works.

The problem is that the official 1.1 specs say to use InitiateControllers this way:
```c
EXPORT void CALL InitiateControllers(CONTROL_INFO ControlInfo)
```

This is how Mupen64 implements it.  Unless you are running the Windows port of Mupen64 0.5, there is no 1.0 spec, so the only available plugin specs for controller plugins with mupen is 1.1.

Now in Project64's case, this largely went unimplemented, all the way from PJ64 1.0 to 1.7.
At the last minute, finally Project64 1.7 implements it like this:
```c
EXPORT void CALL InitiateControllers(CONTROL_INFO * ControlInfo)
```

What's with the `*`?  That wasn't there in the official specs.
From what I gathered talking to zilmar over IRC this was meant to be in the official specs.

It may be a typo then, but we don't need the pointer type because `CONTROL_INFO` structure members `HEADER` and `CONTROL` are pointers, not arrays.  Inside of the struct, we already have the pointers to ROM header and controller enabled/mempak/raw_data flags.  Therefore, the pointer variation is unnecessary because we were already given pointers to begin with.

The only certain disadvantage with the non-pointer method is that passing `InitiateControllers(ControlInfo)` probably floods the call stack with a gigantic structure of data, whereas `InitiateControllers(&ControlInfo)` merely sends a pointer register on the call stack which is more efficient.  But as far as stability I think there is no advantage.


cxd4:
Update:  Just for good measure, I tested my keyboard plugin changes on Linux Mupen64 0.5.

Now my input plugin works with both Linux controller 1.1 and Project64 (after this PR's changes are compiled) controller 1.1.  Just to help show this change is stable and isn't some "lucky guess" undefined behavior thing that only works on Windows or whatever.

cxd4:
merge conflicts resolved :P
https://github.com/project64/project64/commit/fbe62d95670a12d1c127e102b418bf3925247f8d#diff-914550a4c9875095df408a75a47c5685L77

JunielKatarn:
Where are these plugin specs published?

PR #740, opened 2015-11-10
"Project 64" -> "Project64"

toehead2001:
Only use "Project64", rather than mix & match "Project 64" and "Project64"


vgturtle127:
:+1:

PR #741, opened 2015-11-10
Fix build after PR #738

JunielKatarn:
A couple of member variables were not correctly changed from uint32_t to int32_t, causing compilation to fail.

Also, there was a typo in Project64.sln that caused builds to be incomplete (likely not detected because of the encoding change).

### DIFF
--- a/Project64.sln
+++ b/Project64.sln
@@ -9,7 +9,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Glide", "Glide", "{9FE699A5-41C3-4441-92AB-639B3D77DE26}"
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Common", "Source\Common\Common.vcxproj", "{B4A4B994-9111-42B1-93C2-6F1CA8BC4421}"
-EndProjectt
+EndProject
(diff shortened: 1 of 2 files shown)


tony971:
Is it possible to drop the first commit to get this merged?

cxd4:
Yep it was a C++ compiler rule problem.

As converting BOOL between uint32_t vs. int32_t has no real issue I really did not care if the C compiler would throw warnings to alert me to change the rest of them to int32_t, then I would change them to be consistent along the way.

In C++'s case it's a compile error, not a warning.  Otherwise I had no reason to re-compile the entire source just to test my change.  With Project64 I tend to find unnecessarily re-compiling the entire package to waste too much of my time, so I just went with the C rules and pushed my commits without testing.  No harm done. :P

cxd4:
>Also, there was a typo in Project64.sln that caused builds to be incomplete (likely not detected because of the encoding change).

Sorry that was my fault.

I put the extra `t` at the end of `Projectt` to test Git's UTF-8 support of the file during my pull request (which zilmar approved) to fix Project64.sln from being shown as a binary file with no visible text changes.  The extra t typo was a temporary change i tested and I **thought** I hit Ctrl+S to save it when I hit Ctrl+Z to undo it but apparently not. :(

JunielKatarn:
Closing. Will submit the SLN change in another PR...

PR #749, opened 2015-11-11
logging class:  fixed likely regression with create/open conditions

cxd4:
This just looks like a bug because before the code said this:
```c
if (nOpenFlags & modeNoTruncate)
    dwCreateFlag = OPEN_ALWAYS;
else
    dwCreateFlag = CREATE_ALWAYS;
```

It was just recently changed this morning to this:
```
dwCreateFlag = nOpenFlags & modeNoTruncate == 0 ? OPEN_ALWAYS : CREATE_ALWAYS;
```
which I am almost positive is not an accurate translation of the original.

For two reasons:

1.  Operator precedence.  `a & b == 0` by default means `(a) & (b == 0)`.
2.  It should be != not ==, otherwise switch OPEN : CREATE to CREATE : OPEN.

### DIFF
--- a/Source/Common/File Class.cpp
+++ b/Source/Common/File Class.cpp
@@ -79,14 +79,12 @@ bool CFile::Open(const char * lpszFileName, uint32_t nOpenFlags)
     sa.bInheritHandle = (nOpenFlags & modeNoInherit) == 0;
 
     // map creation flags
-    ULONG dwCreateFlag = 0;
+    ULONG dwCreateFlag = OPEN_EXISTING;
     if (nOpenFlags & modeCreate)
     {
-        dwCreateFlag = (nOpenFlags & modeNoTruncate) != 0 ? OPEN_ALWAYS : CREATE_ALWAYS;
-    }
-    else
-    {
-        dwCreateFlag = OPEN_EXISTING;
+        dwCreateFlag =
+            ((nOpenFlags & modeNoTruncate) != 0)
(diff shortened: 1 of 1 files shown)


cxd4:
And yes for Hell's sake I did a Rebuild All to test compile and link these changes.  Did I test run the compiled binary?  No because my opinion is that I prefer not being a complete moron and waste time.

It is a straightforward syntax fix.
I honestly don't even care if it BSODs; it's still the better concept.  "Regressions" caused by proper fixes are external bugs from outside, other inaccurate code from outside the commit and should be separate PRs not all-in-one giant PR.

Source/Common/File Class.cpp
```diff
@@ -79,14 +79,12 @@ bool CFile::Open(const char * lpszFileName, uint32_t nOpenFlags)
...
-    ULONG dwCreateFlag = 0;
+    ULONG dwCreateFlag = OPEN_EXISTING;
     if (nOpenFlags & modeCreate)
     {
-        dwCreateFlag = nOpenFlags & modeNoTruncate == 0 ? OPEN_ALWAYS : CREATE_ALWAYS;
-    }
-    else
-    {
-        dwCreateFlag = OPEN_EXISTING;
+        dwCreateFlag =
```
project64: can you put this on one line

cxd4:
Okay done.

Source/Common/File Class.cpp
cxd4: Done.

PR #751, opened 2015-11-13
Increase g_Plugins (and friends) security against null pointers.

cxd4:
It's gotten complicated enough to the point where I think I just stop here.  I imagine there will be enough things as it is already to criticize for suggestions on how to update the PR.

Only reason I did this was to strengthen the guard from referencing plugin functions under a pointer under a pointer under a pointer under a ... anywhere along the line there may have been one or more null pointers.  While this can usually be dismissed as an unlikely possibility not worth cluttering the code for I felt like giving it a quick try--who knows maybe it could come in handy against subtle bugs in the future when someone tries to change something.

It may also have performance gains as well I guess since `g_Plugins` is a global which can be sensitive to constant change in a multi-threaded scenario.  Here I have sometimes declared local variables to store the pointer tree into a single pointer to shorten much of the code checks down to reading one single pointer.

### DIFF
--- a/Source/Project64/N64 System/Mips/Pif Ram.cpp
+++ b/Source/Project64/N64 System/Mips/Pif Ram.cpp
@@ -89,12 +89,26 @@ void CPifRam::n64_cic_nus_6105(char challenge[], char respone[], int length)
 
 void CPifRam::PifRamRead()
 {
+    CControl_Plugin * control_plugin;
+    CONTROL * Controllers;
+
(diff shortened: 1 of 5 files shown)


cxd4:
I don't care too much about this, but I felt like experimenting with that anyway.

My problem with it is it clutters the functions with extra code & variable declarations.

Maybe some part of it can be adopted in the future if and when it turns out to be relevant to these EXE crashes people are having, but I have not gone out of my way to check.

PR #754, opened 2015-11-13
Simplify ROM browser shell file extensions comparisons.

cxd4:
Just something I happened upon by accident while stumbling through the `g_Plugins` business.
We can loop through a string table instead of iterating multiple `if` a && b && c && d && ... .

The second commit is probably the more important of the two, because it fixes what I think to be a redundancy in the extension analysis.  Or why was the original code this?
```c
if (_stricmp(ext2, ".v64") != 0 && _stricmp(ext2, ".z64") != 0 &&
	_stricmp(ext2, ".n64") != 0 && _stricmp(ext2, ".rom") != 0 &&
	_stricmp(ext2, ".jap") != 0 && _stricmp(ext2, ".pal") != 0 &&
	_stricmp(ext2, ".usa") != 0 && _stricmp(ext2, ".eur") != 0 &&
	_stricmp(ext2, ".bin") == 0)
{
```
MSDN's personal [documentation of _stricmp](https://msdn.microsoft.com/en-us/library/k59z8dwe.aspx) agrees that a return value of 0 indicates equality between the string operands, while a nonzero return value indicates inequality.

So instead of str != a && str != b && str != c ... && str == z, why not say only `str == z`?

### DIFF
--- a/Source/Project64/User Interface/Rom Browser Class.cpp
+++ b/Source/Project64/User Interface/Rom Browser Class.cpp
@@ -545,8 +545,13 @@ void CRomBrowser::NotificationCB(LPCWSTR Status, CRomBrowser * /*_this*/)
 	g_Notify->DisplayMessage(5, Status);
 }
 
+static const char* ROM_extensions[] = {
+    "zip", "7z", "v64", "z64", "n64", "rom", "jap", "pal", "usa", "eur", "bin",
+};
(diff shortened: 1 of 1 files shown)


cxd4:
New commit just added:  https://github.com/cxd4/project64/commit/2a17441f522317c6a35ab0e62c0e8b3eafefa72e

Because I made a slight mistranslation of the original code that caused the loop to go on after only one condition test passed (which should have been enough).

Issue #683, opened 2015-10-23
call ID_CURRENT_SAVE_DEFAULT is not translated


project64:
"Save Slot (%s) selected" this text should be translated ... maybe the code needs a way to replace variables .. like should the translated string have %s works fine here but not good if I need to add 2 params

cxd4:
Don't know the details but it looks like it could still be legible if it said "Save Slot %i selected", instead of "Save Slot (%s) selected".  That way if it was just %i (or %d or %u) then the decimal wouldn't need a second multi-language param since it would be the same in all languages.

project64:
more in languages like c# you o $1 for the first param $2 for the second. This is for the string to be translated where sometimes the sentence structure is reversed

cxd4:
I would just let the language file author move the `%i` around to accompany sentence structure.

project64:
closing old issue, if this is still an issue it can be reopened

PR #755, opened 2015-11-14
Fixed "Save Slot (%s) selected" Text

samuelyuan:
The text was not showing up correctly because it was not using wide
strings. This worked in an earlier build, but was later changed and stopped displaying correctly. 

**Before**
before fix

**After**
after fix

On a side note, the text can now be translated easily to other languages.


samuelyuan:
This should fix Issue #683

Source/Project64/User Interface/Main Menu Class.cpp
```diff
@@ -531,8 +531,8 @@ bool CMainMenu::ProcessMessage(HWND hWnd, DWORD /*FromAccelerator*/, DWORD MenuI
     case ID_CURRENT_SAVE_7:
     case ID_CURRENT_SAVE_8:
     case ID_CURRENT_SAVE_9:
-    case ID_CURRENT_SAVE_10:
-        g_Notify->DisplayMessage(3, stdstr_f("Save Slot (%s) selected", GetSaveSlotString((MenuID - ID_CURRENT_SAVE_1) + 1)).ToUTF16().c_str());
+	case ID_CURRENT_SAVE_10:
+		g_Notify->DisplayMessageWide(3, GS(MENU_SLOT_SAVE), GetSaveSlotString((MenuID - ID_CURRENT_SAVE_1) + 1).c_str());
```
project64: actually it should just be Save Slot (%ws) selected, not $s

toehead2001:
You've put in tab indentions back in after zilmar replaced them with spaces. :smirk: I recommend using spaces so the files are consistent.

cxd4:
Unfortunately an easy mistake to make, as Visual Studio and Notepad++ both default to tabs.

samuelyuan:
I made some changes to the files.

Source/Project64/User Interface/Notification Class.h
```diff
@@ -35,6 +35,8 @@ class CNotification :
 	//User Feedback
 	virtual void DisplayMessage(int DisplayTime, const wchar_t * Message) const;
 	virtual void DisplayMessage(int DisplayTime, LanguageStringID StringID) const;
+	virtual void DisplayMessageWide(int DisplayTime, const wchar_t * Message, ...) const;
```
project64: I has to removed the ... version before for an upcoming change so I can not add it back now

Lang/English.pj.Lang
```diff
@@ -88,6 +88,7 @@
 #198#  "Slot 8"
 #199#  "Slot 9"
 #200#  "Slot 10"
+#201#  "Save Slot (%ws) selected"
```
project64: I know this works .. but I really hate putting %ws in to a language file

Source/Project64/User Interface/Notification Class.h
samuelyuan: Yes, I saw that you removed it and I suppose you have a good reason for that. The only problem is that wide strings aren't displayed properly anymore. I tried this using an earlier version and didn't have this problem.

Is there a better way to solve this than just add a duplicate function?

project64: the problem with the current code uses stdstr_f to format, which is converting from utf16 to utf8/ansi and converting back which does not seem to work to well. We probaly need a stdwstr_f to be able to format strings in pure wchar_t

samuelyuan: I think creating a stdwstr_f class solves the problem.

Lang/English.pj.Lang
samuelyuan: It doesn't look good, but I didn't want to hard code the %ws part, since the word order can be switched in different languages.

PR #769, opened 2015-11-19
Upgrade wxWidgets to version 3.0.2

JunielKatarn:
Two main reasons for this change:
1. Version 2.8 is rather old, and the wxWidgets forums have mentions of many improvements in newer versions (since 2.9).
2. Some dependencies used by the 2.8 codebase made it impossible to build Project64 with a non MSVC toolset with xp compatibility (e.g. vs14_xp would work, but vs14 would fail). This change fixes this, allowing compilation in other compilers such as LLVM 3.8 and Intel Compiler Collection (verified myself). This will help the transition to other compilers easier.

Tested:
* Built all flavors (Win32, x64, Debug, Release) and combinations.
* Tested the Glide64 plugin configuration GUI (which uses wxWidgets) successfully.
* Ran Ocarina of Time, World Cup 98 and Lem's crap test ROM without issues.

PR #774, opened 2015-11-22
reverting #769

cxd4:
Because of the changes in question, VS2008, VC2010 (which I, myself, ran), and VC2013 all fail to compile with almost a hundred errors due to breaking conflicts between these hacky WX header adjustments and the system headers reserved by the compiler platform.

I will not bother attempting to debug a conflict of this scale.

Sabotaging older VC support without permission is also not appreciated.

If you can't submit huge, giant pull requests without either the experience necessary or testing on older compilers to validate their portability (as newer compilers don't throw as many strict language requirements as old ones do), then do not bother submitting them.  Divide your pull requests.

I apologize for not posting a comment on all of this in the first place to try to prevent this outcome.


JunielKatarn:
Kind of weird to be affected (allegedly) by someone else's commit, instead of being the build breaker isn't, it? XD

I'll validate this conjecture is correct and if so, resubmit a corrected version that does not disrupt older VC versions.

Thanks for testing.

cxd4:
And when I say, "divide" pull requests, I mean don't attempt to accomplish multiple things in one PR.

i.e., don't combine any two or more commits of unrelated accomplishments (as I have been observing these past few weeks seems to be a common habit in these PRs).  So make a) smaller changes, b) fewer changes (only what is required), c) more of them and more often in simultaneity to a) and b).

Frank-74:
Does https://github.com/project64/project64/commit/fa60e55f57134f689f6038882f811191f17d9bc7 need to be reverted as well then?  As it fixed all but Glide64.

Project64 settings and the UI in general works faster with the updated wx.  Before opening settings for the first time took a few seconds to appear.

So I don't doubt the upgrade is a good idea.  It probably fixes the wxwidgets errors as well.

cxd4:
I think so.  Wasn't fa60e55 just to try fixing all the build errors after Juniel's PR?

So I chose to revert project64's commit too, as it was preventing me from doing an automatic "Revert" through the GitHub website UI.  I think he only did that commit because of the PR that broke it.

>Project64 settings and the UI in general works faster with the updated wx. Before, opening settings for the first time took a few seconds to appear. Now its instant.

Project64 GUI is currently being redone/refactored for separation.

So the best thing that could happen is the dependency on wxWidgets could be removed altogether, rather than having to be constantly maintained/increase the size of the repository.

>Kind of weird to be affected (allegedly) 

lol, _"allgedly"_?

3 people have shown that their builds with non-VS2015 are broken (myself, Frank74 and project64's commit) and you emphasize that I am _allegedly_ affected?

That's a curious thought.  Changes are usually best done based on the concept of what is being changed, even if it does break the build temporarily changes can still be a good thing.

In this case the build is broken in exchange for a number of mixed-purpose commits that do not highlight any conceptual understanding of the changes being made (or even where to begin debugging what you assumed wrong), or else the problem will be fixed and this giant "revert" PR I am making will not have to be merged and redundantly bubble the repository history.  So if you can succeed at that, great and thanks.

Frank-74:
```
Project64 GUI is currently being redone/refactored for separation.

So the best thing that could happen is the dependency on wxWidgets could be removed altogether, rather than having to be constantly maintained/increase the size of the repository.
```
Wooohooo!  I've always hated wxWidgets with a vengeance.  Its unnecessary.

cxd4:
Yeah, to have an Android GUI the whole UI stuff is being changed to divide exactly how dependng on wxWidgets Project64 really needs to be.

I know there will be a command-line implementation (no GUI/no WxWidgets) at the very least because I will be needing that to finish a basic Linux port and transition further towards an Android one.

cxd4:
But, putting that aside even.

Really I have no issue with updating Wx to newer versions.  I'm not even in a hurry to need to compile again the Project64 core again anytime soon, but in case anyone else needs to it was just way easiest for me to restore the repository to its earlier state.

My only basic point here is that if there were multiple, small PRs, I would just be able to debug those and find out where vc2010 and vc2013 broke.  Since I cannot test vc2008 and I cannot tell which of the changes within any given commit made the assumption(s) that broke all 3 I really do not value trying to find the time.  Again, if Juniel does, even better because if this PR gets merged it's going to spam my statistics as well as the repo history.

cxd4:
...Or it can be merged and spam my stats anyway.

But hey, look at the bright side.  You can re-organize your pull request from the master branch and divide the commits for cleanly fun this time. :)

project64:
> Wooohooo! I've always hated wxWidgets with a vengeance. Its unnecessary. 
wxWidgets  is just being used for glide64 .. maybe I should refactor the UI in that to be more basic but that is not currently in my plans.

JunielKatarn:
@project64 That means the only dependent component on wxWidgets is Glide64's settings GUI, right?
If so, why don't we migrate such GUI to native Win32 APIs and get rid of the dependency?
I made a test build, and it cuts over 2 minutes build time.

JunielKatarn:
> 3 people have shown that their builds with non-VS2015 are broken (myself, Frank74 and project64's commit) and you emphasize that we are allegedly affected? That's a curious thought.

Not curious at all. As usual regarding build break topics, and to no one's surprise, you are wrong. My change did not break any compilation using VCXPROJ (VS2010+) makefiles. A thousand people could have gotten VS2013 errors, and none of them would have been caused directly by #769.

Here is my research on this issue:
* My change on zilmar's master branch (b5a4dc4eefd964629b0e65f574994b7a3ec51cfa) you rushed to revert (for it didn't cause the 2013 breaks) builds and runs correctly under the following conditions:
  * Using Visual Studio 2010, 2012, 2013 and 2015.
  * Using Visual Studio 2008 compilers, using  MSBuild or Visual Studio 2010+ as the build tool.
  * All configurations (Debug Release)

The problem was: Visual Studio 2008 as a build tool uses VCPROJ makefiles, which I overlooked and did not update to use the wx 3.0 codebase.

* @project64 submits fa60e55f57134f689f6038882f811191f17d9bc7 to fix building under Visual Studio 2008. This may have worked for that version, but broke VCXPROJ versions (VS2010+). That's why some of you got build breaks using version 2013. I checked out that specific commit and, logically, it also broke 2015.

* @cxd4 Submitted #774, which reverts wxWidgets to version 2.8, and compilation got back to normal.

I can re-submit a PR that now includes the necessary changes to VCPROJ (2008) makefiles, which of course I will test first. But at this point, I'm wondering why do we need to keep supporting 2008 (as an IDE and build tool. MSBuild/VCXPROJ can use the VS2008 compiler toolset seamlessly).

Is there a specific technical or logistical need to keep using this old version?
Keeping it creates the need to keep both VCXPROJ and VCPROJ makefiles in sync, which is a duplicate effort. I'll follow-up this topic in a separate PR.

cxd4:
>Here is my research on this issue:

My research on this issue was that it never compiled simultaneously on VC2010 and VC2008.

In response, you're blaming zilmar's commit for your own hacky Wx adjustments.
Not much more to be said past that.

>I can re-submit a PR that now includes the necessary changes to VCPROJ (2008)

No large pull requests.

Start small, learn from your past mistakes, and make individual pull requests.

Those who express no effort at all to allocate and plan their steps are not likely to be trusted.

>The problem was: Visual Studio 2008 as a build tool uses VCPROJ makefiles,

As usual regarding build break topics, you are wrong.

Did you not read the part where Frank74 and project64 **still** could not get it compiling on VS2008?
Did you not read the part where most of Frank's compile errors _were not even related to missing include errors caused by the VCPROJ vs. VCXPROJ build systems_?

Being forced to make a pull request I did not want to have merged does not equate "rushing" to PR.

Though, since you raise the topic, making revert pull requests for a simple C vs. C++ compiler linkage one-line difference (and falsely labeling it a "compile" error), however...if you want to talk about rushing to make revert commits :P then yes. If you're intending to make yourself look as jealous as possible, then yes, you were at least right about that much.

JunielKatarn:
That's a lot of words. Wow.

I tested the changes. Nothing breaks in VCXPROJ (all versions) builds using #769.
Later commits make it break.
That's a fact, not my opinion. I detailed the issues in my last post. If you have first hand evidence of the contrary, please provide the scenario and I'll happily fix it and learn my lesson.

While I agree with the "smaller PR" paradigm, I don't think I can make this kind of change in smaller PRs while keeping the builds sane. In any case, my PR was not large enough (it missed the 2008 changes).

How would you go about upgrading the wx code base in "smaller chunks"? I tried a few files a t a time, but got horrendous build errors.

The biggest size factor of #769 is the actual wx code base files changed to 3.0.2. Other changes in the PR are quite small.

Unless anyone has a better idea, I think the only way for a clean upgrade is to restore and improve the change, and thoroughly test all versions.

JunielKatarn:
> Those who express no effort at all to allocate and plan their steps are not likely to be trusted.

Neither should regular build breakers who consider running a compiler "a waste of time".
Spare me the lecture. This is the first time a change of mine breaks something, and I have proven it was only VS2008.

I don't care about ego wars. Can we please focus on the solution?
How would you make a modular (multi-PR) upgrade of the wx code base, when the header and source files change significantly from 2.8 to 3.0?

Again, it's only a "large PR" because of the actual wx source files.

cxd4:
umm...lot of words, lulz? ^^

>Neither should regular build breakers who consider running a compiler "a waste of time".

I don't recommend getting scarred over a grand total of 2 times I broke a compile.

>This is the first time a change of mine breaks something,

I'm sure if this was true, you wouldn't feel the need to always test compile with 2015.

>I have proven it was only VS2008.

What you have "proven" is VC2008(-) and VC2010+ use different build systems.
We are familiar with that...that's not a new concept.

You talk about VCPROJ vs. VCXPROJ like it has anything to do with the vast majority of code errors when it really has little to do with it.  Again, details in my last post.  Try actually learning about code instead of relying on plug-and-play assumptions from what your compiler screens tell you all the time.

>While I agree with the "smaller PR" paradigm, I don't think I can make this kind of change in smaller PRs while keeping the builds sane. In any case, my PR was not large enough (it missed the 2008 changes).
How would you go about upgrading the wx code base in "smaller chunks"? I tried a few files a t a time, but got horrendous build errors.

It can be done.  Just avoid simultaneously submitting things like:
https://github.com/JunielKatarn/project64/commit/8534ac665efc47f85ab7e6f8a15353cf6f10f5ae
https://github.com/JunielKatarn/project64/commit/a1536ee6e4cc77bdafe8a967c5559575ed9945ec

Instead of a "few files at a time", it should be possible to just add new files (no matter how many/large) and/or disabled code that wasn't already present, without yet attaching them to either the VCPROJ or the VCXPROJ.  Then, in future pull requests, work on compatibly attaching them.

JunielKatarn:
> Just avoid simultaneously submitting things like:
JunielKatarn@8534ac6
JunielKatarn@a1536ee

That has nothing to do with size modularity.
The second commit you mention didn't make it to the final PR. It was a temporary experimental change. I did not submit both.

I tried your suggestion before, actually. Just added the newer files. Problem is: replacing older versions without changing make files made the build fail. That approach won't help.

I'm not scarred about your build breaks. I'm just surprised about your demand for quality standards you certainly don't hold up to in some of your PRs.

Now, I don't know why I need to keep clarifying this:
My change only broke 2008 (OK, and earlier versions) because I didn't also update those makefiles. I built all later VS versions and everything went fine.
I'm not "blaming" anyone, but later changes fixes 2008 but made 2010+ incompatible.

I don't see a way to make partial source upgrades (unless I mimic commits from the wx repository, which is too much to ask just to comply with your demand for small changes).

What I can do, is do more extensive testing making sure no VS build tool versions break.

cxd4:
I'm not singularly talking about size modularity.
(And yes you're right I did not confirm if you added a new commit that canceled your old commit...it's still hard to read because commit-by-commit is the cleanest way to try reading huge pull requests.)

I'll try to highlight my point a bit better.  It's not just about "is it 900 line changes + 900 line changes"; it's also about "is it 900 line changes + 1 unrelated line change".  That 1 unrelated line change impacts the readability, _sometimes_, even more than the other 900 lines do, if it's intended to accomplish something in a different dimension.

>I tried your suggestion before, actually. Just added the newer files. Problem is: replacing older versions without changing make files made the build fail. That approach won't help.

Not what I meant.

To put it another way, making huge pull requests is easily fine to accept (given a purpose) if all they do is add say 300, huge files.  If that's ALL they do, nothing can break.  They're not attached to the build, right?  They're not reserved file names for existing #include's, right?  So how can anything break?  Size is a factor, but not everything.

What I'm saying is, some of your Wx additions featured new files that weren't attached to the build already.  Potentially you could add those files to the repository, not attach them to the project yet, and there is no way the pull request at that stage could break any compiles.

>I'm not "blaming" anyone, but later changes fixes 2008 but made 2010+ incompatible.

I would like to politely re-state that 2008 was never fixed.  We tried, but hundreds of errors remained even after zilmar's commit.  I would also like to politely restate that vcproj vs. vcxproj was a coincidental factor in the equation and not the whole story.

Yes I lecture more philosophies about "standards" than most people do.
No I do not tell people to always test compile--just to post readable changes, even if it breaks.

JunielKatarn:
> If that's ALL they do, nothing can break. They're not attached to the build, right?

The problem is, it's not only new, non-registered files. It's also older files and headers that change (in some cases, significantly).

The changes on those existing files after upgrading the code base (meaning, H, CPP and RC files within the wx folder structure) break compilation if the build files are not also upgraded. I tested this.

> Potentially you could add those files to the repository, not attach them to the project yet, and there is no way the pull request at that stage could break any compiles.

Based on what I just wrote, no. It can and does break because of updated files.

The only alternative I can think of, is to manually track attached files and avoid updating those, adding unattached new files and then updating attached files and build files in a third PR. It's too much work in my opinion, and I don't consider it necessary, because I already found the root cause (non-updated VCPROJ files) and I can test all relevant configurations.

cxd4:
Up to a point then right, one can no longer divide large pull requests when it begins to replace attached header files.  All I'm saying is it can be partially done--likely never completed though.

This fundamental problem of large PRs I think is not really something that should be your fault anyway.
* The header files traditionally should be kept to compiler includes, not statically uploaded to the Project64 repository.  On Windows this is convenient because people don't generally have as easy of a time installing the Wx dependencies / building out of the box, but on saner platforms the idea of statically including WxWidgets in the PJ64 repo source is of less pragmatic benefit I believe.
* As you said it would be even better to remove the Wx dependency from Glide64 altogether.  While I would like to assist with this, I also cannot risk replacing Wx with Win32-only code that won't compile on other systems.

PR #779, opened 2015-11-24
Fix default toolset when using VS 2010

JunielKatarn:
Within the VCXPROJ Visual Studio versions, only 2010 does not have an `_xp` flavor (it is XP-compatible by default).
It also des not set the `$(DefaultPlatformToolset)` variable.
This causes build failures when compiling from within VS2010 and requiring to set the toolset manually.

This change skips setting the `<PlatformToolset>` if the `$(DefaultPlatformToolset)` variable does not exist. Visual Studio 2010 falls back to its own toolset (`v100`), which is the expected behavior.


cxd4:
ya to test compile with 2010 yesterday I had to change it to "v100" or whatever

Looks good to me.

PR #780, opened 2015-11-24
Upgrade wxWidgets to version 3.0.2 [UPDATED for VS2008]

JunielKatarn:
This change brings back #769, fixing the build problems with Visual Studio 2008.

Besides the same runtime tests (OOT, Lem's Crap Test ROM), the following successful build were verified (using the Visual Studio IDE as a build tool):
* 2008
* 2010
* 2012
* 2013
* 2015

Configurations:
* Debug
* Release

In short, this change now bulds without issues in all visual studio versions.
Note. Because it upgrades the wxWidgets codebase from 2.8 to 3.0, this change is rather big.
But the bulk of the differences lie in the wx-specific CPP, H, and RC files.

The relevant part of the change (which is easier to analyze) is the build files (wx/core,base) both VCPROJ and VCXPROJ.

@Frank-74, @cxd4, @tony971, @project64: Please take a look if you like.


JunielKatarn:
Also note: this change is NOT related to #781.

cxd4:
I would tend to assume it does build on 2010+ (again, w/o using my time to test :P I am not interested in proving anybody wrong here) but am only curious about 2008, which I cannot test.

@project64 I would like to use this as yet another opportunity of example of why at least separate repositories not all-in-one single repository is a good idea and b) storing header files like Wx headers in the repository (which, preferably, should not be modified in the first place) is not preferable to having users install them to their compiler system include directories.

But provided you continue to prefer it the way as it stands, then as far as I'm concerned this should be merged provided someone with vs2008 access can first test it.

tony971:
I'd really appreciate pj64 remaining a clone-and-compile solution.

cxd4:
Which is independent of what I suggest.

You're not supposed to modify, include, or hack APIs into source repositories to begin with.  Tell me if Dolphin doesn't apply the same tradition to it. :)

With a little imagination you can see that a one-time installation of system headers to your compiler directory does not preclude Project64 from remaining a "clone-and-compile" solution.

cxd4:
By the way, I cannot read this commit.
https://github.com/JunielKatarn/project64/commit/c4091c40b47c09bb555a1cfcd15dc493f611d97d

So I have no way of knowing whether the changes there are all optional, all required, all good or all bad.  Judging by the fact that over 50% of both vcproj's were updated, I'd venture to speculate that very many of those changes did not need to all go in one commit.

PropertySheets/Platform.props
```diff
@@ -25,7 +25,7 @@
 
   <ItemDefinitionGroup>
     <ClCompile>
-      <PreprocessorDefinitions>WIN32;_WIN32;_WINDOWS;WIN32_LEAN_AND_MEAN;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>WIN32;_WIN32;_WINDOWS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
project64: I am not removing WIN32_LEAN_AND_MEAN, I know this is where a lot of the complications comes from having it, it is where I I had to tweak the source so that it included other headers because they now were no longer included.

Frank-74:
I built from your wx_3.0 tree.
```
========== Build: 13 succeeded, 4 failed, 1 up-to-date, 1 skipped ==========
```
Full build log here :> https://dl.dropboxusercontent.com/u/9498358/n64/wx_3.0_build%20log%20VS2008.txt

PropertySheets/Platform.props
JunielKatarn: Version 3.0.2 needs it or won't compile.
That was not the cause of the errors.

This change builds out of the box (no tweaks) in all VS versions.

JunielKatarn:
Most weird. I don't get that error.
Can you make sure you `git pull` the latest changes?

JunielKatarn:
Possible explanation: this PR has the same branch name as #769, which had VS2008 issues.

JunielKatarn:
Possible explanation: this PR has the same branch name as #769, which had VS2008 issues.

Frank-74:
I download the zip and build solution everytime.  I don't use Git app.

And I have to delete my previous builds first, as I'm running out of disk space.

So which branch of yours is the correct one?  The wx_3.0 branch is the only one with this PR in, thought it was the right one.  Fix a branch with all the updates, and I'll delete, download, and try again.

Your previous PR after project64 fixed it, built ok for me.  Well Glide64 didn't build, but that hadn't een fixed.  Only [Project64] was fixed.

Frank-74:
If I try to pull into an up to date branch from wx_3.0 (The branch with this PR), it has these commits as well as this one.
Revert "reverting #769"
Revert "[Project64] Trying to get wx to compile"

cxd4:
He's trying to revert the revert because he thinks he can fix 2008 just by touching the VCPROJ.

LegendOfDragoon:
I really do not like reverting. Just seems messy imo.

cxd4:
Can totally agree.  I genuinely was not happy making the revert commit that I had to make.

JunielKatarn:
> So you've reverted the revert?

Yes, I reverted the revert. I've done it many times in several projects, and it's harmless. It's also the cleaner, less time-consuming way to do it.

@Frank-74 I'll try downloading the zip and see if I get the same behavior as you do.

JunielKatarn:
> By the way, I cannot read this commit.
JunielKatarn@c4091c4

@cxd4 I guess, because it has a lot of changes in the VCPROJ files, GitHub won't render them. You would need to pull my branch and diff those files against master. I'm going to take a wild guess that you are not going to do that.

Essentially, that commit replaces the <Files> section from master (wx 2.8) with the <Files> section of wx 3.0.

JunielKatarn:
@Frank-74, I have confirmed the downloaded zip does fail to build. I'll look into this...

JunielKatarn:
Closing for now. There are inconsistent build experiences with VS2008.

JunielKatarn:
@tony971 I agree, but for the time being, I just want to make this upgrade work.
In the future, I hope we can find a different solution to integrate wxWidgets without "forking" their source. Moreover, only Glide64 uses it. I'd love to see the dependency fully gone...

Issue #675, opened 2015-10-18
wxWidgets error.


Frank-74:
Using default plugins and settings, start F1 Racing Championship (E) (M5), and when the video starts (in wrong aspect ratio), open Settings.

![wxerror](https://cloud.githubusercontent.com/assets/11056843/10563290/52f68ee0-757c-11e5-9688-afbf1a460388.PNG)

Seems to only happen after Jabo writes a d3d syslog.

Frank-74:
Happens also after adding a cheat.  So its not just a Jabo problem.

JunielKatarn:
What exactly is wxWidgets used for?
Only GUI window elements?

Frank-74:
@JunielKatarn 
I think only the UI, zilmar will have to confirm for certain.

Its a recent regression.  Wasn't getting this error a few weeks ago.

I using Windows 8.1.

cxd4:
I always got this error just by having AziAudio.dll and Glide64 final both installed to the Plugin subfolder, regardless of which plugins were currently selected.  But I can't remember what I did to fix it--probably just deleted Glide.

Frank-74:
#783 fixes these errors.

oddMLan:
Fixed with the removal of wxWidgets 68a8ed03696f3365430d8cca14d5084100165cd0
please close

PR #783, opened 2015-11-25
Upgrade wxWidgets to version 3.0.2 [THIRD IS THE CHARM]

JunielKatarn:
Found the issue with previous attempts. Somehow, I managed to leave the `WIN32_MEAN_AND_LEAN` untouched in VS2008 props files submitted to this branch. My local changes did remove it, and that's why only I got successful builds before.
3579400a41af58d3d671261eb5a3a2cd9960d24f does the trick.
Please note, leaving it out does not affect the rest of the projects.

@project64: On the previous PR #780, you didn't want to touch that flag. While this might be valid for the previous wxWidgets code base (2.8), it breaks the upgraded one (3.0).
Moreover, if you look at wxWidgets 3.0.2 and their VS makefiles (all versions), none of them set the flag. Removing it is the final step to get this change to work for good.

Steps to test (I took them myself, of course)
1. Checkout or download this branch (wx_3.0)
2. Open whichever version of Visual Studio suits you (preferably 2008, which is the controversial one).
3. Build from within the IDE (F7, SHIFT+CTRL+B, RIGHT-CLICK+Build, or whatever method...)

I buit and ran VS2008, 2010, 2013 and 2015. The Glide64 GUI works correctly.

@Frank-74, this time it will build fine in VS2008, I swear :angel:


JunielKatarn:
@Frank-74, if you give it a try, maybe this will help making #675 go away.

Frank-74:
```
========== Build: 17 succeeded, 0 failed, 1 up-to-date, 1 skipped ==========
```
Success!

project64:
> you didn't want to touch that flag. While this might be valid for the previous wxWidgets code base (2.8) , it breaks the upgraded one (3.0).

it is not valid for 2.8 code base either, I modified the code and spent a decent amount of time getting it working with the flag, the problem is that change effects every other project, which is why I do not want the property sheet modified.

Frank-74:
@project64 
Doesn't appear to affect any other project here after building this.  Haven't noticed anything different, file sizes are even the same.

project64:
https://support.microsoft.com/en-us/kb/166474

To speed the build process, Visual C++ and the Windows Headers provide the following new define WIN32_LEAN_AND_MEAN You can use them to reduce the size of the Win32 header files. 

So it should not necessarily effect the the file size, more the build time.

JunielKatarn:
> https://support.microsoft.com/en-us/kb/166474
To speed the build process ...

@project64:
I'm familiar with that article. It is mainly used in older versions of Visual C++ (up to 6.0, which are older than VS2005).

It renders no benefits in newer versions of Visual Studio, including 2008.
However, due to changes in wx 3.0.2, removing it will cause compilation to fail.

Again, both @Frank-74 and I tested that no components are affected, and I am able to run games and configure Glide64.

JunielKatarn:
@project64:
According to @Frank-74, this PR fixes issue #675.
As shown before, `WIN32_MEAN_AND_LEAN` should not be a big deal (it certainly doesn't break anything if removed from the props files).

tony971:
If you resolve the merge conflicts, I'd like to try this out.

Frank-74:
I timed the build solution time for both wx2.8 and wx3.0.  wx2.8 takes 6 minutes to build the whole solution, and wx3.0 just over 10 minutes.  So it is quite a big difference without the WIN32-MEAN_AND_LEAN.

cxd4:
I'd be more concerned with how to remove the wxWidgets dependency, tony.

The core project is still undergoing fundamental structural changes where things are already apt to easily break.  We can't afford to risk any competing fundamental changes like Wx header motifications, removing global cross-project macro defines from the property sheets, etc..  The most important priority is cleaning the code first.

JunielKatarn:
> So it is quite a big difference without the WIN32_MEAN_AND_LEAN.

I had a different experience. Mi compilation time increased less than one minute, but I get the point.
4 minutes is too much.

> I'd be more concerned with how to remove the wxWidgets dependency,

Me too! But I think you described the problem before:
* Glide64 needs a configuration GUI (I mean, maybe not "NEEDS", but it would be inconvenient to just throw it away and tell people to hack the CFG files).
* The config GUI could be ported to Win32 APIs, thus making wxWidgets unnecessary, but then porting to other platforms would be even harder.

I see two possible courses of action:
A. Try to install wxWidgets separately (I heard this was a hassle before on Windows, but maybe it's not the case anymore).
B. Separate Glide64 into 2 projects: Glide64 (core, no GUI) and Glide64GUI. That way we could detach the concern for a GUI framework and have a running plugin in less compilation time.

For now, I'll hack the makefiles around to re-introduce LEAN_AND_MEAN at a global level, but drop it in the wx makefiles (this would also make this PR compliant with @project64's requests).

project64:
I personally do not care about the version of Glide64 being used on another OS .. it is easy enough that the UI is changed to windows API and #ifdefed out for other OS's and they have to either implement a new UI or you modify the config.

cxd4:
That's the way my angrylion fork does it.
```c
#ifdef _WIN32
/* low-level Windows-framework specific GUI code here */
#else 
/* SDL windowing code for a UI, but did not care to add GUI because you can edit the conf[] */
#endif
```
So works for me.

Frank-74:
I could replicate the Glide GUI to be built in MinGW without wx easy enough.  Just use a resource editor to make a new gui.  Most resource editors can export the c++ code too.  Its then just a matter of attaching it to Glide.  Its not difficult to make a gui, no programming knowledge is needed at all.  Problem is attaching it to the code.

JunielKatarn:
OK, I managed to restore WIN32_LEAN_AND_MEAN in the default property files.
@tony971, @Frank-74 take a look if you like. It builds and runs successfully at least in VS2008 and VS2015.

@project64 Now that the Lean&Mean stuff is restored, woud you reconsider merging?

project64:
@JunielKatarn  yes willing to I will see if I get time on the weekend to do some testing with it and just verify things are working, then I will hopefully merge.

Frank-74:
@JunielKatarn Just timed the build.  It took exactly 8 minutes 20 seconds, compared to 6 minutes with 2.8.  It was over 10 minutes before.  I suppose the extra 2 minutes is down to undef WIN32_LEAN_AND_MEAN for just wx3.0.

Its working though.

JunielKatarn:
@Frank-74 These are the build times I got:
* master@https://github.com/project64/project64/commit/89886420d14d4587eda1fa4bbe61a0d9c006a50c: `Time Elapsed 00:01:52.32`
* https://github.com/JunielKatarn/project64/commit/653007a02a84702ba15007b60171b70b85047bee: `Time Elapsed 00:02:13.33`

They did increase for me, but not too significantly.

oddMLan:
wait, 6 minutes to build the current master? lol. it usually takes like 3 minutes in my computer.

JunielKatarn:
> wait, 6 minutes to build the current master?

I hear you. It takes rightly 2 in mine, so removing the lean&mean thing globally posed no issue in my case.
This hardware and software variety in the community ends up being very useful for testing all aspects of this project, including build time.

JunielKatarn:
Please note latest commit https://github.com/JunielKatarn/project64/commit/bf4d928c1f044e2d3f7c2b94d8c227b404f770ac . It's huge, because it's a merge with this change: e3b32c572db1ebe7f5c8cf60e4db148eebd7f83f.

Builds and runs using VCPROJ and VCXPROJ MSVC versions.

project64:
OK finally removed wx from the project completely so no need to upgrade it now

cxd4:
That's very useful news!  When I get into a more available space I'll check out the commits in detail and look at updating.

JunielKatarn:
That's great to know. The less dependencies, the better.
Too bad I spent a lot of hours on this change, but I prefer WX gone altogether.

project64:
I also spent a lot of hours on redoing all the UI in wtl so wx is not needed

JunielKatarn:
BTW, just tested this change in build [425](http://hyvart.com/download/project64/ci/0.0.0.425/). The Glide64 config UI works great.

cxd4:
yep definitely takes a few hours to remove the integration of some of those dependencies...not usually as big as this one

PR #784, opened 2015-11-26
fixed random crashing if End then Start emulation

### LINKED ISSUE #776: End Emulation > Start any game causes crash
I've just downloaded the new Alpha v2.2-971 and tested.

I extracted and ran Project64, selected ROM directory, turned off hide advanced settings, started GoldenEye.  On first run its ok.

Then End Emulation, restart GoldenEye, Project64 crashes during Rare logo.  Sometimes no crash error, just closes.

Does the same in my previous builds, but I never noticed it before.

Also, not sure if its related, but another problem is restarting a game after an error/end emulation.  It restarts without the render window.  Going full screen works, but exiting full screen shows the ROM browser again.

cxd4:
Fixes #776.

Hard to discuss the change...it's mostly self-explanatory but whether it was an accidental "mistake" on zilmar's part or if it was done on purpose and/or wrong is kind of open to interpretation.  Nonetheless it does fix the issue so I am proposing it.

Before this commit, the crashes all came from NTDLL.DLL so I fear that this "fix" may just solve hiding an underlying bug even beneath this one.  But it gets PJ64 back to the way it was originally before the commit so I think it's stable in that sense.

### DIFF
--- a/Source/Project64/N64 System/Emulation Thread.cpp
+++ b/Source/Project64/N64 System/Emulation Thread.cpp
@@ -36,7 +36,7 @@ void CN64System::StartEmulationThread(ThreadInfo * Info)
 
     CoInitialize(NULL);
 
-    EmulationStarting(Info->ThreadHandle, Info->ThreadID);
+    EmulationStarting(*Info->ThreadHandle, Info->ThreadID);
     delete ((HANDLE  *)Info->ThreadHandle);
     delete Info;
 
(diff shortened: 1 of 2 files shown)


cxd4:
Alright, now I did compile this on Windows first to observe the fix, but I have no Internet when I boot into Windows, only when I boot into Linux.  So I have to push this commit from Linux today.

Therefore I guess it can't hurt to ask @Frank-74 to confirm first if it fixes his issue.

Frank-74:
Yes, this fixes the crashing.

I'll close #776 when this is merged.  And open another issue about the File > End Emulation > File > Start Emulation bug.  ROM browser stays on top.  Double clicking in ROM browser, and File > Start Emulation take different paths.  Also happens after any error and then starting another ROM.

Thanks.

Issue #731, opened 2015-11-08
Character corruption in Rom Browser


Nekokabu:
File name in Japanese are not correct
![pj64](https://cloud.githubusercontent.com/assets/11252775/11020347/5709d98e-8660-11e5-9f2e-94251335e1f2.jpg)

![pj64_2](https://cloud.githubusercontent.com/assets/11252775/11020373/5b85d1ec-8661-11e5-8fc3-bbdb1c3e21a6.jpg)

Frank-74:
ROM name (Internal Name) is not using UTF.
![romnameerror](https://cloud.githubusercontent.com/assets/11056843/11316925/363dbbba-9012-11e5-855b-3049255c273e.PNG)
![saveerror3](https://cloud.githubusercontent.com/assets/11056843/11316930/6771aeb2-9012-11e5-9336-0a185d50dccf.PNG)

Nekokabu:
I think ToUTF16 () is wrong.

Nekokabu:
change CP_UTF8 -> CP_THREAD_ACP in MultiByteToWideChar in ToUTF16.
ROM Browser is so good, but GUI is BAD.
![pj64_3](https://cloud.githubusercontent.com/assets/11252775/11443418/652b5452-9560-11e5-9088-e3ec3f9886b2.jpg)

MELERIX:
is this related ? https://github.com/project64/project64/issues/427

Frank-74:
@MELERIX Yes, probably.

Nekokabu:
if change Windows regional settings,  Japanese filenames are recognised.

oddMLan:
I think this still needs to be addressed. Roms with Japanese filenames are not recognized by the Rom Browser if Windows regional settings are not set to Japanese. And if the Rom internal name is using japanese characters, it shows up as a garbled mess.

PR #786, opened 2015-11-27
fix Character corruption in Rom Browser

Nekokabu:
fix Character corruption in Rom Browser
https://github.com/project64/project64/issues/731

### DIFF
--- a/Source/Common/std string.cpp
+++ b/Source/Common/std string.cpp
@@ -214,7 +214,7 @@ stdstr & stdstr::FromUTF16 ( const wchar_t * UTF16Source, bool * bSuccess )
 	return *this;
 }
 
-stdstr & stdstr::FromUTF16(DWORD CodePage, const wchar_t * UTF16Source, bool * bSuccess)
+stdstr & stdstr::FromUTF16(uint32_t CodePage, const wchar_t * UTF16Source, bool * bSuccess)
 {
 	bool bConverted = false;
 
@@ -277,7 +277,7 @@ std::wstring stdstr::ToUTF16 ( bool * bSuccess)
 	return res;
 }
 
-std::wstring stdstr::ToUTF16(DWORD CodePage, bool * bSuccess)
+std::wstring stdstr::ToUTF16(uint32_t CodePage, bool * bSuccess)
 {
 	bool bConverted = false;
 	std::wstring res;
--- a/Source/Common/std string.cpp
+++ b/Source/Common/std string.cpp
@@ -214,6 +214,41 @@ stdstr & stdstr::FromUTF16 ( const wchar_t * UTF16Source, bool * bSuccess )
 	return *this;
 }
 
+stdstr & stdstr::FromUTF16(DWORD CodePage, const wchar_t * UTF16Source, bool * bSuccess)
+{
+	bool bConverted = false;
+
+	if (UTF16Source == NULL)
+	{
+		*this = "";
+		bConverted = true;
+	}
+	else if (wcslen(UTF16Source) > 0)
+	{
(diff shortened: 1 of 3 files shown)


Frank-74:
I'm getting build errors after applying this fix.  I get 3 failed after Rebuild All.
For example:
```
17>Main.cpp
17>c:\Users\Mary\Documents\25-11-15\project64-wx_3.0\Source\Common/std string.h(36) : error C2061: syntax error : identifier 'DWORD'
17>c:\Users\Mary\Documents\25-11-15\project64-wx_3.0\Source\Common/std string.h(38) : error C2061: syntax error : identifier 'DWORD'
17>.\Main.cpp(1097) : error C2668: 'stdstr::ToUTF16' : ambiguous call to overloaded function
17>        c:\Users\Mary\Documents\25-11-15\project64-wx_3.0\Source\Common/std string.h(38): could be 'std::wstring stdstr::ToUTF16(void)'
17>        c:\Users\Mary\Documents\25-11-15\project64-wx_3.0\Source\Common/std string.h(37): or       'std::wstring stdstr::ToUTF16(bool *)'
17>        while trying to match the argument list '(void)'
17>.\Main.cpp(1097) : error C2228: left of '.c_str' must have class/struct/union
```
But I'm using wx3.0, because it gets rid of the initiate problems when using Azimer's 0.7 and having Glide Final in the plugins directory.

cxd4:
>error C2061: syntax error : identifier 'DWORD'

In this case it's supposed to be `uint32_t`, not `DWORD`, in the commit.

I have no comment on the C++ errors.

Frank-74:
: fatal error C1083: Cannot open include file: 'stdint.h': No such file or directory

cxd4:
Heh.  Yeah one thing I like about VC2008:  It doesn't compile extensions like stdint.h. =]

You have to include zilmar's `<Common/systypes.h>` or something, whatever it was called.  That's where he works out all the stdint.h types for compilers that don't support stdint.h (and in my not-so-humble opinion, have no obligation to).

Frank-74:
It's `<Common/stdtypes.h>`

Nekokabu:
OK?

Frank-74:
Seems to be ok, but why do I get these strange characters for (J) Internal Name?
internalerr1

project64:
I made some changes (https://github.com/project64/project64/commit/5b7020453e184ccc2af04b5cd5b0d9b1d95aa894) so that the rom browser is using W char version instead of A version so this might change if this fix is needed any more .. tho it might cause other problems

Frank-74:
@project64 Now all Internal Names are Chinese.

internalerr2

cxd4:
lol.  I love it when that happens (to me, when I used to mess with ansi vs. unicode)

project64:
@Frank-74  thanks add a fix (https://github.com/project64/project64/commit/2cc85dd2cfbdfd4bf6c5f8dbc941314a1c211608) should fix the internal names .. at least for ansi based ones

Frank-74:
@project64 Ok, now its back to how it was before.

internalerr3

project64:
@Frank-74 great, tho not sure about internal names for jap games, I am converting ascii to tuf16, maybe it should be a different code page or somthing

Frank-74:
@project64 Maybe https://en.wikipedia.org/wiki/Code_page_932 ?

Frank-74:
Erm...  Not sure how I managed it yet, but....

japnames

Frank-74:
```cpp
wcscpy(pRomInfo->InternalName, stdstr(InternalName).ToUTF16(932).c_str());

PR #790, opened 2015-11-29
Fix Japanese Internal Names

Frank-74:
japnames

### DIFF
--- a/Source/Common/std string.cpp
+++ b/Source/Common/std string.cpp
@@ -241,3 +241,31 @@ std::wstring stdstr::ToUTF16 ( bool * bSuccess)
 	}
 	return res;
 }
+
+std::wstring stdstr::ToUTF16(uint32_t CodePage, bool * bSuccess)
+{
+	bool bConverted = false;
+	std::wstring res;
+
+	DWORD nNeeded = MultiByteToWideChar(CodePage, 0, this->c_str(), (int)this->length(), NULL, 0);
+	if(nNeeded > 0)
+	{
+		wchar_t * buf = (wchar_t *)alloca((nNeeded + 1) * sizeof(wchar_t));
+		if( buf != NULL )
+		{
+			memset(buf, 0, (nNeeded + 1) * sizeof(wchar_t));
+
+			nNeeded = MultiByteToWideChar(CodePage, 0, this->c_str(), (int)this->length(), buf, nNeeded);
+			if (nNeeded)
+			{
+				res = buf;
+				bConverted = true;
+			}
+		}
+	}
+	if (bSuccess)
+	{
+		*bSuccess = bConverted;
+	}
+	return res;
+}
--- a/Source/Common/std string.h
+++ b/Source/Common/std string.h
@@ -3,6 +3,7 @@
 class stdstr;
 
 #include <stdarg.h>
+#include <Common/stdtypes.h>
 #include <vector>
 #include <string>
 #include <list>
@@ -34,6 +35,7 @@ public:
 
 	stdstr   & FromUTF16 ( const wchar_t * UTF16Source, bool * bSuccess = NULL);
 	std::wstring  ToUTF16 ( bool * bSuccess = NULL);
+        std::wstring  ToUTF16 ( uint32_t CodePage, bool * bSuccess = NULL);
 
 	void ArgFormat( const char * strFormat, va_list & args);
 
--- a/Source/Project64/User Interface/Rom Browser Class.cpp
+++ b/Source/Project64/User Interface/Rom Browser Class.cpp
@@ -474,7 +474,7 @@ bool CRomBrowser::FillRomInfo(ROM_INFO * pRomInfo)
                 InternalName[count + 1] ^= InternalName[count + 2];
             }
             InternalName[20] = '\0';
-            wcscpy(pRomInfo->InternalName, stdstr(InternalName).ToUTF16().c_str());
+            wcscpy(pRomInfo->InternalName, stdstr(InternalName).ToUTF16(932).c_str());
         }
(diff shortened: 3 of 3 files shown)


Source/Common/std string.cpp
```diff
@@ -241,3 +241,31 @@ std::wstring stdstr::ToUTF16 ( bool * bSuccess)
 	}
 	return res;
 }
+
+std::wstring stdstr::ToUTF16(uint32_t CodePage, bool * bSuccess)
```
project64: I like the change to use the code page to get the correct internal name, needs to also be applied to the rom information dialog.

But I dislike duplicating ToUTF16 can we not just add the CodePage param to the current version with a defualt of CP_UTF8?

Frank-74:
Ok, I'll apply it to the Rom Information.  And add the CodePage param to the existing version as suggested.  I just copy/pasted @Nekokabu 's method to get it working.

It's 9am here, been up all night, need sleep.  Will make the changes tomorrow.  Thanks.

Frank-74:
@project64 I cannot get it working.  If I add the CodePage param to the existing version, I get errors.
> : error C2660: 'stdstr::ToUTF16' : function does not take 0 arguments

And using W instead of A, has broken reading of non english filenames.  I think https://github.com/project64/project64/commit/5b7020453e184ccc2af04b5cd5b0d9b1d95aa894 needs to be reverted.

project64:
ok I made some changes to stdstr (https://github.com/project64/project64/commit/942e07569afb4416028405d141a97e387add52c5) to add codepages, and (https://github.com/project64/project64/commit/655a73e1face690986901a1f9f9dd279e3f308f7) to use code page 932

Frank-74:
@project64 Just built the current master.

Internal Name is ok, but all japanese filenames don't show up in the ROM browser at all.

I used google translate to rename a J game to Japanese language.
Air Boarder 64 (J) [!] translates to > エアボーダー64（J）[！]

Frank-74:
Also, changing just this String to use code page 932, fixes the Internal Names also.
https://github.com/project64/project64/blob/master/Source/Project64/User%20Interface/Rom%20Browser%20Class.cpp#L1243

So https://github.com/project64/project64/commit/655a73e1face690986901a1f9f9dd279e3f308f7 wouldn't be needed.

Probably fixes the filenames as well, if they would show up.

project64:
the:
AddRomToList(stdstr((std::string &)SearchPath).ToUTF16().c_str(), stdstr(lpLastRom).ToUTF16().c_str());

might just need a different code page in the name conversion .. or I need to do the searching for names in the W version

Frank-74:
From > https://support.microsoft.com/en-us/kb/108450

```
CP_ACP instructs the API to use the currently set default Windows ANSI codepage. CP_OEMCP instructs the API to use the currently set default OEM codepage.

If Win32 ANSI APIs are used to get filenames from a Windows NT system, use CP_ACP when converting the string. Windows NT retrieves the name from the physical device and translates the OEM name into Unicode. The Unicode name is translated into ANSI if an ANSI API is called, then it can be translated back into Unicode with MultiByteToWideChar().

project64:
ok does using CP_ACP or  CP_OEMCP allow the files to be listed ?

Frank-74:
> ok does using CP_ACP or CP_OEMCP allow the files to be listed ?

@project64 It doesn't allow the files to be listed here.  Tried also codepage 932.

I think @Nekokabu needs to test if it works for him.  It might need a Windows with Japanese as default.

Can't you just rename a file using google translate to test it?
https://translate.google.co.uk/?ie=UTF-8&hl=en&client=tw-ob#en/ja/

Frank-74:
I don't know if this is useful, but I found this comment in a post.

> . A filename selected (e.g. from a file
open dialog box) that include Chinese characters will fail if a
boost::filesystem::path object is constructed from the const char*
ANSI path. If we just pipe the const char* directly into std::ifstream
then it works ok.

project64:
I will do some testing tonight, I might have to keep the path as ansi and just translate for display or look at converting path class to work as W function ..

the first one might be the easiest to do....

project64:
I changed the file handling back to ansi so it should be back to what it was originally. I tried making a file with jap file name and the ansi file name returned is useless. So no idea how it would work originally. I might need to change file handling to wide char through the whole application

Frank-74:
Just had a little experiment using File > Open ROM.

Japanese named files return 'Failed to find filename in 7z file'.

Zipping up the same Japanese file, but name the 7z with English, and it opens ok.

project64:
I do not see how non ansi files ever worked, this is as close as I can get with the ROM browser I believe with out changing all file handling to use utf-16 which I am not inclined to do at this point of time

Frank-74:
@project64 
@Nekokabu found a way in his original PR, he showed it working for him in the last comment i this issue.  https://github.com/project64/project64/issues/731

cxd4:
Codepage 932 doesn't support any of my filenames written in Arabic gibberish.

Frank-74:
@cxd4 Do Arabic named files display in the ROM browser Filename column correctly?

cxd4:
Dunno, I was trolling.

I think English is the ANSI-standard language for flamewars and trolling on the interwebz, so people should learn English or they won't survive.  (Or at least some Latin-based language that allows their N64 file names to be normal ones.)

932 sounds as good to me as about anything else, although some guy I am sure is going to rename all their things to be written in gibberish sign-language characters that would probably just break something no matter how it's done anyway.

cxd4:
I do like seeing the ROM "internal name" fields in the correct language though in the browser, for those Japanese titles that use NOJ's font code page.

PR #792, opened 2015-12-01
Improve Vector Dest Analysis

LegendOfDragoon:


Source/RSP/Recompiler Analysis.c
```diff
@@ -675,9 +678,25 @@ BOOL WriteToVectorDest2 (DWORD DestReg, int PC, BOOL RecursiveCall) {
 			case RSP_LSC2_HV:
 			case RSP_LSC2_FV:
 			case RSP_LSC2_WV:
-			case RSP_LSC2_TV:
 				if (DestReg == RspOp.rt) { return TRUE; }
 				break;
+				
+			case RSP_LSC2_TV:
+				if (8 <= 32 - RspOp.rt) {
```
cxd4:
```c
if (8 <= 32 - rt) {
if (0 <= 24 - rt) {
if (rt <= 24) {
```
I think I have an easier time reading it when written in this form.  To me, at least, it seems to better show that 24 is the magic number for not having to wrap around the final GPR[31], or you could also say `if (rt <= 32 - 8)`.

PR #793, opened 2015-12-01
updated spanish translation

MELERIX:


Lang/Spanish.pj.Lang
```diff
@@ -419,7 +420,7 @@
 //Edit Cheat
 #1027#  "Editar Truco"
 #1028#  "Actualizar Truco"
-#1029#  "El Truco ha sido cambiado.\n\n¿Deseas actualizar?"
+#1029#  "El Truco ha sido cambiado.\n\n¿Quieres actualizar?"
```
JunielKatarn: @MELERIX We had discussed this topic before.
"Deseas" is more appropriate for an option offering in a software application cation menu.

Please take any professional application translated to Spanish, and the definition for "desear" in the RAE dictionary as reference.

"Desear" does not semantically translate as "to desire" in this context.

Lang/Spanish.pj.Lang
```diff
@@ -177,7 +178,7 @@
 #320#  "Seleccionar directorio de ROMs actual"
 
 //Messages
-#340#  "¿ROM Dañado? Usa GoodN64 y verifica que la RDB esté actualizada."
+#340#  "¿ROM Dañada? Usa GoodN64 y comprueba por una RDB actualizada."
```
JunielKatarn: "Comprobar por" is not an appropriate Spanish phrases verb.
You shouldn't try to make a literal translation from English.
This particular statement might be confusing for native Spanish speakers.

The previous version would translate to English as "... verify that the RDB is up to date", which is the right idea.

Please rephrase or revert this line.

Lang/Spanish.pj.Lang
MELERIX: the right one is "quieres".

btw, now I'm participating in google translator team and checked what is the right translation with others members, and the right translation for "do you want" is "quieres".

you can verify it now in google translator or in bing translator also.

Lang/Spanish.pj.Lang
JunielKatarn: The main issue with this change is "por".
You can "verify for" in English, or maybe "test for".
But you cannot "comprobar por" in Spanish.

Suggestion: use "comprueba que la RDB esté actualizada".

MELERIX: in english is "check" which in spanish is "comprueba", while "verify" is "verifica".

MELERIX: "comprueba que la RDB esté actualizada" can be valid, but for that we need to modify also the english translation to: "check that RDB is updated", otherwise it will not match the right translation.

Lang/Spanish.pj.Lang
JunielKatarn: Translators are useful, but hardly an academically valid resource.
"Quieres" may be understood, but it's a matter of quality writing.

Please do take a look at the resources I referred you to.
It's a terrible idea to use ONLY google translate. It translates chunks of text, but is no authority for "reglas de redacción".

Are you a native Spanish speaker, may I ask?

MELERIX: english and spanish, btw you can also verify the translation in bing.

Lang/Spanish.pj.Lang
JunielKatarn: It may not match word by word, but that is not important.
What is important is that the idea (up to date RDB check) is translated, and readable following best practices and common terms and style in the target language.

You can change this translation without needing to modify the English version.
Also, translated words can have very different contextual meanings.
If you are curious, compare "condescending" and "condescendiente". They are NOT perfectly equivalent.

MELERIX: > You can change this translation without needing to modify the English version.

yes, but that is not right for a proper translation, in a proper translation you have to preserve the original context of the original lang.

about "condescending" the right spanish translation is "condescendiendo", while "condescendent" is "condescendiente".

Lang/Spanish.pj.Lang
JunielKatarn: I'm talking about formal/business writing.
Can you name a piece of professional, even commercial software that offers options using "Quieres" instead of "deseas"?

I recall providing you an example of some Microsoft website, translated to Spanish, that uses the latter.

Again, please don't use translators for semantics or redaction accuracy. If those tools were an appropriate source, there would be no people translating books or movies.
Such tools are great for basic translations, but fall short for more complex or subtle texts.

MELERIX: > Can you name a piece of professional, even commercial software that offers options using "Quieres" instead of "deseas"?

in Windows 10 you can notices that now some words have been translated to "quieres" instead of "deaseas".

and btw I don't use translators, I translate everything myself, just mentioned that I participate in google translation team.

Lang/Spanish.pj.Lang
JunielKatarn: `con·de·scend·ing
ˌkändəˈsendiNG/
adjective
having or showing a feeling of patronizing superiority`
(From Google)
See what I mean?
The same word could refer to the continuous of "to condescend", but also to an adjective, which has a different meaning in Spanish.

> preserve the original context...

What exactly do you understand by context?
Preserving context does NOT mean translate word by word. The context is the conceptual "environment" surrounding the text. Again, conceptual, not literal.

If we enforce to keep the word sequence so tight between languages, translations to both Spanish and Italian, or Chinese would be close to impossible.

JunielKatarn: Let me ask you this.
What part of the context is lost if you use 'comprueba que la RDB esté actualizada'?

Is the -idea- communicated in Spanish different from the idea that would be communicated in English?

Lang/Spanish.pj.Lang
JunielKatarn: Understood.

Please take a look at this book fragment:
https://books.google.com/books?id=AGJwGzTJaS4C&pg=PA40&lpg=PA40&dq=deseas+o+quieres&source=bl&ots=1sTsZDBFb4&sig=CW_plCgutBmwOBw_GYhliJQR_So&hl=en&sa=X&ved=0ahUKEwjcvpTY1rrJAhWElIgKHYalDC0Q6AEIJjAF#v=onepage&q=deseas%20o%20quieres&f=false

Note that both "Quieres" and "deseas" are valid to express an offering.
However, "deseas" is used more often in formal/professional contexts (i.e. in a restaurant, when they ask whether you wish to order).

Lang/Spanish.pj.Lang
MELERIX: "to condescend" can be translated to: "a condescender" o "para condescender", depending of the context of the original phrase.

"comprueba que la RDB esté actualizada" is ok, just that sounds different than english in context of the phrase (personal opinion), anyway we can modify the english phrase to: "check that the RDB is up-to-date" and it will match perfectly I guess.

JunielKatarn: I wouldn't thane an issue with that. Thanks.

As a side note, "to condescend", "to be" , "to do" or any other verb is just a way to refer to its infinitive form. Therefore, it's the right way to translate "condescender", "ser", "hacer", etc.

Think about "I like to play". It translates to Spanish as "me gusta jugar", and NEVER as "me gusta a jugar".

Lang/Spanish.pj.Lang
```diff
@@ -186,7 +187,7 @@
 #400#  "Ajustes"
 
 //Tabs
-#401#  "Plugins"
+#401#  "Complementos"
```
JunielKatarn: Good catch. I hadn't thought of a good translation for 'plugin'.

Lang/Spanish.pj.Lang
MELERIX: "deseas" is more used for a choice related with a pleasure question, for example: a food or sex.

"quieres" is more used for a choice related with a common question, for example: an action, a step.

JunielKatarn: That is not correct.
"Desear" has three possible meanings, that can be translated to English:
* To wish
* To desire
* To want

I just gave you a sample text that shows that verb is NOT limited to "desires", as nouns. Please don't disregard it.

I've studied and applied formal Spanish in many countries through many years (also a native speaker, BTW) I'm not making up these concepts.

JunielKatarn: Oh, and I don't ask you to take my word for it. Check the RAE definition ;)

I'll provide more convincing examples later.

MELERIX: http://lema.rae.es/drae/srv/search?key=desear

PR #798, opened 2015-12-02
Add 64DD IPL detection in rom browser

Nekokabu:


Source/Project64/User Interface/Rom Browser Class.cpp
```diff
@@ -781,6 +781,8 @@ int32_t CRomBrowser::GetCicChipID(uint8_t * RomData)
     case 0x000000D6497E414B: return CIC_NUS_6103;
     case 0x0000011A49F60E96: return CIC_NUS_6105;
     case 0x000000D6D5BE5580: return CIC_NUS_6106;
+	case 0x000001053BC19870: return CIC_NUS_5167; //64DD CONVERSION CIC
+	case 0x000000D2E53EF008: return CIC_NUS_8303; //64DD IPL
```
cxd4: Would it be possible to get Visual Studio to make these spaces instead of tab characters?

Just for future changes at least.

PR #802, opened 2015-12-03
Fix "Unable to load state"

samuelyuan:
Before
before

After
after

To get this message,
Select a Save Slot that doesn't exist (such as Save Slot 10)
Click "Load State (F7)"
The message "Unable to Load State"  will appear along with the full path

### DIFF
--- a/Source/Project64/N64 System/N64 Class.cpp
+++ b/Source/Project64/N64 System/N64 Class.cpp
@@ -1546,7 +1546,7 @@ bool CN64System::LoadState(const char * FileName)
             OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL | FILE_FLAG_RANDOM_ACCESS, NULL);
         if (hSaveFile == INVALID_HANDLE_VALUE)
         {
-            g_Notify->DisplayMessage(5, stdstr_f("%s %s", GS(MSG_UNABLED_LOAD_STATE), FileNameStr).ToUTF16().c_str());
+            g_Notify->DisplayMessage(5, stdwstr_f(L"%ws %ws", GS(MSG_UNABLED_LOAD_STATE), FileNameStr.ToUTF16().c_str()).c_str());
             return false;
         }
 
--- a/Source/Project64/User Interface/Main Menu Class.cpp
+++ b/Source/Project64/User Interface/Main Menu Class.cpp
@@ -523,7 +523,7 @@ bool CMainMenu::ProcessMessage(HWND hWnd, DWORD /*FromAccelerator*/, DWORD MenuI
     case ID_DEBUGGER_INTERRUPT_PI: g_BaseSystem->ExternalEvent(SysEvent_Interrupt_PI); break;
     case ID_DEBUGGER_INTERRUPT_DP: g_BaseSystem->ExternalEvent(SysEvent_Interrupt_DP); break;
     case ID_CURRENT_SAVE_DEFAULT:
-		g_Notify->DisplayMessage(3, stdwstr_f(GS(MENU_SLOT_SAVE), GetSaveSlotString(MenuID - ID_CURRENT_SAVE_DEFAULT).c_str()).c_str());
+        g_Notify->DisplayMessage(3, stdwstr_f(GS(MENU_SLOT_SAVE), GetSaveSlotString(MenuID - ID_CURRENT_SAVE_DEFAULT).c_str()).c_str());
         g_Settings->SaveDword(Game_CurrentSaveState, (DWORD)(MenuID - ID_CURRENT_SAVE_DEFAULT));
         break;
     case ID_CURRENT_SAVE_1:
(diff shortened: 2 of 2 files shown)


Frank-74:
How do I make this message appear?

cxd4:
That always happens with some projects of mine, too.  I don't know why.

You can see it in plain view as well:
https://github.com/Frank-74/project64/blob/6007d9bce0144b45bc065422fb3886b75d97a92e/Source/Project64/N64%20System/N64%20Class.cpp#L1383

My guess is GitHub has a table of valid sprintf specifiers (like %s for string, %i for int, %X for hex ... then it saw %w[...anything at all] after your change and did not recognize the wide-string mode of a printf token as a valid usage of printf.

Frank-74:
I copied the line into the load state to see it displayed, because I couldn't make it fail.
It shows the whole directory string before the state name.

This will make it say for example `Unable to load state Castlevania (U) (v1.2).pj.zip`
```cpp
	g_Notify->DisplayMessage(5, stdstr_f("%ws %s", GS(MSG_UNABLED_LOAD_STATE), CPath(FileNameStr).GetNameExtension().c_str()).ToUTF16().c_str());
```

cxd4:
```c
printf("%a", x);
printf("%b", x);
printf("%c", x);
printf("%d", x);
printf("%e", x);
printf("%f", x);
printf("%g", x);
printf("%h", x);
printf("%i", x);
printf("%j", x);
printf("%k", x);
printf("%l", x);
printf("%m", x);
printf("%n", x);
printf("%o", x);
printf("%p", x);
printf("%q", x);
printf("%r", x);
printf("%s", x);
printf("%t", x);
printf("%u", x);
printf("%v", x);
printf("%w", x);
printf("%x", x);
printf("%y", x);
printf("%z", x);
```

Frank-74:
```cpp
printf(L"ws", x);
```

cxd4:
GitHub's coolness points just increased.

It only supports ANSI C89 printf specifiers, not that C99 horseshit.

Frank-74:
@cxd4 Why then does samuelyuan's change not have a red %?

samuelyuan:
Is the message supposed to show the entire path or just the name of the save state?

Also, I can't get Save State to work for some reason.

cxd4:
um...dunno Frank..

Guess we'll find out if and when zilmar merges this.

For now samuelyuan's `%` is white, not red.  I guess after the PR is merged it turns red?

Frank-74:
@samuelyuan 
Save State (F5) is working fine here.  Can you give more details, what game?

samuelyuan:
I'm not sure which game. 

I get this error message:
error

cxd4:
Did you "Enable Debugger" checkbox through the pj64 settings to see if that error adds any detail?

Frank-74:
I only get that message when trying to Load State when a save state doesn't exist.

samuelyuan:
<code>
Error  : Break point found at N64 System\Mips\Memory Virtual Mem.cpp 2014
</code>

Frank-74:
Enable Debug messages as well in App logging.
I get this trying to load a save that doesn't exist.
```
: Debug  : CMainMenu::ProcessMessage: ID_SYSTEM_RESTORE
: Debug  : CN64System::LoadState(C:\Users\Frank\Documents\02-12-15\project64-master\Bin\Package\Save\CRUIS'N WORLD.pj): Start
: Error  : Unknown memory action

Emulation stopped
```
And its cleaner just showing the filename instead of the whole directory.  I cannot even see the filename, its cut off because of my long directory.

Frank-74:
@samuelyuan 
Try changing the g_Notify message here : https://github.com/project64/project64/blob/master/Source/Project64/N64%20System/N64%20Class.cpp#L1339
From this:
```cpp
g_Notify->DisplayError(GS(MSG_FAIL_OPEN_SAVE));
```
To this:
```cpp
g_Notify->DisplayMessage(5, stdwstr_f(L"%ws", GS(MSG_FAIL_OPEN_SAVE)).c_str());
```
Do you still get Unknown memory action on Save State?

samuelyuan:
There's still an error mainly because the program doesn't reach that point at all. I think the error happens when the program tries to write to a zip file in the lines above. 

I think this discussion should be moved to a separate issue, just in case anyone else experience a similar problem.

Frank-74:
Yes, create a new issue for this.

In the meantime, try adding debug messages throughout the SaveState() section, add 1,2,3,4 etc for each one, so you know where the program got to by reading the Project64.log.

Example:
```cpp
 WriteTrace(TraceDebug, __FUNCTION__ ": 1");
 WriteTrace(TraceDebug, __FUNCTION__ ": 2");
```

Frank-74:
Only just noticed your edit in the top post.
`To get this message,
Select a Save Slot that doesn't exist (such as Save Slot 10)
Click "Load State (F7)"
The message "Unable to Load State" will appear along with the full path`
Before this PR, I got Unknown memory action when doing this, not corrupted text.

Frank-74:
@samuelyuan 
```
This line seems to be the problem.
https://github.com/project64/project64/blob/master/Source/Project64/N64%20System/N64%20Class.cpp#L1323
zipWriteInFileInZip(file, m_MMU_VM.Rdram(), RdramSize);
```

samuelyuan:
It works after I unchecked 'Automatically compress saved states'. Thanks.

PR #805, opened 2015-12-04
SaveState Maintenance

Frank-74:
I've tested saving and loading with several games.  Using both compressed and uncompressed save states.

I also changed it to not delete the compressed save if saving non compressed file, and vice versa.


Frank-74:
Need to make some more changes to this.  @samuelyuan reported in #802 , that only uncompressed save is working for him.  Compressed save only gets to this line before crashing.
https://github.com/project64/project64/blob/master/Source/Project64/N64%20System/N64%20Class.cpp#L1323

PR #806, opened 2015-12-04
Tweak dialogs

Nekokabu:


oddMLan:
Um, interesting, what does increasing the numbers to 10 exactly do?

PR #808, opened 2015-12-04
Fixed "Saved current state" and "Loaded State"

samuelyuan:
These two strings won't display properly for every language.

If the language doesn't use the Latin alphabet at all, the program will crash.
If it uses mostly Latin characters, some characters won't display properly.

The top string is "Saved current state" and the bottom one is "Loaded State".

Before
before

After
after


samuelyuan:
I had to handle some merge conflicts, which happened to close the request

PR #817, opened 2015-12-06
Fix Visual Studio 2010+ builds on Windows XP

JunielKatarn:
Recent refactorings (removing spaces from file names) removed a small patch that made VCXPROJ builds work fine in Windows XP:
https://github.com/project64/project64/blob/012a16910f21cbeeba507543aa7e0ab5b309fd79/Source/Project64/User%20Interface/Rom%20Browser%20Class.cpp#L620

For some reason, those two `WriteTraceF` calls make displaying the ROM list impossible in XP. It has also been hard to debug. I'll look for a root cause during the week.


project64:
i re added it to see it helped with testing to see what values you got when refreshing the directory with non ascii file names. 

any idea why this actually would cause a crash ?

JunielKatarn:
> any idea why this actually would cause a crash ?

All I notice is within FillRomList, before the do-while block, calling WriteTraceF with any arguments throws an exception and makes the ROM browser list not appear at all.

The application doesn't crash at any point, though.

Something changed starting Visual Studio 2010, which triggers this behavior.
My bet is it's related somehow to WriteTraceF's implementation and Windows XP internals.

JunielKatarn:
@project64 Found the offending line:
https://github.com/project64/project64/blob/master/Source/Common/Trace.cpp#L69
`TCHAR pBuffer[nMaxSize];`, where nMaxSize is 32*1024.
For some reason, allocating any value greater than 26K throws an exception, but only on those two spots where I commented out `WriteTraceF`.

project64:
so your saying that the write trace is a problem, then any other trace would cause issues as well.

if you want a solution of how the write trace should work, look at stdstr::ArgFormat it will calculate the buffer size:

eg:
	size_t nlen = _vscprintf(strFormat, args) + 1;
    char * buffer = (char *)alloca(nlen * sizeof(char));
    buffer[nlen - 1] = 0;

cxd4:
This could be a good time for having the CRT dynamically linked, instead of statically linking it as a dependency as has always been done.  This way, debugging the run-time versions of the functions called within `WriteTrace` (assuming it is the problem).

So with a dynamically linked `msvcr100.dll` (which is what VC2010 uses under Visual Studio 2010), it would just depend on the externally debug-able functions within msvcr100 from X month of Y year that Microsoft released, which could show what, if any, point in time for a regression in the C library implementation of one or more functions crashing the ROM browser on XP.

JunielKatarn:
The issue seems to be resolved by some recent change (wild guess: commit 73549ee72f242e92083ddc6bc28c5bcdef478d38).

Closing...

PR #821, opened 2015-12-08
[core] Removed spaces from file names in VCXPROJ.

JunielKatarn:
Visual Studio 2010+ couldn't open some files in Project64-core because they still had spaces in file names.


Source/Project64-core/Project64-core.vcxproj
```diff
@@ -71,13 +71,13 @@
     <ClCompile Include="N64System\Recompiler\RecompilerOps.cpp" />
     <ClCompile Include="N64System\Recompiler\RegInfo.cpp" />
     <ClCompile Include="N64System\Recompiler\SectionInfo.cpp" />
-    <ClCompile Include="N64System\Recompiler\x86CodeLog.cpp" />
+    <ClCompile Include="N64System\Recompiler\X86CodeLog.cpp" />
```
cxd4: Correct file name was `x86CodeLog.cpp`, with a lowercase `'x'`.

Source/Project64-core/Project64-core.vcxproj
```diff
@@ -71,13 +71,13 @@
...
     <ClCompile Include="N64System\Recompiler\X86ops.cpp" />
     <ClCompile Include="N64System\SpeedLimitorClass.cpp" />
     <ClCompile Include="N64System\SystemGlobals.cpp" />
     <ClCompile Include="Plugins\AudioPlugin.cpp" />
     <ClCompile Include="Plugins\ControllerPlugin.cpp" />
-    <ClCompile Include="Plugins\GFXplugin.cpp" />
+    <ClCompile Include="Plugins\GFXPlugin.cpp" />
```
cxd4: Correct file name was `GFXplugin.cpp`, with a lowercase `'p'`.

In the future please refrain from mixing unrelated and/or wrong changes.  If a commit is named "removed spaces" it should just be about removing spaces.

JunielKatarn: In the future, I'll just rename the PR as "<....> AND fix filename casing".
Good catch, though. I believe the actual issue is the inconsistent naming (the surce file is named GFXplugin, while the header is named GFXPlugin. The second seems more correct to me).

cxd4: Thanks for the consideration.

No complaints about the PR name--mostly just the commit's name generalizing more than one change.  (Two separate commits can be debugged when doing checkouts.)

PR #833, opened 2015-12-20
[core] fixed almost all fatal compile errors for Unix build

cxd4:
Took me a lot of days, but I finally have a working Unix shell script for compiling Project64 on GNU.

All of the fatal "no such file" errors were fixed up until `#include <windows.h>`, at which point I stopped.

It's actually sort of a shame since I'm not really a fan of capitalizing file names unnecessarily--I do prefer `<common/stdtypes.h>` over `<Common/stdtypes.h>` and dislike the use of capital letters in compliant file names, but in this case we're just trying to avoid intervention with existing file names staged to the Git repository and simply try to fix all of those on the C source's end.

### DIFF
--- a/Source/Project64-core/AppInit.cpp
+++ b/Source/Project64-core/AppInit.cpp
@@ -1,6 +1,6 @@
 #include "stdafx.h"
 #include <Common/path.h>
-#include <Common/trace.h>
+#include <Common/Trace.h>
 #include <Common/Util.h>
 #include <Project64-core/N64System/Mips/MemoryVirtualMem.h>
 #include <Project64-core/N64System/SystemGlobals.h>
--- a/Source/Project64-core/Logging.cpp
+++ b/Source/Project64-core/Logging.cpp
@@ -9,7 +9,7 @@
 *                                                                           *
 ****************************************************************************/
 #include "stdafx.h"
-#include "logging.h"
+#include "Logging.h"
 #include <Common/path.h>
 #include <Project64-core/N64System/SystemGlobals.h>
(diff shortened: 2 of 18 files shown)


vgturtle127:
We can fix all the capitalization stuff later, right?

cxd4:
I'd rather not rename files in the Git history index if I don't really have to.

When you rename a file in Git, it makes it more tricky to see the full history for that file, because then you have to add the commit history of the original file to the newer commits history of the re-named file.  I made an exception for "SpeedLimitor" file names because that was just a spelling typo that would scare grade-school kids from wanting to contribute.

vgturtle127:
@cxd4 :laughing: 

OK, I was just kind of curious.

Lithium64:
@cxd4 What's missing to pj64 run on linux?

cxd4:
A working batch file that everybody can use to compile it without errors. :)

cxd4:
Get rid of some Windows-specific code is the most of it I think.  You can substitute pretty much anything you want or even delete most features for the Linux build so that the CPU emulation itself still runs.

vgturtle127:
How will we compile plugins for Linux? Will we just compile them into .so files or whatever they are? Sorry, I haven't been on Linux in a while. :blush:

cxd4:
Well we already have plugins available for Linux so I don't really care about that.

But I have also started compiling pj64 RSP plugin to a Linux port as well...Glide64 too if we can get Common.lib to link in correctly and fix those namespace issues I was having.

vgturtle127:
@cxd4 Alright, cool.

PR #835, opened 2015-12-20
[core] Staticize the CPU loop's next instruction type updates per cycle.

cxd4:
By taking advantage of special patterns between the `enum` values, we can further optimize the running CPU loop from instead of having to do this:
```c
while (CPU_running) {
    ...

    switch (next_instruction_type) {
    case NORMAL:
        JMP end_of_loop;
    case DELAY_SLOT:
        MOV DWORD PTR m_NextInstruction, JUMP;
        JMP end_of_loop;
    case PERMLOOP_DO_DELAY:
        MOV DWORD PTR m_NextInstruction, PERMLOOP_DELAY_DONE;
        JMP end_of_loop;
}
```
... to being able to do this:
```c
while (CPU_running) {
    ...

    MOV edx, (next_instruction_type);
    ADD (next_instruction_type), (next_instruction_type); # or SHL by 1
    switch (%edx) {
    case NORMAL:
        continue; # JMP start_of_loop;
    case DELAY_SLOT:
        continue; # JMP start_of_loop;
    case PERMLOOP_DO_DELAY:
        continue; # JMP start_of_loop;
}
```
So this makes the code more static and, according to a preliminary benchmark result, faster.  This code could actually be slower due to the loss of the dynamic code behavior optimizing the NORMAL case (due to the extra two instructions), but I have a subsequent idea for greatly optimizing this switch statement overall in a future pull request to continue making it faster.  It depends on these changes being tried first.

### DIFF
--- a/Source/Project64-core/N64System/Interpreter/InterpreterCPU.cpp
+++ b/Source/Project64-core/N64System/Interpreter/InterpreterCPU.cpp
@@ -302,15 +302,17 @@ void CInterpreterCPU::ExecuteCPU()
             NextTimer -= CountPerOp;
 
             PROGRAM_COUNTER += 4;
-            switch (R4300iOp::m_NextInstruction)
+            const uint32_t next_instruction_type = R4300iOp::m_NextInstruction;
+            R4300iOp::m_NextInstruction *= 2; // DELAY_SLOT --> JUMP; NORMAL --> NORMAL
+            switch (next_instruction_type)
             {
             case NORMAL:
                 break;
             case DELAY_SLOT:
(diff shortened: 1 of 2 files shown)


project64:
While I can agree that the speed increase is nice and it would make the interpreter faster, I do not want to compromise clean readable code in the interpreter.  I know the 64bit does not have a full working recompiler which is why you want to make the interpreter faster but these type of fixes I would prefer to leave for the recompiler.

cxd4:
Understandable.

It's actually sort of a surprise you've even merged my interpreter speed-ups to this point as it is, which also slightly compromised the cleanness of the code (though not quite as much as this PR did).

PR #846, opened 2015-12-23
Update .gitignore and add .gitattributes

vgturtle127:
This is just an initial PR, we can discuss changes if you guys like.

This should help in the future when we port to other systems, or just
with general cleanliness of the repo.

Feedback would be nice.

Some of this might be able to be removed, but most of it is probably system stuff or Visual Studio-related.


cxd4:
I understand you want to help take some initiative for future situations, but when it comes to the `.gitignore` I am sure the smartest thing to do is to keep it down to a minimum.  When it comes to cleanliness, nothing beats deleting code down to a minimum.  In this case you're just adding all kinds of things that I have no idea what they even are.

It should be updated on an as-needed basis.

vgturtle127:
Well, do you guys use any of the Visual Studio plugins/addons?

That would remove quite a lot of stuff. Like, most of it.

I use this .gitignore as a template, not as a definitive global .gitignore.

cxd4:
If Visual Studio plugins/add-ons was the only thing changed in your PR, I could comment better.

But in this case it isn't.  Why things like this are being inserted?
```
# Linux package management formats

*.dmg
*.xpi
*.gem
*.egg
*.deb
*.rpm
```
What do Debian, Red Hat etc. distro package file formats have to do with Project64 sources?

LibreOffice, I see you added a file extension about that.  I fail to see what Microsoft Office, LibreOffice, InstallShield and compiled Java JAR executables have to do with Project64.

I also fail to see what CVS/SVN or Mercurial files have to do with Git.

Archive files...TAR, RAR, 7z, ZIP, ISO, ... unless you are anticipating ROM images being a part of the repository I can't take a guess why you added these.

vgturtle127:
If we ever compile for Linux, people may accidentally have compiled stuff in their directories.

I removed the Microsoft Office and LibreOffice stuff, as well as the Java and InstallShield stuff. I added some stuff about Inno Setup, though.

I removed the archives and disc images and that stuff.

Git is compatible with SVN, and as such the .svn folder may be in somebodies directories when the do PR's or whatever. My thought was that we could prevent that, and not have to do a complete surgery later if we ever found out the repo was getting huge.

I can remove those too, if you want. Although from experience, I probably wouldn't recommend it.

vgturtle127:
@cxd4 I'd like some feedback on the .gitattributes as well, if you don't mind.

vgturtle127:
I'm trying to rebase, but I'm having difficulties. So there may be a slight delay.

vgturtle127:
@project64 @cxd4 You guys should review ASAP since I'm bad at upstream merges. :laughing:

vgturtle127:
@cxd4 I think I will close this, I am having trouble with the merging of newer commits. Should I bother opening another PR?

cxd4:
I'm at a loss for the right weed video to link here.

vgturtle127:
@cxd4 lolwut

Vaporeon:
https://www.youtube.com/watch?v=Puh3VxlmlRs

vgturtle127:
@Vaporeon :laughing:

PR #860, opened 2016-01-06
[core] Don't demand Win32-specific features like __cdecl.

cxd4:
Right now, this is the most important commit to have for porting to Unix because without it, the compile output spams the same hundreds of error lines related to `__cdecl` being unknown, hundreds of times.  I swear that just by making this change the console output has lost tens of thousands of bytes, making it way easier for me to read the actual, unique remaining errors with compiling it outside Windows.


cxd4:
After some discussion I was happy to change it to use a new CALL macro to replace each _cdecl.

PR #866, opened 2016-01-08
Allow for pushing two C Buttons at once when using an analog stick

Jeremy517:
These values specify how far an analog stick has to be pressed to count as a C Button press when C Buttons are mapped to an analog stick.  The old values did not allow for pressing two C buttons at once, because diagonals did not meet the values.  These values specifies that an analog stick must be pressed roughly halfway to count as a C Button press, which allows you to push two C Buttons at once.


AmbientMalice:
Very good.This'll help with games that use the c-buttons for movement.

Jeremy517:
These #defines aren't used for anything else.  Are there any concerns that are keeping this from getting merged?

vgturtle127:
@Jeremy517 I don't believe so, why? Are you thinking about changing something?

Jeremy517:
No, I was just curious as to why it hasn't been merged.

vgturtle127:
@Jeremy517 I don't think @project64 knows about/has seen it yet. :wink:

vgturtle127:
@Jeremy517 See. :wink:

PR #869, opened 2016-01-11
Add 256x256 icon

DaMan69:


toehead2001:
Your new 256px icon size has transparency holes in it.
holes

Use this as your source:
https://github.com/project64/project64/blob/master/Artwork/pj64%20icon.psd

You also removed the 24-bit icons from the file. Probably not an issue on modern systems, but you should have mentioned that.

DaMan69:
Fixed, I hope.  My git isn't gud.
If you care about Win2K/9x zilmar I'll keep the 24bit icons.

PR #876, opened 2016-01-13
Make games using FAT dynamic.

AmbientMalice:
This is needed, IMO, because the manual values were prone to crackling and poor timing. RE2 especially.


vgturtle127:
:+1: 

I think this will help with RDB cleanup as well.

project64:
the default AiCountPerBytes is 400 should this be 0 so it is calculated, or should it still be 400?

LegendOfDragoon:
@project64 0 is the best number for default AiCountPerBytes. Especially since some games change frequency mid-game (although not too many do, from what I've seen).

project64:
then lets set the default to 0 and remove the entries from the rdb

vgturtle127:
@project64 What is the maximum number?

project64:
@vgturtle127 well I guess like 0xFFFFFFF not sure what a reasonable max value would be

oddMLan:
Along setting the AI count to dynamic, setting Fixed Audio Timing to ON and Sync using Audio to OFF would make the best settings for the smoothest audio

project64:
@AmbientMalice are you able to update the pull request with what @oddMLan is suggesting .. I do not usually play with audio so I assume he is correct

AmbientMalice:
@project64 I made a pull request to do more or less what he is suggesting some time ago. https://github.com/project64/project64/pull/535 I could create a new, cleaner PR if you want. Basically, strip out all the AI entries except for that stupid Basketball game, set FAT enabled globally, and then disable it for games like World Driver Championship. Plus also remove conflicting audio sync entries.

I think this would be a huge improvement, since PJ64 is infuriatingly crackly with Jabo's audio plugin currently.

cxd4:
I would imagine separate PRs for separate improvements but I don't think it's important.

If you wanted you could even edit the topic branch (`patch-10` in this case) for the PR using GitHub website interface; as long as the branch is selected they would just get added on into the PR without having to install Git.

AmbientMalice:
I'll close this and the other PR and make a series of smaller PRs for each change.

vgturtle127:
@project64 OK, I was wondering because I was getting some error messgaes about values that were incorrect when I modified the settings, but I found out I had accidentally added some non-number characters and it was getting confused. I feel really stupid. :disappointed: 

It was a period too, so I could barely see it. :confused:

PR #878, opened 2016-01-13
[core] compile errors from not declaring standard <string.h> functions

cxd4:


project64:
I wonder if it is better to add the #include stdio and #include string to the stdafx

cxd4:
Possibly.

The con to that is more intervention from the pre-processor...if `stdafx.h` includes `string.h` in every single .cpp file throughout the source, even when string.h isn't needed, then more things like unneeded macros get added every time.

But for `<stdio.h>` maybe stdafx.h is better because that header is larger and contains more things than string...string.h I don't imagine should slow down the compile time all that much just by including it.

cxd4:
Maybe somebody with a slow computer can run some tests on how long it takes to compile Project64 with all the standard C header includes (stdlib, stdio, string, memory, math, float, limits ... http://www.tutorialspoint.com/c_standard_library/index.htm) moved out of the .cpp files and into afxres.h ?

cxd4:
https://github.com/cxd4/project64/compare/2ebf64503a...bc1405b6af

Issue #811, opened 2015-12-05
[GLideN64][Feature Request] Add support for frame buffer extension API


Frank-74:
https://github.com/gonetz/GLideN64/issues/808

@project64 
Can this be implemented in Project64?

project64:
not at the moment, 

I plan to re-write some of the memory handling soonish for android, when that happens it should be

vgturtle127:
> for android

Does Project64 work on Android? And if so, does that mean it works on Linux? Even if only basic functionality? Or am I getting my hopes up? :laughing:

project64:
@vgturtle127 it does not work for Android. But all my changes lately is so that it will run on android in the future, which is why the project64-core was made so the UI can be windows based but the core will be able to run on other OS's, mac, android, linux, etc. Still a lot more to do tho

vgturtle127:
@project64 OK. Glad to hear progress is being made, though. :smile:

oddMLan:
Related issues of interest
#72 #574 
This issue kinda could be considered a duplicate of #72, actually thinking about it.

Same goal: Faster frame buffer emulation.
Come on!! It needs to be done to truly take N64 emulation to the next level of efficiency :+1:

LegendOfDragoon:
Isn't this a duplicate of 574?

project64:
close enough to #574, do not need multiple open issues

Issue #574, opened 2015-08-26
convert CPU reads/writes into callbacks


purplemarshmallow:
I got this advice from degasus, a dolphin developer

> you should look for a way to detect when the CPU is trying to access the memory range. Those read & writes can be transformed in callbacks which to all of this ugly copying work "live". 

Copying to and from RDRAM is very ugly indeed. CPU based effects should be emulated this way. Emulator and gfx developers need to work together on this one.

This must be supported to write an accurate hardware renderer

cxd4:
https://github.com/project64/project64/issues/72

purplemarshmallow:
This is not the same as #72 
#72 is about implementing ShowCFB and FBInfo extensions to the plugin spec. These extensions just notify the gfx plugin. ShowCFB is supposed to notify the plugin about direct cpu rendering. FBInfo is supposed to notify the plugin about CPU reads from the colorbuffer area. Then the plugin must copy data to RDRAM. Copying data to RDRAM is problematic

What I mean here, any request by the CPU to read from and write to the framebuffer area must be forwarded to the gfx plugin. I think dolphin found a way to convert this into a GPU call and it uses this for correct frame- and depthbuffer emulation

cxd4:
#72 is actually not related to ShowCFB at all; that was just a late edit guessing what Gonetz meant.

It is about the slight problem of having to invent secondary workarounds to avoid bugs from asynchronous CFB accesses without dropping the performance to the tactics of software mode.

>What I mean here, any request by the CPU to read from and write to the framebuffer area must be forwarded to the gfx plugin.

Can't tell what you mean by this.  The physical CPU is perfectly capable of writing to the frame buffer on its own, independent of either the RSP and/or the RDP.  Same goes for reads.

purplemarshmallow:
>Can't tell what you mean by this. The physical CPU is perfectly capable of writing to the frame buffer on its own, independent of either the RSP and/or the RDP. Same goes for reads.

Exactly this is the problem. Currently only software gfx plugins can emulate this properly. If you use a hardware renderer the frame buffer is stored on the video card. The CPU must access data in the video card instead.

cxd4:
Right, but even with a hardware-accelerated render plugin, the frame buffer can still be stored on the emulated N64 video card (either in emulated SP DMEM, TMEM or RDRAM) and not solely, just on the physical PC video card.

In the case of GL[ES], that traditionally is done with `glReadPixels`.

[...]

Issue #72, opened 2015-02-11
Bug in GFX plugin spec. implementation


Fanatic-64:
Re-reading Gonetz's GLideN64 blog I noticed this bit:

> Technically, emulation of CPU rendering is easy. Image to display is stored in RDRAM, its address is known. The image is just loaded as texture to video card and rendered as full screen rectangle. The problem is the same as with CPU based fb effects: graphics plugin does not know that CPU started rendering by itself, emulator does not notify the plugin about it. **To be honest, plugin's specification has the function for such notification, but it almost never works.** Thus, plugin can’t rely on emulator’s help.

It would be nice if the function to notify the graphics plugin that the CPU has begun writing to the color framebuffer was fixed. It would allow correct emulation of such instances (which currently relies on a hack), and frankly it's a bit embarrassing that Project64 doesn't have a fully correct implementation of Zilmar's specification.

And by the way, would it be hard to document the 1.7+ plugin specs (Audio 1.2, GFX 1.4, Input, 1.2, RSP 1.2)?

EDIT: It appears this is the function in the spec that needs fixing:

> /******************************************************************
>   Function: ShowCFB
>   Purpose:  Useally once Dlists are started being displayed, cfb is
>             ignored. This function tells the dll to start displaying
>             them again.
>   input:    none
>   output:   none
> *******************************************************************/ 
> EXPORT void CALL ShowCFB (void);

death-droid:
What your interested in is the FrameBufferRead, and FrameBufferWrite functions that are defined in the plugin specifications but are currently disabled. Definitely something thats more up Zilmars alley, as to be honest... I have no idea how it works :P

BTW ShowCFB is unrelated to this, and as far as I know, should work as intended.

@project64

cxd4:
They're not really defined in the plugin specs, but in schibo-Rice extension to them.  Many graphics plugins include them anyway, but the spec itself mandates nothing about supporting them.  zilmar would not have known to implement something like this, so it's an unofficial extension.

It's not really a "bug" so much as it is just not knowing enough about hpw graphics work.  The spec was mostly created to just "work", and it does.  The unfortunate thing about it is that other emulators have to follow the same example or introduce unofficial extensions like this.

cxd4:
> It would be nice if the function to notify the graphics plugin that the CPU has begun writing to the color framebuffer was fixed. It would allow correct emulation of such instances (which currently relies on a hack),

works already in angrylion RDP; I think he's just talking about a hardware-accelerated approach

[...]

PR #883, opened 2016-01-15
Glide64 change default plugin

AmbientMalice:
I'm splitting the old PR into two different PRs. One (this one) to switch the default plugin -- and another, far more extensive and painful, to update the RDB to reflect Glide64's much higher compatiblity. (And unique bugs of its own.)

### DIFF
--- a/Source/Project64-core/Settings/SettingsClass.cpp
+++ b/Source/Project64-core/Settings/SettingsClass.cpp
@@ -343,7 +343,7 @@ void CSettings::AddHowToHandleSetting()
 
     //Plugin
     AddHandler(Plugin_RSP_Current, new CSettingTypeApplication("Plugin", "RSP Dll", "RSP\\RSP 1.7.dll"));
-    AddHandler(Plugin_GFX_Current, new CSettingTypeApplication("Plugin", "Graphics Dll", "GFX\\Jabo_Direct3D8.dll"));
+    AddHandler(Plugin_GFX_Current, new CSettingTypeApplication("Plugin", "Graphics Dll", "GFX\\PJ64Glide64.dll"));
     AddHandler(Plugin_AUDIO_Current, new CSettingTypeApplication("Plugin", "Audio Dll", "Audio\\Jabo_Dsound.dll"));
     AddHandler(Plugin_CONT_Current, new CSettingTypeApplication("Plugin", "Controller Dll", "Input\\PJ64_NRage.dll"));
 


project64:
maybe we should look at showing different status in the rombrowser based off the plugin you have selected.

AmbientMalice:
It's not a bad idea, but it's a very complicated one.

cxd4:
Tbh, I typically find your comparisons between Jabo's Direct3D being inferior to Glide64 to be rather exaggerated, so I think it's necessary evidence.  The method project64 suggested I think is overboard, though, because infinitely many games' statuses could all change at once for any and all plugins under development.

I don't really care what the default plugins are, though.  Me typically using PJGlide64 by default is just representative of me not caring && that it's open-source control in the rest of the repo so why not.

Frank-74:
It would be better to be able to select your prefered default plugins in the first run wizard.

AmbientMalice:
>It would be better to be able to select your prefered default plugins in the first run wizard.

There are several problems.

1: Jabo isn't portable. It's automatically out of the running for any ports of PJ64.

2: If PJ64 doesn't have a "default" video plugin, the entire compatibility list becomes a headache. (Bear in mind this doesn't invalidate the idea of showing players the available plugins.) Jabo's video plugin is broken. Glide64 is also broken, but it's *less* broken. At least, in the sense that games like Paper Mario and Conker and Banjo don't spew rainbow vomit of the screen. In the sense that Perfect Dark has functional lighting.

3: Glide64's bugs can be a problem. Perfect Dark crashes when entering the combat simulator if framebuffer is enabled. There is a specific fairy that can't be photographed in DK64 using Glide64. Hardware framebuffer support has caused no end of stability headaches for Glide64 and GLideN64.

4: The wizard could show "helpful" images like this, with the message, "Which of these is more visually appealing to you? If you prefer the left image, select "Jabo's", if you prefer the right image, select Glide64".

visual_appeal

In the meanwhile, I'm gonna work on rewriting the RDB to be Glide64-specific. That way there's SOMETHING to work with, regardless of the final method. It could be that the revised RDB and the old RDB could be merged or split into plugin-specific versions or something. But first the Glide64 version has to be written which is kind of a big job.

Lithium64:
Glide64 for compatibility and portability should be the default plugin

OpenGL 2.0 support currently is a very low system requirement

LegendOfDragoon:
>Glide64 for compatibility and portability should be the default plugin
OpenGL 2.0 support currently is a very low system requirement

@Lithium64 Glide64 is slow.

AmbientMalice:
>Glide64 is slow.

It's not THAT slow. And in the event PJ64 gets framebuffer notification support, the major performance hurdle will disappear. It's not shocker that Glide64 is somewhat faster than a plugin that crashes into the ground when you turn on its ancient, barely working framebuffer to RDRAM support. Jabo is fast primarily because copying framebuffer to RAM destroys performance and it doesn't even bother, hence the rampant bugs. Heck, FB to RAM destroys performance in Dolphin, and they've been around for donkey's years and had all the "optimisation" an emulator could ask for.

LegendOfDragoon:
>It's not THAT slow. And in the event PJ64 gets framebuffer notification support, the major performance hurdle will disappear. It's not a shocker that Glide64 is somewhat faster than a plugin that crashes into the ground when you turn on its ancient, barely working framebuffer to RDRAM support. Jabo is fast primarily because copying framebuffer to RAM destroys performance and it doesn't even bother, hence the rampant bugs.

For me it is too slow. I already disable fb emulation in most cases because of the abysmal performance. Yes, Jabo's framebuffer copying is even slower, but no hardware rendering plugin has good fb copy performance. I'm better off using an optimized software renderer, for games that heavily use frame buffer.

A good example of bad performance on my end is Dark Rift. I sometimes get 60 VI/s but it often drops below with Glide64. With Jabo's, the performance is better but I still occasionally drop below.

AmbientMalice:
>I'm better off using an optimized software renderer, for games that heavily use frame buffer.

There is no affordable PC in existence that can currently satisfactorily run games that use heavy frame buffer using Angrylion's.

>A good example of bad performance on my end is Dark Rift. I sometimes get 60 VI/s but it often drops below with Glide64. With Jabo's, the performance is better but I still occasionally drop below.

I'll take a look at this and compare. What is your hardware? You've probably mentioned it before, but I've forgotten.

LegendOfDragoon:
>There is no affordable PC in existence that can currently satisfactorily run games that use heavy frame (or complex visual effects) buffer using Angrylion's.

Aside from vigilante 8, what games heavily use frame buffer? True, that is the case for Angrylion's current code. But it has so much potential that, so I have more faith in that than any currently existing hardware rendering plugin. I heard even star wars ep 1 racer doesn't run full speed in GlideN64. To me, that is bad news.

>I'll take a look at this and compare. What is your hardware? You've probably mentioned it before, but I've forgotten.

I have a 2.667 ghz i5 quad, dell laptop with intel integrated graphics.

AmbientMalice:
>Aside from vigilante 8, what games heavily use frame buffer?

Have you tried Quake II? I think "heavy" might be the wrong word. High resolution framebuffers and odd framebuffer behaviors as seen with Vig 8 and Quake II and possibly some other games seem to hit Angrylion's the hardest.

>I have a 2.667 ghz i5 quad, dell laptop with intel integrated graphics.

You're running in performance mode, I assume? That's not bad hardware.

>I heard even star wars ep 1 racer doesn't run full speed in GlideN64. To me, that is bad news.

GLideN64 is broken in a variety of ways. Ep 1 Racer acts up partially because I set counter factor to 1, which forces the plugin to copy far more frame and depth buffers that it ordinarily would have to. The game also has unusual depth buffer behavior.

The plugin crawls with Mortal Kombat 4, to pick a game. It's not that unusual that it has performance problems in addition to its half-broken depth buffer and resource-intensive framebuffer. Take Ziggy's LLE plugin, on which GLideN64's LLE emulation is based -- it has horrible performance issues in certain games that are caused by bugs. Turok games don't run terribly with z64 and GLideN64 in LLE because they're super, super complex. They run terribly because there's something wrong with the plugins, as hinted at by the game's geometry crapping itself when you look in certain directions.

AmbientMalice:
@LegendOfDragoon To put things in perspective, Episode I racer runs at a blistering 10VIs on my machine with Angrylion's. (Slowing to 7VIs when I pause.) It runs between 50-60VIs on GLideN64 with the depth buffer enabled. Quite a bit faster with it disabled. GLideN64 has some serious performance problems, but they're nowhere near as dramatic as Angrylion's, at least on an older 3.2Ghz CPU.

LegendOfDragoon:
>Have you tried Quake II? I think "heavy" might be the wrong word. High resolution framebuffers and odd framebuffer behaviors as seen with Vig 8 and Quake II and possibly some other games seem to hit Angrylion's the hardest. edit: Last I checked, all the Factor 5 games run really, really badly in High Resolution mode with Angrylion's. We're talking single digits in places.

Are you playing with counter factor 1 for quake? Counter factor 2 seemed ok, judging by the fact that most people seem to have better hardware than me. Many games will be full speed if we have an optimized 64bit version of Angrylion's + 64bit CPU and RSP recompilers, even on my machine.

Actually i just tried counter factor 1 and the performance was significantly worse in Quake II. I'm starting to wonder if maybe part of the performance problem in some of these intensive games is that the emulator might be running at a higher speed than it's supposed to. For instance, star wars episode 1 racer is much faster in CF=3, but I did not notice any frame skipping. I tried frame stepping a few dozen times and it appeared to be fine.

The interesting thing about Quake II is that I'm pretty sure it doesn't use VI filters at all. At least the parts of the game I checked anyway. So optimizing the VI emulation code could be helpful for that game.

>You're running in performance mode, I assume? That's not bad hardware.

Yes, I run it in performance mode. I'm not sure why the performance is hit or miss on my end, depending on the game. I need to look into figuring out how to profile OGL and D3D code, so that I can find out what's going on and maybe fix the problem myself.

>Take Ziggy's LLE plugin, on which GLideN64's LLE emulation is based -- it has horrible performance issues in certain games that are caused by bugs.

I wonder if that's why I get bad performance in SD Hiryuu with z64gl.

>@LegendOfDragoon To put things in perspective, Episode I racer runs at a blistering 10VIs on my machine with Angrylion's. (Slowing to 7VIs when I pause.)

I take it you were running counter factor 1 and high res? I guess part of the difference is that I usually play games in low res, so that I can get better performance.

>It runs between 50-60VIs on GLideN64 with the depth buffer enabled. Quite a bit faster with it disabled. GLideN64 has some serious performance problems, but they're nowhere near as dramatic as Angrylion's, at least on an older 3.2Ghz CPU.

I guess I underestimated the performance of GLideN64. Are there any games that will need 4+ ghz cpu with GLideN64 to run full speed?

Lithium64:
@LegendOfDragoon Even on my worst hardware it runs fine

purplemarshmallow:
>Take Ziggy's LLE plugin, on which GLideN64's LLE emulation is based -- it has horrible performance issues in certain games that are caused by bugs.

I fixed TMEM emulation in z64gl with code from angrylion's plugin: https://github.com/purplemarshmallow/z64/tree/angrylion-integration . Many bugs are gone but there is this regression where some textures get corrupted like the title screen in Super Smash bros. I was not able to fix it yet unfortunately.

https://github.com/purplemarshmallow/z64/tree/angrylion-integration

>I wonder if that's why I get bad performance in SD Hiryuu with z64gl.

SD Hiryuu is fixed in my branch I tested it

LegendOfDragoon:
>@LegendOfDragoon Even on my worst hardware it runs full speed

What game(s)?

>SD Hiryuu is fixed in my branch I tested it

I tested the game yesterday with your branch and even tried to profile to see what was causing the poor performance in-game. The very first menu is really slow. Also triangles are very buggy in this game.

purplemarshmallow:
I don't get these results. The official z64gl is slow for me. But my branch is really fast always runs fullspeed easily. But triangles are buggy I noticed it too.

AmbientMalice:
@purplemarshmallow When GLideN64 is rendering LLE, how much of it is copy-pasted z64gl code? Will fixing z64gl have benefits for GLideN64? Or maybe it should go the other way? Port GLideN64's features over to z64gl to make it a full featured LLE hardware plugin? I'd like to see PJ64 switching to LLE graphics with HLE as a fallback for old hardware. The compatibility improvements would be dramatic.

(This PR is spiralling off topic, but I suppose there's no harm in it.)

Lithium64:
@LegendOfDragoon Every game that I have tested

LegendOfDragoon:
>@LegendOfDragoon Every game that I have tested

I wish I could say the same.

>I don't get these results. The official z64gl is slow for me. But my branch is really fast always runs fullspeed easily. But triangles are buggy I noticed it too.

That's odd. I just tried original z64gl and the speed's hardly any different. In a match, I barely get full speed, but since my performance fluxes, i typically need to be able to run well past 60 for consistent full speed.

>Will fixing z64gl have benefits for GLideN64?

Certainly. I think the texture loading code is already different in GLideN64, but the LLE triangle code is similar if not the same.

>I'd like to see PJ64 switching to LLE graphics with HLE as a fallback for old hardware. The compatibility improvements would be dramatic.

I don't think it would be suitable anytime soon. Not just because of poor performance either.

AmbientMalice:
>I don't think it would be suitable anytime soon. Not just because of poor performance either.

What do you mean? Arguably LLE performance can be dramatically improved by optimising the RSP and the video plugin. I'm not saying it'll happen overnight, but I can't really see a long-term future for HLE video. It's such a buggy, hacky mess. Are there other reasons LLE is a problem?

cxd4:
Not really.  HLE is fundamentally based on the fact that LLE is possible, so I think he's just talking about lack of information and experience.

LLE naturally comes first because HLE derived from anything else is essentially just using somebody else's code to fill in the gaps.  It's also at a greater risk of premature optimization if you start from HLE and then do LLE; it needs to be the other way around for a smarter long-term goal.

LegendOfDragoon:
>What do you mean? I'm not saying it'll happen overnight, but I can't really see a long-term future for HLE video. It's such a buggy, hacky mess. Are there other reasons LLE is a problem?

z64gl has many graphical glitches. For instance, I'm better off using Glide64 Final if I want to play SD Hiryuu. I'm sure GLideN64 LLE also has its fair share of problems other than performance. These things will take quite a bit of time to fix, especially considering that not many people are actively working on it. But ya, my main point was just that it will take a while to improve LLE graphics to the point where it would be the default.

>Arguably LLE performance can be dramatically improved by optimising the RSP and the video plugin.

It could certainly improve, but even I don't know how I could speed up z64gl. I want to try looking into figuring out how to profile OpenGL code. Otherwise I'm out of luck on that.

purplemarshmallow:
>Will fixing z64gl have benefits for GLideN64?

They are very different only the LLE triangle command in GLideN64 is taken from z64gl. Improving LLE triangles will help both plugins.

>In a match, I barely get full speed, but since my performance fluxes, i typically need to be able to run well past 60 for consistent full speed.

I get 200% speed ingame with z64gl and my system is not so powerful. Geforce GT 525M + intel i5 2450M 2.50 GHz

LegendOfDragoon:
>I get 200% speed ingame with z64gl and my system is not so powerful. Geforce GT 525M + intel i5 2450M 2.50 GHz

Graphics hardware makes all the difference I guess.

I'd be more in favor of making Glide64 the default if it was ported completely to OpenGL and improved certain things like gamma. I think currently it's intended to be on if you're running it in full screen. But when I tested it with Star Fox 64, it would only appear to be on briefly on my end. W/o gamma, I have no real reason to use Glide64 for Star Fox 64.

I believe Glide64Mk2 is purely OpenGL.

project64:
@LegendOfDragoon  in what way is not 100% opengl .. The only thing I can think of is the code has few extra layers making a call like glide that is then translated in to the open GL call .. but that is just collapsing functions.

AmbientMalice:
@LegendOfDragoon Fullscreen gamma correction seems to be working fine with Star Fox 64 on my end. Not sure whether it's accurately set, but the screen brightens and then darkens just before fullscreen is disabled.

The real problem with Glide64 is, ironically, that it has gamma correction and if the emulator crashes in fullscreen, the gamma correction won't be disabled unless you use a tool to reset it to 1.0 or reboot.

LegendOfDragoon:
>@LegendOfDragoon in what way is not 100% opengl .. The only thing I can think of is the code has few extra layers making a call like glide that is then translated in to the open GL call .. but that is just collapsing functions.

I just feel it's more maintainable to use OpenGL, instead of using a wrapper. I'd be more inclined to work on Glide64 if I understood the code better. I'd rather learn OpenGL than to try and understand the original Glide code. For instance, I tried seeing what's wrong with gamma on my end and I have no clue how this code works https://github.com/project64/project64/blob/master/Source/Glide64/Main.cpp#L1972 .

>@LegendOfDragoon Fullscreen gamma correction seems to be working fine with Star Fox 64 on my end. Not sure whether it's accurately set, but the screen brightens and then darkens just before fullscreen is disabled.

Well, there goes another reason why my opinion may differ from others for this plugin. Why does it require full screen mode anyway? I actually prefer playing in windowed mode. If it worked on my end, I'd probably use Glide64 for StarFox64 ;/ .

AmbientMalice:
https://github.com/project64/project64/blob/master/Source/Glide64/Main.cpp#L2251

Basically, the first stage of setting gamma is checking if fullscreen is enabled. Maybe it could check an ini value like WindowedGammaCorrection? The gamma correction method seems to be really, really janky.

cxd4:
>>@LegendOfDragoon in what way is not 100% opengl ..

>I just feel it's more maintainable to use OpenGL, instead of using a wrapper.

This doesn't seem to answer his question, though, because it is in fact using OpenGL.
There's no run-time wrapper if the functions are statically compiled in.

LegendOfDragoon:
>The gamma correction method seems to be really, really janky.

How is it janky?

>This doesn't seem to answer his question, though, because it is in fact using OpenGL.
There's no run-time wrapper if the functions are statically compiled in.

Well the main problem for me is the source code itself. I'd rather work with OpenGL source code, than to try understanding the Glide code as well. The first thing I'd like to do is get gamma working on my end.

LegendOfDragoon:
>@LegendOfDragoon Even on my worst hardware it runs full speed. Every game that I have tested

@Lithium64 I have a problem with the fact that I get better speed with a LLE plugin for Dark Rift.. Can't blame hardware if I can get full speed with LLE.
LLE
dark rift lle software render

Glide64Final
dark rift glide64

AmbientMalice:
@LegendOfDragoon It's fairly normal for certain titles to crap themselves performance-wise in HLE.

LegendOfDragoon:
>@LegendOfDragoon It's fairly normal for certain titles to crap themselves performance-wise in HLE.
>Isn't Dark Rift a turbo3d game? Probably bears investigation.

Yes it is a turbo3d game.

cxd4:
_Dark Rift_ constantly switches microcodes between e.g. Turbo3D and friends, so HLE isn't that efficient.

AmbientMalice:
I was reminded that turbo3d games crash if the framebuffer is turned on, so https://github.com/project64/project64/pull/913

AmbientMalice:
>How is it janky?

It uses Glide gamma control, as far as I can tell, which probably explains why Gamma Helper fixes the gamma if Glide64 crashes. A far more sensible solution would be applying the gamma adjustment to the framebuffer output, like GLideN64 does.

LegendOfDragoon:
>It uses Glide gamma control, as far as I can tell, which probably explains why Gamma Helper fixes the gamma if Glide64 crashes.

Oh wow lol. Now it all makes sense now.

>A far more sensible solution would be applying the gamma adjustment to the framebuffer output, like GLideN64 does.

Absolutely. I think we'd be better off if GLideN64 just had better hardware compatibility. It would be better than putting a lot of work into Glide64, now that I think about it..

project64:
I know I am focusing on glide64 mostly because it has been well tested and generally good compatibility .. but I do dislike that there are two main plugins being worked on .. well really that glide64 is more being maintained then worked on .. I would prefer to focus on the one that is actually being worked on, just need better compatibility

with something like grGetGammaTableExt and GetGammaTable it is the type of thing where grGetGammaTableExt  should be collapsed. where I mean delete the function and take the code and move it in to GetGammaTable so you do not have this extra layer of abstraction

purplemarshmallow:
I want to bring this up since Glide64 does not support LLE GFX which LLE GFX plugin will PJ64 use?

There are: 
z64gl:
+fastest
+compatible with games that need LLE (Factor 5 games, BOSS games)
-low compatibility with other games

Jabo's LLE:
+possibly fastest on systems with poor OpenGL driver
-low compatibility
-closed source

GLideN64 LLE:
+supports texture enhancements
~average compatibility
-slow
-high system requirements
-no fog emulation

angrylion's plugin:
+highest compatibility
+highest accuracy
-slowest

AmbientMalice:
>I want to bring this up since Glide64 does not support LLE GFX which LLE GFX plugin will PJ64 use?

That's a good question. I don't think Jabo's is suitable as an LLE plugin simply because it is so hopelessly buggy, especially on the games that actually need LLE. A polished z64gl would be an excellent option, assuming the bug that causes up-and-down juddering with PJ64 gets fixed.

GLideN64 is wonderful, but its LLE is woefully buggy, and also very slow.

AmbientMalice:
With your z64gl fork, @purplemarshmallow, I get around 160-170fps in World Driver Championship. That's over 2x faster than GLideN64.

AmbientMalice:
The angrylion z64gl fork runs Turok 3 better than any other hardware plugin, ignoring some clipping issues when standing close to walls.

Runs 60-200fps? Check.
Shadows work? Hard black, and a bit flickery, but... Check.
Goggles work? Check.
Dynamic lighting works?

depressing

I sit here and despair that a polished, optimised version of this plugin hasn't replaced everything else already.

purplemarshmallow:
>I kinda dislike how GLideN64 and z64gl are both trying to eat the same piece of pie.

GLideN64 never focused much on LLE

>The angrylion z64gl fork runs Turok 3 better than any other hardware plugin, ignoring some clipping issues when standing close to walls.

angrylion's changes are also in my fork

z64gl needs some work to support more games. I am now trying to fix TMEM emulation this is a major problem with z64gl. The framebuffer code is also very unstable but it works surprisingly well in many cases. If you compile the plugin with NOFBO defined it will be much less likely to crash.

AmbientMalice:
@purplemarshmallow I was referring to your fork, dude. It's amazing. Toy Story 2 runs near-perfectly. 200-300VIs. Even the little laser bar thing renders correctly.

purplemarshmallow:
I want to fix z64gl from the ground up. Not try to get games working first. First try to have correct code and use the information from people like angrylion and mooglyguy who reversed the RDP and tested their code against real hardware. And compare results.

This is more an experiment. In long-term it should give better results. But in short term it can cause regressions if other parts of the plugin rely on incorrect code.

Parts of z64gl are similar to MAME and angrylion's plugin because all 3 emulators are based on Ville Linde's MESS code.

AmbientMalice:
>I want to fix z64gl from the ground up. Not try to get games working first.

I agree 100%. Clean, accurate code trumps immediate speed gains. I see nothing wrong with accuracy improvements leading to apparent regressions. But this is really, really promising stuff.

LegendOfDragoon:
My reasons for caring about performance extend beyond just getting 60 VI/s. Even after factoring in temperature (my laptop has poor cooling), there are other important things such as audio performance. Some games are really sensitive when it comes to audio. I can run Mario Kart at ~90 VI/s in my Angrylion fork, but the audio still suffers. Therefore it's not yet suitable for that game unless I'm playing with no sound on. Body Harvest is probably the most sensitive game I've tested, just about any video plugin makes the audio worse for me. Only No Video works fine ;/ . Top Gear Rally is pretty sensitive too, at least on my end. I can run Glide64 at ~ 150 VI/s, but it crackles a lot. Yet D3D9 fork on Rice Video runs nearly 400 Vi/s and I get nearly perfect audio! So in that case, I'd gladly sacrifice visuals for better audio. Finding out how to get good audio in Top Gear inspired me to look into either speeding up an existing hardware rendering plugin, or writing my own.

I think z64gl has the most potential for LLE, since people here want hi-res and performance. However, I think it needs lots of work, but the same is true for the other plugins. I'd like to see the OpenGL code become more stable.

purplemarshmallow:
>I can run Mario Kart at ~90 VI/s in my Angrylion fork, but the audio still suffers.

Nice work that's a huge speedup. If angrylion's plugin was faster I think it would be a good alternative.

>I'd like to see the OpenGL code become more stable.

It's the framebuffer emulation code If you compile with NOFBO and lowres = 1 it should be stable. Framebuffer emulation is hard but I'd love to get games like Super Bowling or Vigilante8 working. No hardware gfx plugin can display them correctly. I want to experiment with it but I can't promise anything.

LegendOfDragoon:
>Nice work that's a huge speedup. If angrylion's plugin was faster I think it would be a good alternative

Thanks, I should have mentioned that it was with filters off. Although the filter code could easily be sped up. The loops can be optimized. Since many games are not 60 fps, you could take big advantage of skipping frames too. Sure it's a little hacky, but worth it imo. Angrylion's code can easily be sped up, it just becomes a lot messier. Function splitting yields very good improvements, but requires you to write a ton of code.

>It's the framebuffer emulation code If you compile with NOFBO and lowres = 1 it should be stable. Framebuffer emulation is hard but I'd love to get games like Super Bowling or Vigilante8 working. No hardware gfx plugin can display them correctly. I want to experiment with it but I can't promise anything.
Does Mario Tennis work properly in z64gl? If not, maybe you could start with that game since GLideN64 does a decent job with it.

purplemarshmallow:
I'm afraid of bringing over some misconceptions. I think GLideN64 framebuffer code relies too much on VI values. On real hardware the VI and the RDP are very separate. z64gl's framebuffer code is not as clean and easy to work with unfortunately.

z64gl can render the splitscreen effect before a match in Mario Tennis, GLideN64 can't. But z64gl won't last long until it crashes and many other things don't look correct.

LegendOfDragoon:
>I'm afraid of bringing over some misconceptions. I think GLideN64 framebuffer code relies too much on VI values.

In that case, never-mind. I wish I could figure out why some games have poor performance on my end with z64gl.

AmbientMalice:
>I wish I could figure out why some games have poor performance on my end with z64gl.

Which games?

LegendOfDragoon:
>Which games?

One example is World Driver Championship, I get ~50 something VI/s in game with z64gl. Yet with Jabo's I get around 95 VI/s. Then there's also SD Hiryuu where some menus are really slow, but that happens all hardware plugins for me.

dsx-:
threaded=1 and async=1?

AmbientMalice:
>threaded=1 and async=1?

purplemarshmallow's fork doesn't come with an ini file, and looks to have threaded disabled, based on the most recent commit. I'm not sure whether it can read the generic one. Maybe people should delete the z64gl ini file if it exists in their PJ64 installation. I get high framerates with no settings adjustment.

Multithreading z64gl was always a bad idea, at least with Ziggy's builds. It made the plugin very unstable and bought minimal speed increases.

dsx-:
was trying to think of possible reasons why z64gl is slow for legend but yeah youre right, threaded=0 here makes very little difference for me.

dsx-:
gonna try out purple's fork now actually

AmbientMalice:
@dsx-

Use this branch. https://github.com/purplemarshmallow/z64/tree/angrylion-integration

And grab the SDL.dll and glew.dll files from here and place them in the PJ64 main directory. https://drive.google.com/file/d/0B7Y6r4SpC_QQdzVuWURyMmkxX2M/view?usp=sharing

dsx-:
yeah already did that haha, man wdc runs really fast

LegendOfDragoon:
Tried async and it's a little faster, but I still don't get 60 unfortunately.

dsx-:
ok yeah z64gl.conf does work with purple's fork. async=1. just for reference, the VI/s I get in the menu with his fork are 250 with async=0, 500 with async=1.

dsx-:
when building with that branch: http://puu.sh/mDTwt/5768259407.png

dsx-:
ok yeah the angrylion-integration branch i get 250VI/s in menu, 120VI/s in-game

Lithium64:
@project64 The best way would be to create a code to check the user's hardware and determine the default plugin based on this information

AmbientMalice:
@Lithium64 On what metric, though? Glide *should* work on any piece of hardware made 2006. Maybe even earlier for CPUs. Is Glide64 using GL 2.1? I can't recall. If it's using an older spec, that pushes the acceptable hardware back even further.

Lithium64:
@AmbientMalice OpenGL 2.0 or higher Glide64, OpenGL 1.5 or lower Jabo and on future OpenGL 4.2 and higher GlideN64

Some bad drivers should be blacklisted for compatibility reasons

Lithium64:
Another possibility would be porting Glide64 for DirectX 9 on Windows since it has better drivers

vgturtle127:
@Lithium64 Porting Glide64 to DirectX 9 seems like a lot of work for an older plugin. Detecting user hardware would be better, but that seems like a pain to maintain. Also, we would have to keep adding older and newer hardware and changing stuff for driver, BIOS, and software updates. Seems messy.

Bad drivers should probably not be blacklisted, again that seems like it would be hard to maintain. I'm not advocating that we hack together workarounds either, but that also seems like it would be difficult to maintain, and also add to the hackish nature of N64 emulation.

Does anybody here use Windows XP, or older hardware like Glide-supported hardware or older Pentium 4 or older CPUs?

cxd4:
Right, because BIOS updates are an important part to any hardware compatibility detection involved with Direct3D 9 vs. GLIDE.

ExtremeDude2:
Do you even bios bro?

vgturtle127:
@cxd4 @ExtremeDude2 I meant graphics card BIOS'. lol

cxd4:
You need DirectDraw to use those, not glide.

vgturtle127:
@cxd4 You need DirectDraw to use a graphics card BIOS?

cxd4:
Either that or Direct3D9 like you said.

vgturtle127:
@cxd4 That's weird, so you have to have a card with a certain graphics API in order to support a graphics card BIOS? I had never heard of such a limitation.

I suppose they didn't implement them sooner than that anyway...

project64:
I might look at it for 2.4, I was looking more at glide64 lately .. there are small things that jump at me that need fixing first like fullscreen should default the current resolution that is being run. Also with me changing so much it probably needs more testing.

AmbientMalice:
@project64 The biggest problem with Glide64 remains PJ64's lack of framebuffer notifications. Without them, stuff like Banjo Kazooie can't be played with horrible performance overheads or broken puzzle pieces. Have you made any progress in that area?

Also, one "fix" Glide64 needs is a second aspect ratio override. Basically, some games NEED a non 4:3/16:9 setting. Starcraft and RE2, for example. But if we could have an aspect ratio setting in the main graphics tab, that would be a huge help for normal users playing natively widescreen N64 games. Maybe the aspect ratio tab should be moved to the main window, trimmed to 4:3, 16:9, and then those special modes set in the ini and those modes override the user's aspect ratio choice.

Now that I think about it, any N64 game set to "stretch" for whatever reason will fill a 16:9/16:10 screen. Can't remember which ones do that, but it's a problem.

project64:
@AmbientMalice I got as far as showing that using function based look up was to big of a performance hit to justify using it ... I would much prefer to spend my time looking at how do I create a hack for the individual games to have a good effect.

theboy181:
If you want accuracy use AL's RDP :)

Zilmar on track when it comes to performance.

AmbientMalice:
>I would much prefer to spend my time looking at how do I create a hack for the individual games to have a good effect.

That sounds highly improbable. Reading and writing to RDRAM is extremely crucial to many N64 games. There's no practical way to avoid that as far as I'm aware. (Not to say it's impossible, but seriously...) Trying to create hacky solutions for hundreds of games is going to prove a nightmare. And I seriously doubt it will work.

Without fast RDRAM framebuffer emulation, PJ64 doesn't stand a chance in the long run against 1964 and mupen, both of which can run these games accurately with acceptable performance.

Also, N64 emulators don't just need to be able to detect framebuffer reads, they need to be able to detect CPU modifications to the framebuffer. (Games like Pokemon Snap need this for the red dot, for example.)

>Zilmar on track when it comes to performance.

What's the point of "performance" when the emulator will remain fundamentally broken for a wide range of games? Something has to be done in this area at some point. Otherwise we may as well abandon the entire core and go with 1964's. PJ64 *must* be able to copy specific parts of the framebuffer, and do so rapidly. Or at least do something similar. And do so without a nightmare tangle of per-game and per-region hacks.

theboy181:
Good point about the hacks.. But I like the enhancements..
Maybe we could have more options in PJ just to address accuracy, and enhancements. Have you attempted to create a PR for the feature to be added? Has @weinerschnitzel shared the code?

AmbientMalice:
@theboy181

"You like the enhancements". That's the problem. PJ64 currently can't emulate games ranging from Banjo to Perfect Dark to Conker to random golf games at full speed with framebuffer effects. The emulator suffers huge speed problems with my other PC, which has a 3.5GHz i7 CPU. Copying every frame is just too slow. There is no "enhancement" at stake here. I'm not sure @project64 realises just how bad PJ64's performance is right now. If he does, the performance hit of his FB notification experiments must be WORSE than Perfect Dark slowing to 30-40VIs on a 3.2Ghz CPU, as it does currently with Glide64. Even GLideN64 struggles.

PJ64 can't emulate these games at full speed on most PCs. The overheads of copying 30-60 frames at 320x240-480x360 to RAM every second are obscene.

>Have you attempted to create a PR for the feature to be added? Has @weinerschnitzelshared the code?

1964 works differently. It has been able to notify the plugin for years. The recent changes just make the process faster by preventing unnecessary copies.

vgturtle127:
@theboy181 @weinerschnitzel did share the source of everything he is working on, as far as I know.

* https://github.com/weinerschnitzel
* http://sourceforge.net/projects/surreal64ce/

@project64 When did you do that work and try it? Because if it was quite a while ago, you may want to consider trying it again after the cleanup and portabilization work is done. It may be a huge performance hit, but you could have an option to enable/disable it. Just a thought.

@theboy181 I would argue that "use Angrylion's RDP for accuracy" is kind of like getting compliant with whatever we have. If we continue to improve other plugins, eventually they will be accurate enough to play games at full speed, but also be accurate as well. Besides, you have to have a pretty decent computer to use Angrylion's RDP at full speed with many games. Not that you can't run every game with Angrylion's RDP, and not that you can't run some games at full speed, but you just can't run every game at full speed all the time. Just something to keep in mind for the future.

Anyway, this is about changing the default plugin. :smile: We all obviously think that the plugin should be changed to anything but Jabo's (unless he releases the source or something), and Glide64 and GlideN64 are pretty much the best options. So let's merge this so we can start more work on the RDB and the core. :wink:

project64:
I did the changes locally about a month ago ..  after I get the android version to a basic working version, I am happy to come back to this .. there might be some debug option to be able to detect it and work out what games need a setting .. it might be I add a flag for extra checks on reading/writing to memory or certain addresses. It might be like protected memory where I change buffer ranges to no access and have exception when you read/write to a buffer generated from the gfx plugin .. 

ok how many games are we talking about that need a hack in for some type of cfb, 5, 10 , 100 .... 

there is probably about 5-10 that really need a hack instead of a generic solution .. the ones I can think of that are the top tier off the top of my head are (mario kart,kazooie, tooie, perfect dark, zelda), second tier (1080, pokemon stadium, pokemon snap,dr mario, mario tenis)

out of those 10 are there any else that it is hugely needed for ..

LegendOfDragoon:
>@AmbientMalice I got as far as showing that using function based look up was to big of a performance hit to justify using it ...

Then it's not worth it imo. Glide64 is slow enough with frame buffer emulation disabled. No way would that be good to slow down the CPU core by 20% for some feature that isn't always needed. I greatly appreciate the fact that you care about performance :smile: . I'd actually like to see PJ64 get faster too. It's something I plan on looking into.

>out of those 10 are there any else that it is hugely needed for ..

Maybe Vigilante 8 as well. Although I'm not sure if that game even runs well on any video plugin.

weinerschnitzel:
FBInfo API extension can be completely ignored. I think it is possible to get things working accurately at acceptable speeds and this should be a priority feature. I like hacks, but I think they should be created second.

AmbientMalice:
>Then it's not worth it imo. Glide64 is slow enough with frame buffer emulation disabled.

Bearing in mind that you seem to have a serious hardware incompatibility with the plugin.

>No way would that be good to slow down the CPU core by 20% for some feature that isn't always needed.

The range of N64 games that read the framebuffer in RAM is wide. It's all over the place. You seem to be under the impression this is some nice little feature PJ64 can do without. It's not. If PJ64 can't fix this problem in the near future, it will die. And it's just one half of the puzzle. N64 emulators also need to be able to detect when the CPU has modified the framebuffer in RAM and then copy those modified pixels onscreen. Otherwise we'll never get stuff like Pokemon Snap rendering correctly. Granted, there may be fast hardware hacky workarounds for some of this stuff. But we should be moving as far away from any approach that won't work globally as possible.

>FBInfo API extension can be completely ignored. I think it is possible to get things working accurately at acceptable speeds and this should be a priority feature.

What do you mean by this? FBInfo was never part of the Zilmar spec, right? So what are you thinking in terms of actually achieving this?

project64:
if we ran in a native res .. then we could easily do none hack ways around this and I would be all for it .. but as most of the time we are not running in native method and the way the n64 works with the display memory being inside the normal apps memory is so not the way systems work these days .. it probably always need hacks opposed to a nice generic solution.

AmbientMalice:
@project64 What do you mean by hacks? Hardware framebuffering is a hack. It's always been a hack, and a very problematic one. But we need it because software rendering is far, far too slow.

Video plugins try to accurately detect the framebuffer size, and the framebuffer gets copied into the RDRAM. This can potentially cause crashes. But it can't be helped. The only difference with FB notifications is that the framebuffer copies are done in chunks at the emulator's request instead of expensive fullscreen copies 30-60 times a frame. In Mario Kart's case, 1/6 of the framebuffer gets copied each frame, and updated ingame. It's up to the video plugin to handle the framebuffer size detection and such.

LegendOfDragoon:
>Bearing in mind that you seem to have a serious hardware incompatibility with the plugin.

It's a problem that needs to be fixed in the plugin, considering I do not have that issue with any other HLE plugin.

>You seem to be under the impression this is some nice little feature PJ64 can do without.

Correct. 

>If PJ64 can't fix this problem in the near future, it will die.

You seem to be over exaggerating.

>it probably always need hacks opposed to a nice generic solution.

Agreed.

Lithium64:
@LegendOfDragoon As I said on the issue # 629 the problem of slowness in some systems must be due to bad drivers from AMD, which already has a history of terrible OpenGL drivers. I also have had several problems with OpenGL when I was using the Radeon HD 4650.

project64:
@Lithium64  so does that mean it is a problem with the plugin or all opengl plugins .. and is there anything that can be done .. like upgrade a driver like being able to detect certain drivers and ask the user to upgrade it.

LegendOfDragoon:
>@LegendOfDragoon As I said on the issue # 629 the problem of slowness in some systems must be due to bad drivers from AMD, which already has a history of terrible OpenGL drivers. I also have had several problems with OpenGL when I was using the Radeon HD 4650.

Glide64 is slower, even on good hardware. Also I'm not convinced it's caused by bad drivers in my case because this is literally the only slow HLE plugin. All the rest are fine.

AmbientMalice:
>You seem to be over exaggerating.

You're underestimating the severity of the problem. I'm sure everyone's dying to use the N64 emulator with the glorious choice between games--ranging from Bomberman 64 to Superman 64--being broken in various ways, or framerates that tank down to 50% of full speed on an i7. Meanwhile the cool kids over in the corner are running at full speed on neo-mupen64: Infinity Waifu Edition because many years ago someone realised that copying every frame to RAM was a serious performance hog. Sure, GLideN64 runs a lot faster than Glide64 (dramatically so, in fact) when copying every frame to RAM, but it's still far too slow for comfort. And it's also inaccurate behavior.

Lithium64:
@project64 The main problem is that AMD has already stopped support many hardware still widely used, even on those who still have support rarely they make some improvement on OpenGL drivers probably because the gaming market only uses DirectX on Windows.

@LegendOfDragoon Your GPU driver is updated? you put the amount of VRAM on your GPU manually in Glide64? because the video memory detection does not work on it.

purplemarshmallow:
>ok how many games are we talking about that need a hack in for some type of cfb, 5, 10 , 100 .... 

@project64 About 50 games use fbread effects. In many cases it certainly can't be hacked because the CPU does complex framebuffer manipulations. Keep in mind Glide64 can't scale the buffer on the GPU. This makes per-frame copying completely useless.

Who will hack 50 games? source: Glide64 compatibility list (and this list is incomplete)
https://github.com/gonetz/glidehqplusglitch64/blob/master/Glide64/Help/Glide64%20compatibility%20list.html

AmbientMalice:
@purplemarshmallow Then you're got the LLE-only games. I know Indiana Jones needs FB to RDRAM for its sepia pause menu. Without it, you get nothing but black. Oh, yea, and the game runs at 400x440, meaning it will has unusually high overheads compared to most 320x240 N64 games if copying every frame to RAM.

Framerate+FB resolution is a huge factor. Perfect Dark in High Resolution Mode+CF=1 on PJ64 is a huge strain on my PC, even with the otherwise very fast GLideN64. Granted, that game has some odd FB performance issues, but big framebuffers = big overhead. And some might say, "But you don't need to run in high resolution mode!" You do if you want higher quality mipmapping, because PD ties mipmap distance to resolution.

LegendOfDragoon:
>@LegendOfDragoon Your GPU driver is updated? you put the amount of VRAM of your GPU manually on Glide64? because the video memory detection doesn't work on it.

Last time I checked, it was up to date. Tried manually setting the vram, can't really see a difference. If the problem was my drivers / hardware, then more HLE plugins should be affected.

purplemarshmallow:
>If the problem was my drivers / hardware, then more HLE plugins should be affected.

What other OpenGL plugins work for you?

LegendOfDragoon:
>What other OpenGL plugins work for you?

Both glN64 and Rice Video's OpenGL have better performance.

purplemarshmallow:
I still think that Glide64 is the best choice as default plugin even if some old hardware is left unsupported. It offers the best compatibility and people with bad hardware can manually download other plugins. Glide64's system requirements are not high. But higher than Rice Video's system requirements that's true.

weinerschnitzel:
@project64, would you consider committing you're current work on FBInfo extensions to a separate branch? It would be nice if we could experiment before you look at this again after you finish the android port.

cxd4:
I don't really understand this argument (or the last 50 or so new posts while I was knocked off).

As far as deciding between per-game hacks and FBInfo.  This whole FBInfo thing **itself** is essentially a bypass strategy--a hack even--around having the graphics plugin routinely read and write frame buffer data every time.  It's a bypass for speed.  If any games constantly changed the FB info then this FBInfo feature would become wasteful and obsolete, so why not just try to optimize the worst-case scenario first instead of hacks that aren't guaranteed to be any faster?

purplemarshmallow:
>If any games were to constantly read the FB info then this FBInfo feature would become wasteful and obsolete

why wasteful? If a game reads it then it should be emulated.

>so why not just try to optimize the worst-case scenario first instead of hacks that aren't guaranteed to be any faster?

Glide64's read every frame is the worst case scenario. How can this option be more optimized?

In most cases emulator notifications will work with no slowdown at all because the CPU reads only a single frame for pause menu backgrounds and similar things. It's tested with Mupen and 1964.

LegendOfDragoon:
With Glide64, I barely get ~70 VI/s during the part where jumbotron is displayed. If the CPU emulation gets slower by 20% that could be a problem. So I have to agree with cxd4 about optimizing the worst case scenario.

cxd4:
>why wasteful?

Because in the case I described, there is no reason for the CPU to make extra calls to signify (to the graphics plugin) reads/writes in conjunction with the FB done by the CPU, if it's constantly being done anyway.  Because if it's constantly being done every frame or something, then the CPU notifying the VI becomes extraneous operations and the graphics plugin can just behave accurately, without hacks, and routinely update frame buffer pixels regardless of when the core emulator suggests the need to synchronise (if some hypothetical case in some ROM has this constantly happening all the time anyway).

That's what I'm referring to when I describe optimizing the "worst-case scenario", as far as this goes.

purplemarshmallow:
>With Glide64, I barely get ~70 VI/s during the part where jumbotron is displayed. If the CPU emulation gets slower by 20% that could be a problem. So I have to agree with cxd4 about optimizing the worst case scenario.

I guess for Glide64 it must be tunred off by default unfortunately for Mario Kart it can't help in this case. The Glide api does not allow scaling the buffer on the GPU (not that I'm aware of). With modern plugins it can be optimized. Mario Kart will easily run fullspeed.

Someone could try to implement partial buffer copies into Glide64 that would speed things up (with emulator notification) but I'm not sure if it's worth to put more work into this old plugin.

>Because in the case I described, there is no reason for the CPU to make extra calls to signify (to the graphics plugin) reads/writes in conjunction with the FB done by the CPU, if it's constantly being done anyway. Because if it's constantly being done every frame or something, then the CPU notifying the VI becomes extraneous operations and the graphics plugin can just behave accurately, without hacks, and routinely update frame buffer pixels regardless of when the core emulator suggests the need to synchronise (if some hypothetical case in some ROM has this constantly happening all the time anyway).

It's not clear when the GFX plugin must write to RDRAM and how often. Doing it once per frame is not enough it will break some games. Doing it more often will kill performance. Doing it when FBread is called would be ideal.

cxd4:
>Doing it once per frame is not enough it will break some games.

Because it's a hack.  I said "every frame" for the purpose of example, but it's still pretty frequent.

Think about it:  Most of your reasoning for FBread's usefulness revolves around the fact that you can save performance by not constantly emulating frame buffer reads and writes like the actual hardware does, but if such a case were to arise in that a ROM does constantly need to read and write through the CPU, then you save no performance by checking first whether to bother emulating it (or skip emulating it for the hack) because the checks are a waste of time if it's constantly ongoing.

purplemarshmallow:
Yes in general it will be much faster but It can be slower in cases. And the performance gain comes from the fact that games do not read very often mostly only a single frame for a pause menu. But there can be a game where it's faster to write back without notification.

LegendOfDragoon:
>Someone could try to implement partial buffer copies into Glide64 that would speed things up (with emulator notification) but I'm not sure if it's worth to put more work into this old plugin.

I don't think it is worth doing. It would be better to fix GLideN64's fundamental issues and have that be the new default video plugin.

Lithium64:
Advocates of focus on performance will criticize me, but now the focus ought be on proper emulation of games because even the dolphin (HatCat will punch me) has a better emulation of various Nintendo 64 games. The optimization ought come after reach a better emulation of the games.

AmbientMalice:
Dolphin's compatibility gets exaggerated. Anyhow, my view is that PJ64 needs a fast, accurate-enough hardware plugin. HLE was always a bad idea for the N64. This is the whole awful problem. Suppose GLideN64 gets fixed enough to match Glide64 in HLE. Well, it still won't be good enough unless someone magics up every missing and broken ucode.

AmbientMalice:
>Even if the emulator became much heavier I would like to see a proper HLE emulation of games like Star Wars Rogue Squadron and others Factor 5 games.

I seriously doubt this will ever happen. We can't even get properly working HLE for Acclaim games. (Turok/Armorines/South Park.) Factor 5's stuff is so much more complex.

LegendOfDragoon:
>Advocates of focus on performance will criticize me, but now the focus ought be on proper emulation of games because even the dolphin (HatCat will punch me) has a better emulation of various Nintendo 64 games. The optimization ought come after reach a better emulation of the games.

Sounds like you've never heard of angrylion's plugin :smile: .

vgturtle127:
@AmbientMalice Wasn't Acclaim's engines and some other N64 stuff leaked a while ago?

Lithium64:
@LegendOfDragoon Angrylion plugin is LLE

purplemarshmallow:
Why sould someone want HLE if LLE can be fast as well?

cxd4:
Assuming LLE was even faster than HLE, I still think there may be a cool thing or two about HLE.  For one thing you see the system development for the platform in the same way the makers of the games did, which can be nostalgic for people favoriting the analysis of certain games.  HLE also sometimes gives you access to more features than LLE does, so far as exploiting the emulator's ability to not be the actual system goes.

Though I agree, I would just use LLE every day pretty much.

Lithium64:
@purplemarshmallow I never see a LLE fast as well and these plugins does not have enhancements like the option to Increase the internal resolution.

LegendOfDragoon:
>@LegendOfDragoon Angrylion plugin is LLE

But you said "the focus ought be on proper emulation of games". Right now, Angrylion's plugin is very accurate.

>Why sould someone want HLE if LLE can be fast as well?

HLE will still be useful for extra performance. If I want to play Top Gear Rally, LLE won't be fast enough for good audio. But in many cases, LLE can be good enough. I use LLE for quite a few games such as Last Legion (no real choice, but it's good enough), SD Hiryuu, Kirby 64, etc.

AmbientMalice:
>I never see a LLE fast as well and these plugins doesn't have enhancements like a option to Increase the internal resolution.

z64gl supports higher resolutions. You're confusing LLE with software rendering. z64gl runs everything from Mario 64 to Indiana Jones at full speed-to-significantly more than full speed. With higher resolutions. Even GLideN64 can run some games well in LLE, despite being broken.

Lithium64:
@AmbientMalice I never tested GlideN64 on LLE mode, but z64 I already tested and this was worst than Rice's video plugin in terms of accurate emulation.

cxd4:
Model exporting and replacement, audio/sound track substitution, 3-D redesign or textured polygon strip hi-res, porting from functional high-level MIPS to native Intel executables ... stuff like that is why HLE is still merited even if LLE were faster than HLE.  Basically every advantage having the source to the games gives you, except it's logically unrelated to emulation because emulation really should be LLE.

>HLE will still be useful for extra performance. If I want to play Top Gear Rally, LLE won't be fast enough for good audio.

"Extra" performance I cannot imagine as one of the benefits.

What you're observing as good or bad audio is based on the way that current emulator cores thread the AI and weigh it in with the rest of the RCP.  The real N64 basically has no audio crackling or skipping, even if heavy graphics slow down the actual console.

vgturtle127:
@cxd4 And if you are hearing cracking or popping you should check your audio plugin since that is a somewhat popular feature with audio plugins, I have noticed. :wink: 

Especially Azimer's. but I have not tested as extensively as I would like.

tony971:
This seems to be the wrong ticket for all of these comments. The issue at hand is whether or not to make Glide64 the default plugin or not. Everyone, except maybe @LegendOfDragoon , thinks that this is a good idea. In the future, z64gl, angrylion, and/or GLideN64 will have their day. But not before 2.3.

LegendOfDragoon:
>What you're observing as good or bad audio is based on the way that current emulator cores thread the AI and weigh it in with the rest of the RCP. The real N64 basically has no audio crackling or skipping, even if heavy graphics slow down the actual console.

I did read about that somewhere. I do not expect this to be implemented in emulators any time soon.
I think better performance is always nice have.

>Everyone, except maybe @LegendOfDragoon , thinks that this is a good idea.

If that were true, then it would have been the default already.

vgturtle127:
As long as the default isn't Jabo's (just so we can move forward with an open-source plugin), it could be Jerry's Null plugin for all I care. :laughing:

tony971:
Leave out future speculation. The development of these plugins will be after the 2.3 release. Here's what we have right now:

* Jabo'sD3D - Fastest but missing crucial features. Best for potatoes.
* Glide64 - In all likelihood, will never be seriously developed again. Highest compatibility without outrageous spec requirements. Dead in the water once it's surpassed.
* GLideN64 - The future of HLE. What we all wish was good enough to ship right now but isn't. And even if it were, the minimum requirements would prohibit defaulting right now.
* Angrylion - Makes every computer feel like a potato. Should be the basis for faster plugins.
* z64gl - The future of LLE. Can port angrylion's features fairly easily. Won't be fully developed until after 2.3.

Given these current choices, Glide64 needs to be the default for 2.3, but definitely not in the future.

AmbientMalice:
@purplemarshmallow How deeply broken is GLideN64's LLE, do you think? Could it be salvaged? It has no fog, huge slowdown+glitching similar to old z64gl, missing 2D elements, and non-working depth buffer. (Plus the many FB issues shared by LLE & HLE.) Other than this, it's pretty good, as stupid as that sounds. The problem is that GLideN64 LLE has pretty much gotten no improvement over the past year or so.

Lithium64:
@tony971 In my opinion Jabo D3D would have been removed long ago, because it uses the DirectX 8 which is an extremely outdated API, and moreover it is closed source. Including Glide64 has been placed as the default plugin before, don't know why zilmar returned to Jabo as default later.

Lithium64:
If nGlide64 wrapper could be bundled with Glide64 for PJ64, it could solve the issues with the bad video drivers and perhaps the speed issues.

vgturtle127:
@Lithium64 I agree. With the nGlide wrapper, there are way more systems it works on. Not only that, but nGlide is probably the best wrapper out there.

purplemarshmallow:
>@purplemarshmallow How deeply broken is GLideN64's LLE, do you think? Could it be salvaged? It has no fog, huge slowdown+glitching similar to old z64gl, missing 2D elements, and non-working depth buffer. (Plus the many FB issues shared by LLE & HLE.) Other than this, it's pretty good, as stupid as that sounds. The problem is that GLideN64 LLE has pretty much gotten no improvement over the past year or so.

GLideN64 LLE is many times slower than z64gl I think that's the biggest problem. The difference in speed is really huge. Maybe concepts from z64gl can help to speed it up.

AmbientMalice:
>GLideN64 LLE is many times slower than z64gl I think that's the biggest problem.

Do you have any idea why the plugin is so slow? It reminds me strongly of z64gl and Jabo's, especially for Turok. Your fork of z64gl runs perfectly fine. Old z64gl, Jabo's, and GLideN64 all have the exact same geometry glitching and huge slowdowns when you look in specific directions. I imagine that GLideN64's LLE graphics tearing and such are linked to its performance problems, just as they seem to have been with z64gl.

tony971:
@AmbientMalice it sounds like what you want is for someone to complete z64gl into a playable version of angrylion's plugin and then HLE the known ucodes. Which sounds wonderful... if someone actually puts in the time to do it.

purplemarshmallow:
I'm not really sure why my version of z64gl is faster than the original z64gl. Maybe it's because of the compiler switch.

LegendOfDragoon:
>@Lithium64 I agree. With the nGlide wrapper, there are way more systems it works on. Not only that, but nGlide is probably the best wrapper out there.

Using a different wrapper is a lousy workaround, assuming it even improves the performance. There is something wrong with the code. Blaming anything else is will get you nowhere. Glide64 will likely never see real progress. GLideN64 is much more promising. It is faster than Glide64 in many games, and also has many neat features.

project64:
glide64 for project has a glide wrapper there, if nGlide wrapper is better then happy to replace the current wrapper there .. or just individual functions, is there a like to nGlide wrapper?

LegendOfDragoon:
I heard nglide wasn't as good, but here's a link http://www.zeus-software.com/downloads/nglide .
I don't think it's open source though.

Lithium64:
@LegendOfDragoon It was closed source, but imo is better they added all extensions needed for Glide64 and DirectX 9 drivers are better than OpenGL especially on older hardware.

oddMLan:
Yeah, nGlide is DirectX9, but its not open source.

cxd4:
>and DirectX 9 drivers are better than OpenGL especially on older hardware.

Actually it's the other way around.
In the beginnings of DirectX, the API behind Direct3D was monstrously different from OpenGL (at which point, back then there were a lot more critics of D3D for game design than in today's generation).  Each new released version of Direct3D gradually assimilated more and more of the modern features already supported by GL extensions.

So if you're going to repeatedly raise the assertion that Windows Direct3D drivers are better than Windows OpenGL drivers _irrespective of who the video hardware manufacturer is_ then you'll need a logical pattern to establish that generalization as fact.  Otherwise, it's obvious that you're stating this observation in terms of what cards and drivers you've tested--not what others have tested, not a fundamental property that makes one API perform worse than the other naturally.

Lithium64:
@cxd4 Some old cards has good directx 9 support and not OpenGL 2.0 support at all or bad implementations of this API. This happens because the PC gamer market still today has his focus on Windows and DirectX where is the major market demand. The Nvidia is the only one gpu manufacter who has good OpenGL support, AMD and Intel doesn't care about OpenGL.

This is isn't about my experiences or opinion, but gpu market reality.

cxd4:
"Some" is the key word.

You can read whatever about how Microsoft is buying people's pockets into putting 90% of their testing into DirectX drivers instead, but in spite of that being true or false the "quality" of the implementation is only judged by the issues you can measure.  You can't report a bug you can't see, and you can't judge performance comparisons in terms of code you didn't understand, let alone didn't write.

Lithium64:
@cxd4 I'm working with computer maintenance for over 10 years, so I've used countless hardware and know the reality of the market, beyond what I speak the vast majority of graphics developers speak the same as me. Also just look AMD and Intel forums that you will see the huge amount of complaints coming from its users and developers about the OpenGL drivers. If the drivers were good enough developers wouldn't need to use hacks into their software to circumvent the bad implementations.

cxd4:
>I'm working with computer maintenance for over 10 years,

Essentially anybody who has owned a computer for 10 years has worked with computer maintenance for 10 years.  It says nothing more or less than developer testing of the two platforms.

The countlessness to the amount of video hardware you've tested also establishes nothing conclusive because it still depends how the application is written to draw the primitives.  Until you've studied that, all you can do is test opposite applications using opposite rendering engines, different functions and different design, just to observe on your end whatever looks faster or less buggy.  You have too many experimental variables in that equation to really know much of anything.

And for what it's worth, since we now are on the subject of hearsay and public complaints.

I've seen my own share of articles acknowledging the "network effect" Microsoft has over OpenGL being drowned out by Direct3D while still asserting that OpenGL is faster than DirectX on Windows.

>beyond what I speak the vast majority of graphics developers speak the same as me.

See above.  I've seen several articles and blogs claiming the exact opposite as you.
They still insist GL is faster than DX even in spite of the network effect Microsoft has for gaming.

>If the drivers were good enough developers wouldn't need to use hacks into their software to circumvent the bad implementations.

And yet, you subsequently admitted that other driver manufacturers like NVIDIA can be different from Intel's older flaky drivers.  Intel wasn't even acknowledged like AMD was back then.

So rather than mutating your conclusion of "AMD's and Intel's [older] OpenGL drivers are bad" to "OpenGL is worse" you might want to represent your understanding of the difference between hardware and software better.

Lithium64:
@First of all, I have access to computers since 1996 and was born in 1988 and work on maintenance since 2005 when I specialized in the area.

I'm not saying that OpenGL is bad, but the AMD and Intel drivers are, which happens due to lack of commercial interest of these companies. OpenGL when properly implemented is much faster and better than the DirectX 11.x or earlier.

While Windows dominate the operating system market will be difficult that some company beyond Nvidia (who leads the GPU market by a wide margin) be interested on OpenGL. Also from this year onwards the focus of game developers will be in the low-level APIs due to the release of DirectX 12 and Vulkan which will be released soon.

cxd4:
The real problem with OpenGL is that it's poorly documented and scarcely supported in the wide social spaces.  DirectX has very rampant documentation and exposes itself much more flexibly to learners.

Another major issue with OpenGL is that Khronos imposed their personal opinions on whether it should be considered a "low-level" or a high-level API when they zealously voted in their deprecation of convenient, small- and clean-code functions from the API to better match their OpenGL ES variant of desktop OpenGL.  Now AMD, unlike nvidia, sided with this decision, but at the same time they've seen insufficient benefit to actually remove any of the deprecated OpenGL functions from their drivers.  In fact, **there is no benefit**.  The deprecation was a stupid imposition of personal philosophy.

But to say that OpenGL drivers are worse than DirectX drivers based on the opinions of those affected by these sorts of decisions (plus what you said about the gaming network circle that Microsoft controls) when really at least newer drivers, Mesa3D (which is open-source), NVIDiA and so on have just fine implementations of GL is really just a broad generality.

LegendOfDragoon:
>A polished z64gl would be an excellent option, assuming the bug that causes up-and-down juddering with PJ64 gets fixed.

What game does this happen in? I'd like to fix it, if there are still issues related to that.

oddMLan:
This should be a milestone for Project64 2.4 since the dependancy on SDL and the hugely bloated WxWidgets has been recently removed 68a8ed03696f3365430d8cca14d5084100165cd0 d0f8bfe763f70baf06ca924ec2659c752fb2a95e  #1023

LegendOfDragoon:
>This should be a milestone for Project64 2.4 since the dependancy on SDL and the hugely bloated WxWidgets has been recently removed

I think Glide64 needs some serious improvement. Its framebuffer emulation is so slow that I get better performance with z64gl in mario kart when the monitor is displayed..

purplemarshmallow:
>I think Glide64 needs some serious improvement. Its framebuffer emulation is so slow that I get better performance with z64gl in mario kart when the monitor is displayed.

z64gl supports emulator notifications and copying to RDRAM but it looks a bit buggy. Is it really faster?

oddMLan:
Doesn't Glide64 already support Emulator notifications too? Could it be improved?

Although at this moment it would only benefit Mupen64/Plus and 1964 since... we are still waiting for support from Project64's side #72 #811 
coupled with #574 it would definitely be an immense help in speeding up the Frame buffer

LegendOfDragoon:
>Is it really faster?

At least on my end, it's way faster. I get like 70 VI/s with Glide64 final when the monitor is displayed and ~140 VI/s with z64gl.. I could leave fb emulation on the entire time and still have completely satisfactory performance in Mario Kart with z64gl. That's how big of a difference it is in speed. Using glReadPixels is much faster than doing generic memory copying.

I tried to speed up Rice's fb emulation, but could not find a D3D alternative for glReadPixels, so the best thing I could do was use SSE (although this is really only beneficial on a resolution like 640 x 480 or 320 x 240). I'd be greatful if anyone pointed out a decent alternative for D3D8 or D3D9.

>Doesn't Glide64 already support Emulator notifications too?

Yes, but it's still slow whenever it's doing fb emulation.

>Could it be improved?

Big time.

cxd4:
https://msdn.microsoft.com/en-us/library/windows/desktop/bb172231%28v=vs.85%29.aspx

glReadPixels() is not a 3-D operation.  It's hardly even an OpenGL command; it's more of a system memory bus gateway to read frame buffer data from the GPU to your CPU, which is bound to be slow.

Same can be said of Direct3D.  This is computer science here, not API wars of which API is better.

fallaha56:
'sigh' everyone here fighting and not able to see the forest for the trees

too many plugins and too low a hardware standard (D3D8(!)) will keep holding PJ64 back...JM2C

PR #890, opened 2016-01-17
Fix Castlevania: LoD missing dissolves.

AmbientMalice:
For whatever reason, while old_style_adither was enabled for Castlevania, it was not enabled for the sequel. I'm not aware of any negative effects.

With hack:
glide64_castlevania2_01

Without hack:
glide64_castlevania2_02


Lithium64:
LGTM

PR #891, opened 2016-01-17
Disable texture filtering hotkey.

AmbientMalice:
IMO, it causes too much trouble from accidental presses. Also, some games have their texture filtering mode set a certain way for a reason. An alternative would be to change the hotkey to something like ALT+Backspace.

This PR is very much personal preference, though.

### DIFF
--- a/Source/Glide64/Main.cpp
+++ b/Source/Glide64/Main.cpp
@@ -2038,14 +2038,14 @@ void newSwapBuffers()
       }
     }
     //hotkeys
-    if (CheckKeyPressed(G64_VK_BACK, 0x0001))
-    {
-      hotkey_info.hk_filtering = 100;
-      if (settings.filtering < 2)
-        settings.filtering++;
-      else
-        settings.filtering = 0;
-    }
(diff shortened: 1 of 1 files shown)


Source/Glide64/Main.cpp
```diff
@@ -2038,14 +2038,14 @@ void newSwapBuffers()
...
-        settings.filtering = 0;
-    }
+    //if (CheckKeyPressed(G64_VK_BACK, 0x0001))
+    //{
+      //hotkey_info.hk_filtering = 100;
+      //if (settings.filtering < 2)
+        //settings.filtering++;
+      //else
+        //settings.filtering = 0;
+    //}
```
cxd4: Sometimes you should try `#if 0`, because instead of deleting the 8 `//` comment-out tokens you added, you would only need to change the #if 0 to a #if 1 or whatever instead of changing 8 lines to test the code out quickly.

AmbientMalice: Thanks for the tip.

PR #892, opened 2016-01-17
Change Mario 64 filtering to automatic.

AmbientMalice:
I have no idea why it was changed to Force Bilinear to begin with. None of Glide64's texture filtering options are fully accurate, but Automatic is closest. The text is most certainly not supposed to be filtered the way it is currently.

Angrylion's:
nsme0000

Bilinear:
glide64_super_mario_64_01

Automatic:
glide64_super_mario_64_02


Lithium64:
LGTM

PR #901, opened 2016-01-18
Retail 64DD IPL ROM support (boots)

LuigiBlood:
For now only works when you load retail 64DD IPL ROM like a normal ROM.
I prepared g_DDRom for future 64DD emulation, as it always needs to be mapped at 0x06000000.

### DIFF
--- a/Source/Project64-core/N64System/Mips/MemoryVirtualMem.cpp
+++ b/Source/Project64-core/N64System/Mips/MemoryVirtualMem.cpp
@@ -194,6 +194,33 @@ bool CMipsMemoryVM::Initialize()
         m_Rom = g_Rom->GetRomAddress();
         m_RomSize = g_Rom->GetRomSize();
     }
+
+    //64DD IPL
+    if (g_DDRom != NULL)
+    {
+        if (g_Settings->LoadBool(Game_LoadRomToMemory))
+        {
+            m_DDRomMapped = true;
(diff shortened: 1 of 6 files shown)


Source/Project64-core/N64System/Mips/MemoryVirtualMem.cpp
```diff
@@ -194,6 +194,33 @@ bool CMipsMemoryVM::Initialize()
...
+            m_DDRomMapped = true;
+            m_DDRom = m_RDRAM + 0x06000000;
+            m_DDRomSize = g_DDRom->GetRomSize();
+            if (CommitMemory(m_DDRom, g_DDRom->GetRomSize(), MEM_READWRITE) == NULL)
+            {
+                WriteTrace(TraceN64System, TraceError, "Failed to Allocate Rom (Size: 0x%X)", g_DDRom->GetRomSize());
+                FreeMemory();
+                return false;
+            }
+            memcpy(m_DDRom, g_DDRom->GetRomAddress(), g_DDRom->GetRomSize());
```
project64: is the 64dd all read only?

LuigiBlood: 64DD IPL ROM is all read only. Actual disk drive emulation requires buffers to be emulated for read and write.

PR #904, opened 2016-01-19
Fix Waialae Country Club

AmbientMalice:
Not sure why framebuffer was disabled. Hope it wasn't for a reason. Game works better now.


Lithium64:
fb_smart is default, then turn on it specifically for these games is unnecessary

PR #910, opened 2016-01-19
Twisted Edge Audio Hack (Cheat Ver)

AmbientMalice:
As requested. Kudos to @LegendOfDragoon and @exhalatio


exhalatio:
USA and Japan works fine.
Unfortunately, Europe does not appear to cheat list.
You must be changed Cheat1 to Cheat0.

AmbientMalice:
Should work now. I forgot cheat lists start at 0, not 1.

PR #912, opened 2016-01-19
Glide64: Make fb_render true by default

Lithium64:
On Glide64 Final this is the default, this option enabled by default can fix some regressions.

### DIFF
--- a/Source/Glide64/Main.cpp
+++ b/Source/Glide64/Main.cpp
@@ -1647,7 +1647,7 @@ void CALL PluginLoaded(void)
     game_setting(Set_read_back_to_screen, "read_back_to_screen", 0);
     game_setting(Set_detect_cpu_write, "detect_cpu_write", 0);
     game_setting(Set_fb_get_info, "fb_get_info", 0);
-    game_setting(Set_fb_render, "fb_render", 0);
+    game_setting(Set_fb_render, "fb_render", 1);
 }
 
 /******************************************************************


AmbientMalice:
fb_render is software depth buffering, and it was disabled because it has a moderate performance overhead and a surprisingly small number of games actually require any degree of accurate depth buffer emulation.

project64:
so should it be enabled by default or just for the games that need it, and if it is just for the games that need it @Lithium64 which ones are missing it currently.

AmbientMalice:
It was originally disabled because PJ64's Glide64 had a crash bug. It was fixed, and I made fb_render enabled globally. But then in anticipation of Glide64 being the default plugin, some time ago I made it per-game to reduce CPU overheads, especially on underpowered CPUs. (The performance hit is not huge, but it is significant.) This is especially relevant now, since PJ64 will likely have framebuffer notification support soon, adding another CPU overhead which could push old hardware past breaking point. (But nowhere near as bad as per-frame framebuffering to RAM, which is painfully slow on most hardware.)

It was caution on my part. I was just wary of people being upset with Glide64's performance compared to Jabo's.

Lithium64:
@AmbientMalice Should be done some benchmarks to know the real impact of this change, here I do not feel any impact even on an ancient hardware.

AmbientMalice:
Framerates are hard to pin down because they fluctuate so wildly, but I'm ballparking around 170 with software depth and 190 without in Banjo Tooie. That's a difference of 11% on a bad day, but the numbers are erratic.

LegendOfDragoon:
How many games actually benefit from this option? If not too many, then I see no reason to have it on by default.

>I do not feel any impact even on an ancient hardware.

It's not easy to predict how well things will run on older hardware. Gpu performance is practically impossible to estimate. My hardware isn't ancient, yet I get sub-par performance in some games, without using software depth buffer or frame buffer. It can't entirely be my hardware's fault if I get better performance with LLE plugins for some of those games.

AmbientMalice:
@LegendOfDragoon software depth buffering is very much the "it fixes Resident Evil 2's prebaked depth buffers, bless the gods" setting. As a side effect, it also fixes coronas and lens flares showing through walls in a handful of games. Most games are fine without it because they don't require that degree of accuracy. Whether a game needs it can only be revealed by testing.

As far as I know, there is zero GPU impact, only CPU, since it's rendering the depth buffer without the GPU.

cxd4:
Even if `fb_render=1` becomes the default setting (which maybe it should be, depends on your own personal philosophy really), I don't think that justifies removing per-ROM local copies of `fb_render=1`.  Because this is necessary documentation to show games that really do need the feature, and it makes the INI flexible to either default option set = 0 or = 1.

Lithium64:
@cxd4 I did a rebase and removed INI changes

AmbientMalice:
The basic problem is that ~10% performance drop. Is a 10% performance drop globally enough to justify fixing a few random games with lens flares that haven't been manually added yet? I would say yes qf I could be sure that framebuffer notification support in PJ64 won't eat up too much performance. Zilmar currently estimates 20%.

Looking back, one of the reasons I made software depth buffering per-game was I was hoping it would leave more resources for per-frame framebuffering, which didn't really work out.

cxd4:
Having it enabled by default does have the advantage of being neutral to unknown or undiscovered ROMs, though, so I think it could go either way.  Maybe it should be = 0 by default; maybe it should be = 1 by default with setting it = 0 for speed in the games that need the extra speed.

tony971:
Personally, I'd rather see this default to 1 with RDB settings of 0 for games known not to use it.

AmbientMalice:
>Personally, I'd rather see this default to 1 with RDB settings of 0 for games known not to use it.

Which is almost all of them, thus requiring hundreds of entries of fb_render=0 That's a level of RDB clutter that seems rather excessive. At least to me.

cxd4:
It's also difficult to prove a negative.

Proving that a game DOES need fb_render=true is easy just by encountering it.
Proving that a game NEVER needs it requires more extensive testing...plus maybe GameShark cheats or beta features or whatever could behave differently with or without it emulated, idk.

So in principle default to =1 is accurate, but it's a good thing the user can change it to =0 if they need the extra speed as well.  Setting per-ROM = 0 usually should be done carefully.

PR #915, opened 2016-01-20
add missing romhacks, fix quest 64 (glide64)

AmbientMalice:


cxd4:
_Quest 64_ with point-sampled filtering? :P 

Doesn't look too different from the GameBoy release that the n64 port copy-pasted off of I guess, haha.

oddMLan:
what does the "useless_is_useless" option do?

AmbientMalice:
>what does the "useless_is_useless" option do?

No clue. I wish I knew what half the Glide64 hacks actually did, but nobody seems to have bothered documenting their exact function. It's a framebuffer related hack, though.

Yui12:
I find a way to fix the text in Quest 64, but it requires the use of the latest version of glN64.

In the graphics settings, Make sure fullscreen mode is set to 32bit, Check: Enable 2xSaI texture scaling, Uncheck: Force Bilinear Filtering, Enable Fog (for fog emulation since this game uses it), and Texture Bit Depth is 16-bti and 32-bit (normal)! Not only text will no longer be pixelated, and the graphics will look smoother, the text will look better. It seems that 2xSai seems to work better here than in Glide64 or GlideN64 for me as well.

The only down sides to this is that the text will not look as bold as on a real N64, but it will look a bit better than point sampled filtering from Glide64. Also when you scroll through the menus, you'll get a minor bug if you scroll pass the element status screen showing flashing numbers, but other than that, the game is fully playable to the end and through the credits with no other graphical issues. 

For some strange reason, you can do this with Direct64 which is the reincarnation of glN64.

PR #925, opened 2016-01-23
Fix regression in CRecompilerOps::SPECIAL_DADDU()

LegendOfDragoon:
It should only add carry if GetMipsRegLo(source2) != 0


dsx-:
does this fix any games?

LegendOfDragoon:
My previous commit made the recompiler more accurate for Fifa 64, but it broke Fushigi no Dungeon. So this commit fixes the crash caused by previous commit.

PR #933, opened 2016-01-25
[Glide64] poop :)

cxd4:
According to MSDN, a MessageBox whose title param is a null pointer defaults to the text "Error".
So instead of allocating extra bytes to pass a pointer to the string `"Error"`, why not pass NULL?

Actually the real reason why I made this PR was me having OCD against a cosmetic spacing inconsistency introduced in this commit:  https://github.com/project64/project64/commit/aefe8e603c04d543e2071174fab2f32e1c951476 .  So if you prefer "Error" over NULL I'll just undo that part of this PR and keep the line break change in instead. :P

### DIFF
--- a/Source/Glide64/rdp.cpp
+++ b/Source/Glide64/rdp.cpp
@@ -336,9 +336,9 @@ void microcheck()
 
         ReleaseGfx();
 #ifdef _WIN32
-        MessageBox(gfx.hWnd, stdstr_f("Error: uCode crc not found in INI, using currently selected uCode\n\n%08lx", uc_crc).c_str(), "Error", MB_OK | MB_ICONEXCLAMATION);
-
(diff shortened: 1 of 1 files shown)


project64:
I prefer "error" because at some point I want to get language translation at some point.

cxd4:
Works for me.  Simplified the PR down to the real reason why I started it. :D

vgturtle127:
@cxd4 :laughing:

cxd4:
It just was because zilmar committed something like
```
MessageBox(...
#endif
(blank line)
void function(blah ...)

[... several lines later ...]
MessageBox(...
(blank line)
#endif
void function(blah ...)
```
Which struck me into some OCD state of mind into making a PR for making the spacing consistent, though the NULL instead of "Error" thing was something I thought of alongside it.

cxd4:
Oh, and adding a line break at end of the file was not done by me.

GitHub online website editor put the line break at the end of the file for me; I never told it to do that.  But it's correct to have anyway nonetheless so i didn't care to attempt deleting it.

PR #937, opened 2016-01-26
64DD support

LuigiBlood:
No saving or swapping **yet**.

I tried to make it as seamless as possible to load a 64DD disk file.

I reused quite a lot of code from MAME and Happy_'s ddconvert tool.
I don't even know if the License allows for that.


oddMLan:
Holy smokes! Who'd expected N64DD support in the next version of PJ64? That's an amazing feat! And since the goal is to get a basic version running on Android, that also means we can play N64DD games in Android :D

LuigiBlood:
That will depend on zilmar. He has the final say for that.

cxd4:
You have an even bigger say.

Every 64DD-related pull request before this one you've done has been organized to its own small and readable PR.  I see nothing unique to this PR preventing the same cleanly organization from being done.

LuigiBlood:
Even though this pull request is pretty huge to what I've usually done, I still tried to do things as clean as possible for the most part, and describe the big changes to the commit name. (Doesn't always describe the small changes here and there though, but I tried to write code as clean as possible and portable, even if I can't really be sure of it?)

cxd4:
Kind of the idea.  You did the clean job of organizing it well enough to have types of changes under their own commit.  The confusion is the variety to the commits--as well divided as they are, are all different types of accomplishments, all multiple questions of whether it's safe to merge or not.

LuigiBlood:
Maybe I can seperate the 64DD support itself and the commits related to the user interface?

cxd4:
Short of every commit here looks like it could be its own pull request, but I wouldn't do it like that necessary.  Any commit that doesn't break the code until a subsequent commit depending upon it is made could be its own PR, but really the only important thing is just to have less changes to scrutinize.

Giggybytes:
Holy smoke! it works like a charm. For the first time i can play SM64DD and Mario Artist Studio

PR #940, opened 2016-01-27
Add .gitattributes

vgturtle127:
I am going to start doing all of the small pull requests I have piled up over the last few months. I updated to the latest source, and I want to make a few small pull requests that will hopefully make the repository and development better.

This pull request is just adding a global .gitattributes to the repository. This way, we don't have to worry about EOL, encoding, etc.

I think I will do pull requests one at a time, and only use separate branches for each pull request.

Any suggestions, comments, or questions  just feel free to ask as per the usual.

Hope this is OK, @cxd4 ! :smile:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,44 @@
+# Automatically detect text files and perform LF line ending normalization
+* text=auto
+
+# Files you want to always be normalized and converted
+# to native line endings on checkout
+*.c text
+*.cpp text
+*.h text
+*.in text
+*.dsp text
+*.iss text
+*.rc text
+*.cpf text
+*.sc text
+*.cmd text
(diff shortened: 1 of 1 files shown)


vgturtle127:
@project64 Thanks, Zilmar!

LegendOfDragoon:
What was the purpose of this PR? What problem(s) did it actually fix?

vgturtle127:
@LegendOfDragoon It normalizes the line endings to the users native line endings at checkout, and the only other thing it does is ensures binary contents aren't modified and makes sure source code and text files and such are normalized. This doesn't really "fix problems", it prevents them. Much like having a .gitignore file doesn't solve any problems, it just prevents them.

PR #946, opened 2016-01-28
Change up the way Mempaks are meant to work, use the standard file fu…

death-droid:
…nctions. Mempaks are now loaded in when the game launches, and file handles are only opened durring read and writes

For some reason, and for the life of me i cant work it out, either my GIT or Visual Studio is messing up the files so its changing every line. @cxd4 @project64 any idea what ive done wrong? XD


cxd4:
It's not your fault.  https://github.com/project64/project64/pull/940

death-droid:
Ahhh i see, cheers. Made multiple forks fiddling around with settings to see why. That explains it.

vgturtle127:
@cxd4 Is that commit causing problems?

cxd4:
I can't say.  Why not try actually contributing to Project64's actual code for once and you might see?

vgturtle127:
@cxd4 SHOTS FIRED.

Really, though, if it's causing issues, please tell me so I can fix it.

Besides, you don't want the whole emulator to be broken because you told me to contribute, do you? :smile:

cxd4:
No need to get emotional and turn threatening about it.  Proposing you try doing something useful here isn't the same thing as asking you to damage things any further:  It just means contribute.

But if you would really like to assist, maybe you can tell us what part of death-droid's changes in this pull request are actually changed:  https://github.com/project64/project64/pull/946/files ? Since you seem to not be bothered at all by the fact that the entire files are being highlighted as the change.

vgturtle127:
@cxd4 I wasn't threatening you, I was just kidding around. :blush: 

I want to contribute, I just think I can best help with the things nobody else wants to work on. I have a bunch of stuff locally I am working on, like a huge update to the cheat database, RDB fixes, and some more stuff related to GitHub workflow.

I will try to work on this PR first, though. I am bothered by the fact that the files are seeing changed line endings instead of actual work, I just don't usually have this kind of problem with .gitattributes...

cxd4:
Then you underestimate yourself.  If you really wanted to, I'm sure you could adopt your own abilities to improve the code...whatever, clean-ups, faster code functions, playing around with C math, spacing cosmetics....  But if you really feel specialized in RDB and Wiki stuff and whatever then that's not to blame I guess.

vgturtle127:
@cxd4 mempak.h before:

class Mempak
{
public:
    static void Close();
    static uint8_t CalculateCrc(uint8_t * DataToCrc);
    static void ReadFrom(int32_t Control, uint8_t * command);
    static void WriteTo(int32_t Control, uint8_t * command);
};

After:

class Mempak
{
public:
    static uint8_t CalculateCrc(uint8_t * DataToCrc);
    static void Load();
    static void Format(int32_t Control);
    static void ReadFrom(int32_t Control, uint8_t * command);
    static void WriteTo(int32_t Control, uint8_t * command);
};

So every line after public was changed. I will try to do the other files ASAP.

vgturtle127:
@cxd4 I suppose my lack of support for myself has to do with trying to contribute to various projects before and getting rejected or telling me they wanted me to do it all over again. I guess I assumed my code sucked, and that's why I don't contribute to projects with code as much anymore. :disappointed:

cxd4:
I'm not going to say you did the .gitattributes incorrectly.  (I sure as hell won't white lie and say your .gitignore made any sense though.)  Your .gitattributes wasn't bugged, but it fixed fewer hypothetical "issues" than the realistic problems it causes.

Now, thanks to your .gitattributes I did see a spontaneous Git listing of files stored as CRLF on Git's server-side, even when checking out on Linux (which is wrong and it should be LF), but I don't think fixing all line endings from CRLF to LF is worth making everyone's pull requests unreadable.  If it's that important I or somebody else can submit a PR ourselves to fix the line endings to be LF inside of Git.

PR #949, opened 2016-01-28
Don't force other contributors to make unreadable PRs.

cxd4:
Three commits in this PR--three reasons why the .gitattributes is not necessitated by anything.

Though the first commit is the only real important one.

Deleting the entire file is something I wanted to do, but I decided to leave the mostly harmless (but neutral and invisible) idea to treat binary files as binary files in there.  Best case scenario it prevents people from poking binary changes to Git repositories which should be text, so I left that in.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,12 +1,3 @@
-# Files that will always have CRLF line endings on checkout
-*.sln text eol=crlf
-*.vcproj text eol=crlf
-*.vcxproj text eol=crlf
-*.filters text eol=crlf
-*.props text eol=crlf
-*.vsprops text eol=crlf
-*.vbs text eol=crlf
-
 # All files that are binary and should not be normalized
 
 # Images
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,21 +1,3 @@
-# Automatically detect text files and perform LF line ending normalization
-* text=auto
-
-# Files you want to always be normalized and converted
-# to native line endings on checkout
-*.c text
-*.cpp text
-*.h text
-*.in text
-*.dsp text
-*.iss text
-*.rc text
-*.cpf text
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,30 +1,3 @@
...
-*.iss text
-*.rc text
-*.cpf text
-*.sc text
-*.cmd text
-*.sh text
-*.txt text
-
-# Files that will always have CRLF line endings on checkout
-*.sln text eol=crlf
```
JunielKatarn: Why remove this section?
Made sense to me.

cxd4: Referenced in the second commit.

There's no reason to force CRLF for Visual Studio files, which are flexible to LF anyway.

The Git SCM _almost always_ should store as LF.  When you check out to Windows, the Git *nix server converts these to CRLF for you automatically, so there really is nothing to gain from forcing the VC files to CRLF, especially if some program on Linux or whatever had a program that could use these seemingly Microsoft-specific files.  Nothing about them mandates a requirement for forcing CR-LF when Git for Windows is already automated to presuppose this.

project64:
well might have made sense, but did not seem to work great .. I did a commit (https://github.com/project64/project64/commit/a228a9d1f987fe2d3aed11335589892bd390a3ce) just trying to deal with line endings ...

git has been doing a good job of commit to the server as cr and checking out on windows as crlf

cxd4:
That was the other reason I was thinking to edit in--you still had to commit in those Visual Studio files.

So aside from multiplying the .git repository download size, it also resets GitHub's Blame debugging history (tracking which line change was last done by which user) because the whole file got replaced.  Not worth forcing the internal storage of LF unless honestly required by something.

Issue #948, opened 2016-01-28
Crash on USB flash drive


Lithium64:
Project64 crash if the user tries to run from a USB flash drive
![crash pendrive](https://cloud.githubusercontent.com/assets/10705488/12635110/961a466c-c55e-11e5-8d84-75987f9b34e8.png)

vgturtle127:
@Lithium64 Well, that's a woefully unhelpful exception. :disappointed: 

I mean, I don't even know how to try and fix this.

What flash drive is it?

Lithium64:
@vgturtle127 It isn't a flash drive fault, you could try run from a USB flash drive and see with your own eyes.

vgturtle127:
@Lithium64 Oh, I know. Are you using FAT32?

Lithium64:
@vgturtle127 yes

vgturtle127:
@Lithium64 How did you put the files on the flash drive? Did you just copy the files over from your hard drive?

Lithium64:
@vgturtle127 I just copied the folder to my flash drive

vgturtle127:
@Lithium64 Did you change all the directories and stuff to be relative for your flash drive? That may be an issue.

Lithium64:
@vgturtle127 This crash isn't related to INI files, because I deleted these files from config folder before copy the emulator folder to USB flash drive. The emulator doesn't quite open up, because the error occurs as soon as you try to open.

AmbientMalice:
This error isn't related to flash drives, as far as I can see. It's caused by placing PJ64 one folder level down from a drive root.

C:/PJ64 -- error.

C:/Spatula Wars/PJ64 --fine.

death-droid:
Fixed the issue, will push a pull request shortly.

tony971:
The placement of the configure folder has always confused me. Why not keep it in the main PJ64 directory or the user's documents or appdata folders?

project64:
@tony971  if you install pj64 at the default location then it is in the program files directory, which can not be directly written to and the data should be in the the user's documents or appdata folders as you said. That is the normal behaviour, if you install it elsewhere say "D:\pj64" then it does not need to appdata. 

I also wanted pj64 to be self contained, so if you have multiple versions in different directories changing the settings of one does not effect the other, which was one of the main reason of going to a config file over using the registry.

[...]

PR #952, opened 2016-01-28
Quick fix to the issue that was causing Project64 to crash when used …

death-droid:
…in a base directory such as C:\Project64

### DIFF
--- a/Source/Project64-core/AppInit.cpp
+++ b/Source/Project64-core/AppInit.cpp
@@ -37,7 +37,7 @@ void InitializeLog(void)
 
 void AddLogModule(void)
 {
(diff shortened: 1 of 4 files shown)


death-droid:
Should fix issue #948

project64:
no you change it from setting the path to the dir to the path to be the filename which is wrong.

I looked at why it was causing the crash it was in check directory exists, it was testing C:Project64 which failed. Create a fix for that here (
https://github.com/project64/project64/commit/4203cbd2efc94802e31a3c0fc50a594ea6c2316b) which should have the same effect

death-droid:
Ahh cheers for that, i thought i may of been fixong the issue by accident and not a properer solution to the root problem. Will checkout your commit.

PR #957, opened 2016-01-28
[Settings] InitializeSettings() can return bool.

cxd4:
Mostly to fix the repetitive errors when compiling Glide64 outside of Windows, which blames this.


cxd4:
I would fix that miserable typo as well, `Initilize`, which should be spell-checked to "initialize".  But if I fix the spelling I will probably be blamed as the build-breaker who breaks compilation because I don't run Windows.

PR #958, opened 2016-01-28
64DD IPL ROM Recompiler support

LuigiBlood:


Config/Project64.rdb
```diff
@@ -8113,8 +8113,7 @@ Status=Unsupported
 
 [00000000-00000000-C:4A]
 32bit=No
-Core Note=Recompiler isn't working properly
-CPU Type=Interpreter
+Fixed Audio=0
 Good Name=64DD IPL (JPN)
 RDRAM Size=8
-Status=Issues (core)
```
cxd4: There's a status for this called "Interpreter only", which would be correct for future reference.

Non-working recompiler doesn't qualify as a core issue with Project64; it just means it works interpreter-only.

PR #959, opened 2016-01-29
minor rdb edits

dsx-:


AmbientMalice:
Zilmar wanted the Twisted Edge fix to be an optional cheat to allow for further testing. Also, in the event it does get added to the RDB, the game should have " [audio] bad sound", since the game does, in fact, have bad sound, like most BOSS games.

dsx-:
why is github so fucking hard to use, cant figure out how to revert a commit

dsx-:
ok i think we're good

dsx-:
sorry for the mess but should be good now

AmbientMalice:
I think King Hill/Twisted Edge should be a Glide64 specific error, not video, because GLideN64 and z64gl and others don't suffer from the missing fog.

dsx-:
both those plugins arent included with project64 so thats irrelevant, [video] applies to all included plugins

dsx-:
unless the plugin notes column now applies to all plugins which means there would be a lot of rdb changes to make...

AmbientMalice:
>both those plugins arent included with project64 so thats irrelevant, [video] applies to all included plugins

I assumed it referred to video plugins in general as opposed to a problem caused by a specific plugin. Glide64 is the only plugin with the fog bug, at least from my recent testing. If the issue affects multiple plugins, then it should be [video], just as a bug affecting all HLE plugins should be specified as a [video] HLE issue.

>unless the plugin notes column now applies to all plugins which means there would be a lot of rdb changes to make...

When I overhauled the entries, it was oriented towards Glide64 being the default plugin. Since basically the only viable competitor to Glide64 is GLideN64, an issue marked as being a Glide64 issue is likely one fixed with GLideN64, or else Angrylion's. Jabo's compatibility isn't really irrelevant because it's broken with everything.

dsx-:
ah ok i treated [video] as an issue that affects both glide and jabo only (the included plugins). i will edit twisted edge

dsx-:
it sounds good to treat [video] as a general issue though rather than just confined to jabo and glide, so we're on the same page now

AmbientMalice:
IMO, the rom browser should be fairly simple. People can tell at a glance about major issues. The many, many small issues are the stuff best reserved for the Wiki. So, for example, if there are huge text alignment issues, that's worthy of mention in the rom browser entry. But if there's the typical slightly misaligned text with slight glitching along the bottom, then that's something better left for the Wiki.

Also, I think the rom browser should lean on the slightly negative side. It is absurd how Knife Edge: Nose Gunner was deemed "Compatible" for so long. That game is basically unplayable.

cxd4:
>why is github so fucking hard to use, cant figure out how to revert a commit

That's because commits can't be deleted through GitHub.

GitHub is just a high-level interface to using Git, integrated with your web browser.  If you want to do advanced operations like editing and correcting old commits, you'd need the command line for that.

dsx-:
Git, github, whatever, I meant the command line

cxd4:
To replace/modify the most recently committed commit:
```shell
git commit --amend --no-edit
```
Anything further past that, you may have to research how to use `git rebase` or `git rebase --interactive` to time travel and edit the historical information from the past to go back that way.

dsx-:
yeah i spent a long time researching but every "solution" was completely different. anyway i figured it out it seems since i reverted a few things

PR #965, opened 2016-02-02
Transfer pack

death-droid:
Implement the transferpak within Project 64 itself.
GBCart implementation should be fairly solid, tested it with a variety of games including Mario Tennis and both the Pokemon Stadium games. All appeared to work correctly.
Two things still need to be done, first is the GUI implementation, seccondly i want to change the way both the mempak and the transferpak are initialized, so there only loaded for controllers that are activated and are using that plugin. But need to commit these changes first to stop conflicts with the merge


vgturtle127:
@death-droid @project64 Other than changing the directory that is yours, you should change "Trasferpak.cpp" to "Transferpak.cpp". Otherwise, cool stuff! :+1:

death-droid:
haha wow i didnt even notice i did that :P
The files are now loaded through RDB as a per rom option, thought that was the easiest way to handle it.

project64:
looks good in general, tho I got a little worried when you said it when through the RDB, the code is correct, going though the .cfg file not the rdb.

Good to go through the cfg file, we can add UI latter to be able to set this but can easily test with just manually adding the entries.

A couple of things that would be nice to see just for consistency is the file handling to go Common/FileClass. It means that it is a lot harder to get a resource leak if there is a case of a missing fclose (tho I could not see any in the code).

Same goes for the loadbuffer/rom should be held in a std::auto_ptr.

Also the code uses new/delete not malloc/free.

also you have a very generic named functions "load_file", "read_from_file", etc that are not static or belong to the classes name space.

death-droid:
Cheers @project64 for your feedback :) will definitely look into resolving those problems

cxd4:
Aside from the overall size of the changes (which probably can't as much be helped with feature additions like this) and unchecked mix of tabs and spaces, I kind of like the style anyway.  It's a nice break away from always seeing the C++-oriented design adhered to throughout most of the project.

>also you have a very generic named functions "load_file", "read_from_file", etc that are not static or belong to the classes name space.

Those function names seem like a problem, but not really to me just because of the names by themselves.  It's more of a issue in the sense that all of this is going into controller pak support stuff in the core, which is not the only part of the project that applies load file or "read from file".  So the function names themselves don't seem bad or too generic, just overly localized to a specific module given said generality in their naming that does exist.

Classes or C++ approaches aren't the only clean ways to solve that problem, but I agree that using your nonstandard CFile class instead of the standard fopen/fclose pattern would help definitely keep aspects of that centralized.

cxd4:
The last thing I felt like sharing on was that I saw sometimes you have this type of thing repeated:
```c
if (gb_cart->has_rtc && (gb_cart->ram_bank >= 0x08 && gb_cart->ram_bank <= 0x0c)) {
    switch (gb_cart->ram_bank) {
    case 0x08:  ...
    case 0x09:  ...
    case 0x0A:  ...
    case 0x0B:  ...
    case 0x0C:  ...
```
It just looks slightly harder to maintain because in this situation the `&& (gb_cart->ram_bank >= 0x08 && gb_cart->ram_bank <= 0x0c)` is effectively already being done by the switch, having those 5 cases in it, therefore the extra check to see if 0x08 <= x <= 0x0C is respecification of the same constraint.

death-droid:
Cheers for the suggestions @cxd4, will look into improving the code more over the next few days.
Attempted to switch out the read_from_file, and load_file code for the CFile, but oddly seem to be having problems with it.  http://pastebin.com/SB5izUKc Probably doing something stupidly wrong in it, but so far it seems to refuse to work properly for me.

@cxd4, I'll try and grab a save state durring the GB Tower tomorrow.

purplemarshmallow:
When I tested the GB tower it worked with Gold/Silver/Crystal but Red/Blue/Yellow was not working. Also without CPU interpreter it will be very slow.

cxd4:
That sounds like it might not be RSP-related, then, or was that what Zoinkitty was trying to say....  Kind of left it open to interpretation.  Basically every game "requires" RSP emulation, lol.

I have no idea how CFile works.  I have no interest in how any C++ classes work.  In my opinion @project64 should either be willing to teach some explanation on that for the unwary of the custom classes or just merge the pull request and migrate to CFile on his own, provided any and all other aspects of the PR get fixed.

cxd4:
There's one other thing I forgot to comment on last night.  This function:
```c
static void read_gb_cart_normal(struct gb_cart* gb_cart, uint16_t address, uint8_t* data)
{
	uint16_t offset;

	if ((address >= 0x0000) && (address <= 0x7FFF))	{ //Read GB Cart
		if (address >= gb_cart->rom_size)
			return; //If address is larger then our rome size, bail out
		memcpy(data, &gb_cart->rom[address], 0x20);
	} else if ((address >= 0xA000) && (address <= 0xBFFF)) { //Read from RAM
		if (gb_cart->ram == NULL)
			return; //No RAM to write to
		offset = address - 0xA000;
		if (offset >= gb_cart->ram_size)
			return; //Offset is larger then our ram size

		memcpy(&gb_cart->ram[offset], data, 0x20);
		memcpy(data, &gb_cart->ram[offset], 0x20);
	}
}
```
Based on the number of commented error cases here, next to `return;` statements, for debugging purposes I would think this function should return a error code of some type (like `int` or `bool`).  Because, for each of those error conditions, normally I'd feel like throwing in error traces, message boxes, assertions or printfs to show the user that something when wrong...but if the function wasn't `void` it would just be able to be checked by the return value with a single error message for whether anything in the whole function failed.

This again isn't anything that I think makes an "issue" about the PR.  I don't think it affects whether it should be merged or not; it's just stuff I would probably feel like changing myself after it does get merged.

tony971:
@shonumi

project64:
@death-droid you have:
```
    rom_size = mFile.GetLength();
    rom = new uint8_t[rom_size];
    mFile.Read(&rom, rom_size);
    mFile.Close();
```

I assume it should be
```
    rom_size = mFile.GetLength();
    rom = new uint8_t[rom_size];
    mFile.Read(rom, rom_size);
    mFile.Close();
```

death-droid:
@cxd4 Yeah you make a good point.
@project64 cheers, I thought i was making some stupid little mistake.

death-droid:
@purplemarshmallow is there anything in specific you did, as while i can hear the audio play. Im just getting a white screen

LegendOfDragoon:
>Im just getting a white screen

Me too lol.

purplemarshmallow:
>is there anything in specific you did, as while i can hear the audio play. Im just getting a white screen

You can use a software gfx plugin. With Glide64 it will work with the "detect CPU writes" option but it will crash at the titile screen you have to enable it after that.

death-droid:
hmm not convinced the RTC implementation is correct, going to look into various other emulators implementation of it and see if that improves it.

death-droid:
@cxd4 @project64 Ok, code should be in an improved state, fixed the issues with the RTC, and made the code more consistent with other elements of project64.
let me know of any standout problems with either the code style or potential bugs.

Can make a cleaner pull request if needed.

cxd4:
Not really anything else, mostly I just look for signs of what could be broken.  I nitpick over style, but as far as merging goes (if it is my decision to make; for this repo it isn't) I commonly don't factor that in and would just do my own updates to the style.

I do have a question though about your last commit.  Sometimes you have this;
```c
        if (gb_cart->rtc_data[2] > 23) {
            gb_cart->rtc_data[2] -= 24;
            gb_cart->rtc_data[3]++;
        }
```
other times this
```c
            if (gb_cart->rtc_data[3] > 511) {
                gb_cart->rtc_data[3] %= 512;
                gb_cart->rtc_data[3] |= 0x80;
            }
```
So, sometimes you have `if (data[3] >= limit) data[3] -= limit;`, other times you have `if (data[3] >= limit) data[3] %= limit;` ... they are of course similar in some ways but different; could it be that the RTC behavior is the same for both but just uses subtraction as an efficiency bypass to hack around doing modulus?  Or is guaranteeing of correct results for both algorithms enforced by %= ?

Also zilmar might say don't forget about removing the include stdint.h; you still have that there.

death-droid:
Ahhh, if there is any standout style issues that you see, let me know. I'm interested in any way to improve my coding.

That code is actually siphoned form VBA-M since we make use of there save format, i thought it was safest to handle the RTC time correction in the same manor they do., I meant to make a note of that but forgot. I assume its completely intentional as most implementations of the same method handle it the same way from what I have seen.

project64:
looks good I can merge it tonight if you happy with it.

While there is nothing wrong with this, I just find it strange that you have structs and classes for basically the same thing. eg gb_cart with GBCart and struct transferpak  with  class Transferpak

death-droid:
Yeah I'm happy enough with it to be merged. Any changes needed are more due to code styling then anything that effects the operation of it.

PR #968, opened 2016-02-02
disable HLE gfx for Yakouchuu II - Satsujin Kouro

purplemarshmallow:
this game uses the RSP to decode videos. It cannot work with HLE gfx.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -6896,6 +6896,7 @@ Status=Compatible
 Counter Factor=1
 Culling=1
 RDRAM Size=8
+HLE GFX=No
 
 [2DCFCA60-8354B147-C:4A]
 Good Name=Yoshi Story (J)


cxd4:
I should research that.  Is it similar to the problems (or the video type) of HLE with _Resident Evil 2_?

purplemarshmallow:
Resident Evil 2 works with HLE plugins like Glide64. But if you use a HLE gfx plugin in this game all you can get is a fatal error.

cxd4:
That doesn't mean it "cannot work with HLE", though.

That **was** the case with RE2.  HLE plugins got around this over time by bypassing the null task pointer and letting it partially fall through to LLE, for the MPEG videos.  So actually RE2 isn't really done in pure HLE either, but that didn't stop a HLE graphics plugin from still being plugged in with it.

Basically what I'm curious to look at here isn't, "Are there any HLE gfx plugins that work with this game;" it's, "Why aren't there any?"  Because I'm sure it's never impossible technologically.

dsx-:
99.5% sure these videos worked in HLE when i did my testing in 2013 or so

vgturtle127:
Should we add that RDB entry to Winback, too? LLE is the only way to properly handle that game too.

purplemarshmallow:
>That was the case with RE2. HLE plugins got around this over time by bypassing the null task pointer and letting it partially fall through to LLE, for the MPEG videos. So actually RE2 isn't really done in pure HLE either, but that didn't stop a HLE graphics plugin from still being plugged in with it.

really? I don't think so. RE2's movies show up with zilmar's cfb plugin. There is some palette bug however. But I don't think the cfb plugin has a partial fall through to LLE. The CPU decodes the movies in RE2 that's the reason why it works in HLE. The RSP has nothing to do with it.

But a partial fall through to LLE seems to be a good idea actually. Might be possile to get it working this way with HLE plugins. But if there's a plugin that already does this that would be news to me.

cxd4:
>really? I don't think so. RE2's movies show up with zilmar's cfb plugin. There is some palette bug however. But I don't think the cfb plugin has a partial fall through to LLE. The CPU decodes the movies in RE2 that's the reason why it works in HLE. The RSP has nothing to do with it.

You sound rather sure of yourself.

I wonder if you'd feel the same way if you re-considered the meaning of the plugin's title:  "Basic CFB Plugin".  It's almost as if that, by itself, tells you whether the plugin is HLE or LLE, but you don't honestly know, do you?

Why not have a look at the source to his plugin for yourself and tell me what you think:
http://www.zophar.net/utilities/n64gpuplugins/cfb-plugin.html

LegendOfDragoon:
>really? I don't think so. RE2's movies show up with zilmar's cfb plugin. There is some palette bug however. But I don't think the cfb plugin has a partial fall through to LLE. The CPU decodes the movies in RE2 that's the reason why it works in HLE. The RSP has nothing to do with it.

It's possible to use the RSP w/o using RDP. Pokemon Puzzle League does this at the very beginning. I think the audio in the beginning is done on the CPU as well.

purplemarshmallow:
>You sound rather sure of yourself.

>I wonder if you'd feel the same way if you re-considered the meaning of the plugin's title: "Basic CFB Plugin". It's almost as if that, by itself, tells you whether the plugin is HLE or LLE, but you don't honestly know, do you?

>Why not have a look at the source to his plugin for yourself and tell me what you think:
http://www.zophar.net/utilities/n64gpuplugins/cfb-plugin.html

The cfb plugin works with both HLE and LLE. If I have HLE checked movies in Yakouchuu II don't work but if it's unchecked they work correctly. In RE2's case it does not matter if HLE is checked or not.

cxd4:
>The cfb plugin works with both HLE and LLE.

No, the CFB plugin works with **neither** HLE _nor_ LLE.

What it works with is VI DMA.  There is no RDP emulation in the basic CFB plugin.  The RSP plugin is responsible for handling the graphics microcode.  You seem to think it's up to the graphics plugin, based on whether your Project64 checkbox "Use HLE GFX?" has any control over the graphics plugin.  That's backwards:  It's the RSP plugin that decides when to call the microcode simulation in the graphics plugin.

cxd4:
>If I have HLE checked movies in Yakouchuu II don't work but if it's unchecked they work correctly.

The only thing that proves is that said movies use the RSP.

They don't work correctly if you check Project64's checkbox because that just tells the RSP to ask the graphics plugin to process the lists instead.  In case you have yet to find this, there is no HLE or LLE code in the basic CFB plugin.  The RSP is what processes the microcode.

>In RE2's case it does not matter if HLE is checked or not. 

Again, because RE2 has a null task pointer in the HLE struct, so it cannot be HLE'd directly.  It must be done in LLE or from inside the RSP plugin.  Asking Project64 to "HLE" it is a futile request because PJ64 knows that's the default setting, so it checks if the request is legitimate otherwise it just fall through to LLE in that case.

purplemarshmallow:
Yes it makes sense. Now I remember RE2 does not work with Mupen's HLE RSP. Must be because this plugin has no fall through to LLE. Maybe it's possible to do something similar for Yakouchuu II

PR #981, opened 2016-02-05
[Glide64] undeclared memset and strlen in trace.cpp

cxd4:
I was not getting these errors before because I forgot to add `Config.cpp` and the recently added `trace.cpp` and `Settings.cpp` which @project64 added to the repository a few days ago.

Just added them to my Unix make script to work out some errors Mupen is giving when trying to initialize the plugin so that I can finally start emulation with it.  For now, this just fixes:
```
Compiling Glide64 plugin sources...
./../../Glide64/trace.cpp: In function 'void SetupTrace()':
./../../Glide64/trace.cpp:64:39: error: 'memset' was not declared in this scope
     memset(log_dir, 0, sizeof(log_dir));
                                       ^
./../../Glide64/trace.cpp:67:23: error: 'strlen' was not declared in this scope
     if (strlen(log_dir) == 0)
                       ^
```

### DIFF
--- a/Source/Glide64/trace.cpp
+++ b/Source/Glide64/trace.cpp
@@ -1,5 +1,7 @@
 #include "trace.h"
 #include "Config.h"
+
+#include <string.h>
 #include <Common/Trace.h>
 #include <Common/path.h>
 #include <Common/LogClass.h>


cxd4:
Unrelated to PR, but, anyone know what's up with this?
```
Linking PJGlide64 objects...
./Glide64/Config.o: In function `general_setting(short, char const*, unsigned int)':
Config.cpp:(.text+0x0): multiple definition of `general_setting(short, char const*, unsigned int)'
./Glide64/Config.o:Config.cpp:(.text+0x0): first defined here
./Glide64/Config.o: In function `game_setting(short, char const*, unsigned int)':
Config.cpp:(.text+0x15): multiple definition of `game_setting(short, char const*, unsigned int)'
./Glide64/Config.o:Config.cpp:(.text+0x15): first defined here
collect2: error: ld returned 1 exit status
```
I'm only finding one declaration, not two, of general_setting() and game_setting() in Config.cpp.

PR #991, opened 2016-02-10
Optimize CheckInterrupts, PI_DMA_WRITE, and UpdateTimers

LegendOfDragoon:


Source/Project64-core/N64System/Mips/Dma.cpp
```diff
@@ -198,12 +198,8 @@ void CDMA::PI_DMA_READ()
...
 void CDMA::PI_DMA_WRITE()
 {
-    uint32_t PI_WR_LEN_REG = ((g_Reg->PI_WR_LEN_REG) & 0x00FFFFFFul) + 1;
-
-    if ((PI_WR_LEN_REG & 1) != 0)
-    {
-        PI_WR_LEN_REG += 1; /* fixes AI Shougi 3, Doraemon 3, etc. */
-    }
+    uint32_t PI_WR_LEN_REG = ((g_Reg->PI_WR_LEN_REG) & 0x00FFFFFEul) + 2;
+    /* rounding PI_WR_LEN_REG up to the nearest even number fixes AI Shougi 3, Doraemon 3, etc. */
```
cxd4: I don't know how pedantic this sounds, but that's really not rounding reg up to the nearest even number.

Functionally speaking, this changes the entire algorithm to the functional input-output table of:
```
f(0x000000) = 2
f(0x000001) = 2
f(0x000002) = 4
f(0x000003) = 4
f(0x000004) = 6
f(0x000005) = 6
f(0x000006) = 8
```

Furthermore, I doubt this would be conspicuously faster than just `PI_WR_LEN_REG += PI_WR_LEN_REG & 1;`.

LegendOfDragoon: You're right about it not exactly rounding up to nearest even number ;/ . I'll try and come up with better wording and change the comment.

Sadly with minimize size, ```PI_WR_LEN_REG += PI_WR_LEN_REG & 1;```  actually appears to be better. I prefer using maximize speed, but I can understand why others would rather use minimize size. I can't conveniently check the compiler output for maximize speed because changing the options doesn't seem to work. Not sure if I broke something on my end or what. I'll change it if @project64 prefers ```PI_WR_LEN_REG += PI_WR_LEN_REG & 1;``` .

PR #993, opened 2016-02-11
[RSP] Get Dma.c to start to compile outside Windows.

cxd4:
Most type errors in header files like `Rsp.h` spec header have been fixed now, with just the special struct types PAINTSTRUCT and RECT left over to replace.  Those will be harder to proofread though, so I can try to PR those later.

As all the sources immediately fail with a fatal error of cannot find windows.h or "rsp.h" or the like, most changes were done to `Rsp.h` to get most of them to begin to compile.

### DIFF
--- a/Source/RSP/RSP Command.c
+++ b/Source/RSP/RSP Command.c
@@ -29,7 +29,7 @@
 #include <stdlib.h>
 
 #include "opcode.h"
-#include "RSP.h"
+#include "Rsp.h"
 #include "CPU.h"
 #include "RSP Registers.h"
 #include "RSP Command.h"
--- a/Source/RSP/Cpu.c
+++ b/Source/RSP/Cpu.c
@@ -27,7 +27,7 @@
 #include <windows.h>
 #include <stdio.h>
 #include <float.h>
-#include "RSP.h"
+#include "Rsp.h"
 #include "Cpu.h"
 #include "RSP registers.h"
 #include "RSP Command.h"
--- a/Source/RSP/Interpreter CPU.c
+++ b/Source/RSP/Interpreter CPU.c
@@ -28,7 +28,7 @@
 #include <stdio.h>
 #include <float.h>
 #include "breakpoint.h"
-#include "RSP.h"
+#include "Rsp.h"
 #include "Cpu.h"
 #include "Interpreter Ops.h"
 #include "Interpreter CPU.h"
--- a/Source/RSP/Interpreter Ops.c
+++ b/Source/RSP/Interpreter Ops.c
@@ -27,7 +27,7 @@
 #include <windows.h>
 #include <stdio.h>
 #include <math.h>
-#include "RSP.h"
+#include "Rsp.h"
 #include "CPU.h"
 #include "RSP Command.h"
 #include "RSP Registers.h"
--- a/Source/RSP/Mmx.c
+++ b/Source/RSP/Mmx.c
@@ -26,7 +26,7 @@
 
 #include <windows.h>
 #include <stdio.h>
-#include "rsp.h"
+#include "Rsp.h"
 #include "x86.h"
 #include "memory.h"
 #include "RSP registers.h"
--- a/Source/RSP/RSP Register.c
+++ b/Source/RSP/RSP Register.c
@@ -27,7 +27,7 @@
 #include <windows.h>
 #include <commctrl.h>
 #include <stdio.h>
-#include "rsp.h"
+#include "Rsp.h"
 #include "types.h"
 
 #define GeneralPurpose			1
--- a/Source/RSP/RSP Registers.h
+++ b/Source/RSP/RSP Registers.h
@@ -24,7 +24,7 @@
  *
  */
 
-#include "types.h"
+#include "Types.h"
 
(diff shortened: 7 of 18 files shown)


Source/RSP/RSP Command.c
```diff
@@ -29,7 +29,7 @@
 #include <stdlib.h>
 
 #include "opcode.h"
-#include "RSP.h"
+#include "Rsp.h"
 #include "CPU.h"
 #include "RSP Registers.h"
```
JunielKatarn: Why not leave it as `RSP.h`? Given that other capitalized file names seem to follow that convention (i.e. `CPU.h`, `RSP Registers.h`...)

JunielKatarn:
Head up. This change broke the build:
```
Build (Win32)
8 error(s), 179 warning(s)
Build (Win32) - 8 error(s), 179 warning(s)
Source\RSP\Main.cpp (60): 'CPUCore': redefinition; different basic types
Source\RSP\Main.cpp (66): 'hinstDLL': redefinition; different basic types
Source\RSP\Main.cpp (100): 'DWORD AsciiToHex(char *)': overloaded function differs only by return type from 'uint32_t AsciiToHex(char *)'
Source\RSP\Main.cpp (100): 'AsciiToHex': redefinition; different basic types
Source\RSP\Main.cpp (245): '=': cannot convert from 'void (__cdecl *)(HWND,RECT)' to 'void (__cdecl *)(void *,RECT)'
Source\RSP\Main.cpp (248): '=': cannot convert from 'void (__cdecl *)(HWND)' to 'void (__cdecl *)(void *)'
Source\RSP\Main.cpp (250): '=': cannot convert from 'void (__cdecl *)(HWND,int)' to 'void (__cdecl *)(void *,int)'
(): Unexpected exit code received from msbuild.exe: 1
```

@cxd4 I think you need to also update the source (cpp) files to match the updated method signatures.

cxd4:
I know.  You can fix it yourself.

At this time I'm not responsible for how Windows compiles look, as I have no way to test that.

JunielKatarn:
@cxd4 Yeah, not gonna become your errands boy. Pj64 is still a Windows application, and changes should be designed in a way that don't break everyone else's workflow.
Does this change at least work on UNIX (meaning, RSP fully compiles), or am I going to have to send a PR to revert it?

cxd4:
>Pj64 is still a Windows application,

Which is why it **already** breaks everyone else's workflow. :P A little late for your noble claim, eh?

That being said, this looks like a good opportunity for you to finally start contributing to the actual code.  You can either start by dropping the developer ego and helping make Project64 actually compile, or you can patiently wait for me to finish getting the rest of the remaining GNU compile errors fixed.  At which time, I will have access to the same errors as the ones the Windows build is giving, and will finally be able to fix those.

Source/RSP/RSP Command.c
cxd4: @JunielKatarn If you change it back to `"RSP.h"`, you are going to break non-Windows compiles.

JunielKatarn: @cxd4 Understood. The filename must be Rsp.h, then. I might submit a PR to "standardize" header names and includes under RSP, but it would be useful to test-compile on UNIX.

Can RSP be built out of the box with your UNIX make scripts?

cxd4: This is what ultimately is in progress.

See, I could theoretically do it your way and make the entire RSP compile out of the box on Unix-based systems, in one big PR, so that no blockage can hide Windows compile errors so that no Windows compile errors would be happening.  The problem with this approach, is the PRs become enormous, full of changes all over the files for fixing WIN32-ism's.  So I find it wiser to make smart incremental, smaller, easy-to-review-and-criticize pull requests.

JunielKatarn: I'll just wait for UNIX builds to stabilize, then.
I do think that "large" PRs are inevitable in some cases. We might revisit that issue when I submit one of those in the near future XD

cxd4: Relatively speaking, large PRs indeed are inevitable in some cases.

More so with things like upgrading wx, than this RSP stuff I am doing.  Large PRs would not be necessary here with the RSP if I had a "Find All References" kind of feature here on GNU, like Visual Studio for Windows has.  Without that, I have no way of preventing WIN32 build errors without doing a Ctrl+F find and search through tens of the source files looking for multiple externs or repeated declarations, or waiting for you to post a list of the errors (which you did) and fixing some of those.  Otherwise it's kind of infeasible for me...though I talked to project64 and he said he will fix the WIN32 stuff tomorrow.

JunielKatarn: Not trying to be patronizing but, is it too much trouble to restart on Windows and test-build? (A simple msbuild.exe Project64.sln should do. No need to even open Visual Studio).

I would gladly settle for a warning about incoming build breaks beforehand.

cxd4: With a little re-installation of Windows, I'm sure it's conceivable, but as the project is in the hands of the open-source community I felt that any other contributors would be happy lending a hand.  Besides the fact that these changes get PJ64 portable, they're clean-ups as well for the better.

PR #995, opened 2016-02-11
Force interpreter on x64 builds

Lithium64:
Basically it's the same pull request of #603 since this was probably accidentally reversed by 3e888b9

### DIFF
--- a/Source/Project64-core/N64System/N64Class.cpp
+++ b/Source/Project64-core/N64System/N64Class.cpp
@@ -928,7 +928,7 @@ void CN64System::ExecuteCPU()
     switch ((CPU_TYPE)g_Settings->LoadDword(Game_CpuType))
     {
         // Currently the compiler is 32-bit only.  We might have to ignore that RDB setting for now.
-#ifdef _WIN32
+#ifndef _WIN64
     case CPU_Recompiler: ExecuteRecompiler(); break;
     case CPU_SyncCores:  ExecuteSyncCPU();    break;
 #endif


cxd4:
What about the RSP?  Do you think you prefer 64-bit RSP to use the interpreter?

Lithium64:
@cxd4 I was writing that actually rsp works on interpreter mode, I think the rsp interpreter mode should also be the default while the recompiling mode isn't ported to x64.

cxd4:
>I was writing that actually rsp works on interpreter mode,

Good to hear.  Sounds like that issue may be closed then.

>I think the rsp interpreter mode should also be the default while the recompiling mode isn't ported to x64.

Maybe.  Leaving it set to recompiler could remind someone to port recompiler to x64, but I don't really care much for recompilers so either way is fine to me.  Just curious which you preferred.

loganmc10:
By the way, I can't get Project64 to compile on 64 bit Linux, I get this error:

```
./Project64/Project64-core/N64System/Recompiler/LoopAnalysis.o: In function `LoopAnalysis::SetupEnterSection(CCodeSection*, bool&, bool&)':
LoopAnalysis.cpp:(.text+0x1614): undefined reference to `CX64RegInfo::operator==(CX64RegInfo const&) const'
LoopAnalysis.cpp:(.text+0x165e): undefined reference to `CX64RegInfo::operator==(CX64RegInfo const&) const'
```

project64:
@loganmc10  would not suprise me, project64 does not have a 64bit recompiler so I never test can 64bit build ...

cxd4:
It's been a while since I had Windows but the 64-bit Windows build does work.

You're not supposed to test the recompiler until the interpreter works first though because the recompiler certainly isn't going to work for 64-bit (either Linux or Windows) until it's totally rewritten to take advantage of the architectural upgrades.

cxd4:
As far as just fixing that compile issue here on my 64-bit Linux I could take a look at it but it would be a very time-consuming process compared to what I have to use my time for now.

The fact that a compile error or two exists is pretty insignificant compared to the number of things that just, aren't ported (even though it compiles, links, and runs) and thus just don't work correctly.  For example that CMODULE_DIRECTORY thing I ported from Windows to Linux that zilmar said he was going to test on a weekend break but never found the time to merge for whatever reason.

PR #997, opened 2016-02-12
Updated Spanish Translation

MELERIX:


Lang/Spanish.pj.Lang
```diff
@@ -5,7 +5,7 @@
 #1  #  "Español"		// LANGUAGE ID
 #2  #  "MELERIX"		// Author
 #3  #  "2.2"			// Version
-#4  #  "01/12/2015"		// Date
+#4  #  "12/02/2016"		// Date
```
JunielKatarn: In order to not confuse non-Hispanic users (granted that they might never use this translation), why not use the YYYY-MM-DD format instead?
2016-02-12

I guess the same would apply for other translations.

MELERIX: yes, I would like to see a standard time format for all the language files.

PR #1004, opened 2016-02-15
[RSP] Get Main.cpp compiling (almost) outside Windows.

cxd4:
The remaining errors are caused by the sub-header includes saying DWORD, which I will delay replacing with `uint32_t` in favor of reducing the interval of instantly causing WIN32 compile errors.

If I temporarily add a `typedef uint32_t DWORD;` to rsp.h, then the errors for Main.cpp are all gone.


project64:
this pull request has conflicts so I can not merge

cxd4:
Conflict resolved:
```c
<<<<<<< HEAD
EXPORT void InitiateRSP(RSP_INFO Rsp_Info, DWORD * CycleCount)
=======
__declspec(dllexport) void InitiateRSP ( RSP_INFO Rsp_Info, uint32_t * CycleCount)
>>>>>>> master
```

Really it ought to be `unsigned long * CycleCount` and not `uint32_t * CycleCount` so that you can maximize storage capacity for the number of cycles, but rather than have that argument with you again I think I will just change it to `uint32_t`.  Who knows maybe there is some 32-bit RCP register that cares about it being exactly a `uint32_t`, though I doubt it.

PR #1017, opened 2016-02-28
Fix Smash Bros. PAL

vgturtle127:
This removed the VI Refresh and also adds the Counter Factor=1 entry


LegendOfDragoon:
why not fix the other regions too? I'm not sure if 2200 is good for that game. I use 1500.

vgturtle127:
@LegendOfDragoon You're right! I updated the RDB with your suggestions. Thanks!

PR #1018, opened 2016-02-29
[Glide64] fix up LLE

purplemarshmallow:
Based on GLideN64
World Driver Championship now works with Glide64. A bit buggy still...
glide64_world_driver_champ_01

### DIFF
--- a/Source/Glide64/rdp.cpp
+++ b/Source/Glide64/rdp.cpp
@@ -4093,102 +4093,102 @@ output:   none
 void CALL ProcessRDPList(void)
 {
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


cxd4:
I can't really see what's being changed because your version of Visual Studio seems to be converting everything to tabs, causing the entire function to be highlighted instead of just the changed lines.

purplemarshmallow:
I changed it back to spaces

AmbientMalice:
This is certainly interesting. Biggest problem seems to be depth. Some crazy perspective/depth sorting issues in Mario 64. Speed is okay. Some games render upside down for some reason.

Issue #1015, opened 2016-02-26
Super Smash Bros. (Europe) is broken unless you change the RDB plus it's way to fast


SigmaVirus:
Just thought I would give you guys and gals a heads up regarding Super Smash Bros. (Europe)

If you attempt to play the game on newer versions (Project64 2.2.0.9999, 0.0.0.421 tested) where the entry in the RDB looks like this:

[93945F48-5C0F2E30-C:50]
Good Name=Super Smash Bros. (Europe)
Internal Name=SMASH BROTHERS
Status=Compatible
Culling=1
SMM-Cache=0
SMM-FUNC=0
SMM-TLB=0
ViRefresh=2200

you'll soon to find out that the game will be broken and all you see is a black screen with sound and pressing any buttons on the controller will do nothing.

To fix this, you'll have to add this line:

Counter Factor=1

Now the game starts and can be played but it's way to fast so in order to fix that you'll have to delete this line:

ViRefresh=2200

making ViRefresh defaulting to the value 1500 and the speed will be OK.

So for PAL/Europe owners of the game the RDB entry for Super Smash Bros. have to look like this:

[93945F48-5C0F2E30-C:50]
Good Name=Super Smash Bros. (Europe)
Internal Name=SMASH BROTHERS
Status=Compatible
**Counter Factor=1**
Culling=1
SMM-Cache=0
SMM-FUNC=0
SMM-TLB=0

I have edited my own personal RDB to this and the game is working perfectly.

If someone could fix this in the official/original RDB so that anyone owning a PAL copy can play this game it would be great.

Thanks in advance!

cxd4:
Try clicking this if you are new to Git:
https://github.com/project64/project64/edit/master/Config/Project64.rdb

You can use this web interface to "edit" the RDB file with your suggested change of `Counter Factor = 1` being added in to it.  When you save the "edit" it should create a pull request to the repository, which zilmar can then review and/or merge to adopt your change.

vgturtle127:
@SigmaVirus I went ahead and added it to #1017 , so you don't have to worry about it.

SigmaVirus:
A big thank you for this guys!

However, @vgturtle127 

I don't know, and it's only a hunch but do you think the Australian version might be affected by this too since it's a PAL region?

Obviously the USA version works because I've seen screenshots from it on issues regarding GLideN64 and heard no one complain about it so far regarding this kind of issue.

I know I'm one of the few that emulates the European versions but that's where I'm from and I got my games.

[...]

PR #1019, opened 2016-02-29
SigmaVirus CF=1 fix to PAL Super Smash Bros.

cxd4:
He reported the PAL issue #1015 and his fix:  `Counter Factor=1`, which is all that is relevant to the specific issue here.  Though I was hoping he might have done this PR himself, I can do it if need be.

Sorry, no clue why line number 1 got changed here.
EDIT:  Looks like it's an auto-fix to https://github.com/project64/project64/commit/6d4cb1c2e3268a34bae7c5272533ec9c87de3474#diff-6ed817341fe13121d9cae89c071156ccL1

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -1,4 +1,4 @@
-﻿// ============ RDB for PJ64 v2.2. GoodN64 v321 =====================================
+// ============ RDB for PJ64 v2.2. GoodN64 v321 =====================================
 // PJ64 v2.2 Official RDB
 // Not for use with PJ64 v1.6 or previous
 //---- START OF RDB FILE HEADER ---------------------------------------------------------
@@ -5548,6 +5548,7 @@ ViRefresh=2200
 Good Name=Super Smash Bros. (E) (M3)
 Internal Name=SMASH BROTHERS
 Status=Compatible
+Counter Factor=1
 Culling=1
 SMM-Cache=0
 SMM-FUNC=0


JunielKatarn:
What does 'Counter Factor' mean?
I want to understand what this fixes.

AmbientMalice:
It's a timing hack that most N64 emulators use in the absence of proper CPU timing.

vgturtle127:
@cxd4 Should I close #1017? I also remove the ViRefresh values which fixes some things as well, but I will default to you since you are a contributor.

dsx-:
i feel like cf1 is bad because it causes so much more cpu load than cf2 but i guess if thats the only fix for this then whatever

AmbientMalice:
So why exactly is the PAL version getting CF=1?

LegendOfDragoon:
>So why exactly is the PAL version getting CF=1?

Probably because of VI refresh being at 2200.

SigmaVirus:
Stop the presses........

I just tried to disable the Counter Factor=1 flag in my RDB so it would revert back to Counter Factor=2 in the latest Project64 build from here:
http://hyvart.com/download/project64/ci/

and the game is now booting correctly for some odd reason and I can't replicate this problem anymore that have been haunting me for quite some time.

I have no logical explanation for this, because for the darnest longest time I've had to set Counter Factor=1 in the RDB, now all of a sudden it decided to boot up on Counter Factor=2

Still, I left ViRefresh defaulted to 1500 so that I didn't change.

Is this good or bad?

cxd4:
Can't really test on this machine so can't really say.

dsx-:
He literally just said stop the presses and it gets merged lmao
(E) has booted on cf2 for like 15 years so any problems should be fixed in the core

cxd4:
I was kind of on the verge of just canceling the PR tbh, but I don't think it matters.

CF is a hack because CF=1 is a hack, CF=2 is a hack, CF=3 is a hack...whatever you have it set to corresponds to the presence of a hack regardless.  I only made a PR out of it cause he didn't himself.

Then there's VI Refresh which is also a hack.  Setting that to 1500 instead of 2200 probably caused him to not to change the setting to a different hack like counter factor.

SigmaVirus:
I have not been able to reproduce the original issue anymore since I last posted here earlier today.

I see that this is merged yet deleted. So what should I do here, just close the issue altogether?

cxd4:
I'm sure that's no coincidence.  If anyone else but you could reproduce the issue, it's possible that the official, unmodified RDB file had the issue and you just had some other changes in it that also fixed it.

What is deleted is my local branch used to safely propose the pull request, which simply has been rendered useless by the fact that the proposed changes have been merged.

SigmaVirus:
I'll just close the issue then, thanks for the help guys!

PR #1020, opened 2016-03-02
SigmaVirus Linking=off fix to PAL Shadows of the Empire

### LINKED ISSUE #1014: Level select is freezing on Star Wars - Shadows of the Empire
This is a issue that have been bugging me for quite a while and frankly the only thing that's keeping me from switching to newer versions of Project64 altogether.

I'm on Windows 10 x64 and I'm currently using Project64 2.2.0.9999 (0.0.0.415) 32-bit from here:
http://hyvart.com/download/project64/ci/

As for plugins I use either Glide64 Final from May 2012 or the latest GLideN64 by Gonetz from here:
https://github.com/gonetz/GLideN64/issues/808#issuecomment-183673274

Azimer Xaudio2 latest Git (December 2015) and Nrage Input from here:
https://sourceforge.net/projects/nragev20/

Hardware wise I'm on a i7 4970K CPU, ASUS R9 380 GPU with latest AMD driver 16.1.1, 16GB RAM etc.

OK, now to the issue.

When I start the game, select my game file and about to select level it's as the level select freezes and I can't choose any level (Try going/scrolling up and down)
Instead I can only play the level the cursor stays/lands on which is the previous completed level and the funny thing is it looks unbeaten with no hi-score/status at all.

There is nothing wrong with the controls and control wise I'm using a genuine gray Nintendo 64 controller with an Mayflash USB adapter and to clear/stress things out, this issue does not occur/exist on Project64 1.7.0.50 rev23 which is the version of Project64 I'm currently sticking with/keeps falling back to.

ROM wise I'm using my own dumped PAL/Europe version of the game with the RDB entry:

[4D486681-AB7D9245-C:50]
Good Name=Star Wars - Shadows of the Empire (Europe)
Internal Name=Shadow of the Empire
Status=Compatible
Core Note=Swoop bike speed issue
RDRAM Size=8

I'm hoping I'm not alone with this issue and that it's an easy fix.

This is really the only thing holding me back these days from migrating to newer versions of Project64 since it's become much stabler.

Thanks in advance!

SigmaVirus:
This should fix #1014 - Level select is freezing on Star Wars - Shadows of the Empire at least for the PAL/Europe version.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -5318,6 +5318,7 @@ Good Name=Star Wars - Shadows of the Empire (E)
 Internal Name=Shadow of the Empire
 Status=Compatible
 Core Note=Swoop bike speed issue
+Linking=Off
 RDRAM Size=8
 
 [264D7E5C-18874622-C:45]


cxd4:
Good job making the pull request.

Yeah the ABL=Off copies throughout the RDB file can help document all the reminders of cases where people can tackle bug-fixing the recompiler core, so it's good that you also found that.

JunielKatarn:
What is the `Linking` setting supposed to do?
What, technically, will this change achieve for that game?

SigmaVirus:
@cxd4 
Thanks, turned out making a pull request wasn't so hard that I imagined.
Yeah I think setting Linking=off was the better alternative here then to set the CPU Type=Interpreter because as you say it helps document the reminders of cases where there is bugs in the Recompiler core.

@JunielKatarn 
Well, Linking=off is the same as disabling Advanced Block Linking in the GUI and apparently ABL is specific for the Recompiler core.
That's all I know really, that and it fixes https://github.com/project64/project64/issues/1014

cxd4:
Correct, and the interpreter is always more straightforward (hence stable) in algorithm, just slower.

It is technically harder to make pull requests, but the GitHub website interface for doing it should be fine for things like editing the RDB and support files/plain text.  The finer details of using Git commands through the terminal (with or without a graphical interface like GitHub's online one) are usually beneficial to code contributions that involve compiling and testing--stuff that won't matter for this.

AmbientMalice:
You may find this worth reading. 
https://en.wikibooks.org/wiki/Emulation/How_does_it_work%3F#Interpretation
Modern PCs are generally fast enough to emulate the N64 CPU via interpretation, but that CPU overhead could better be spent on other things.

Issue #856, opened 2015-12-31
Splitting Out Jabo's D3D Settings from RDB


project64:
@AmbientMalice and @vgturtle127 were talking about removing jabo's D3D settings from the RDB. We could split it out like I did for glide settings.

When the d3d plugin registers settings these are the ones it registers:
Direct3D8-ForceFilter
Direct3D8-2xSai
Direct3D8-Brightness
Direct3D8-DesiredAspect
Direct3D8-Direct3DPipe
Resolution Width
Resolution Height
ClearvFrame
Aspect Correction

So I can match any starting with "Direct3D8-" and manually look for the 4 other ones.

interestingly there are options in the rdb that might have applied to 1.6 plugin, but not used now, I am not sure, things like: Emulate Clear, Culling, Self Texture

vgturtle127:
I will gladly work on this, I just need permission and any direction you would like me to go. :smile:

project64:
I figured lets discuss it first which is why I created the issue here.

code wise if you look at Source\Project64-core\Settings\SettingType\SettingsType-RomDatabase.cpp and look at how m_GlideSetting is used then you should get a good idea how it should be done. Just need a new file to read from.

As I do not have the source for jabo's plugin I have no idea if he now goes out side of the emu to read settings, but as long as it goes through the emu then it is all good to redirect them.

tony971:
You could go the way of GLideN64 and have an ini for the plugin, specifically. The RDB should only have universally needed settings.

Frank-74:
Culling is an option I still use.  1.7+ Jabo uses the .cfg and .rdb from Config folder.  Latest 1.6.1 reads only the RDB in emulator folder.

vgturtle127:
@project64 Outside of the emulator? What do you mean?

@tony971 Isn't that already how it works kind of? I mean, Glide and GlideN64 do. And if the emulator's have their own (Mupen64, 1964, Nemu64, Mupen64Plus, Project64) what is left? :laughing: 

@Frank-74 You mean when you use Jabo's plugin, you use the culling options? Is culling an RDB entry that effects multiple plugins? That is pretty much what I want to know, all the exclusively Jabo-specific entries in the RDB that are possible, like Culling and Self Texture.

Frank-74:
[...]

Issue #235, opened 2015-03-07
[Suggestion] RDB should be region-independent


oddMLan:
I'm not good at explaining things, so I'll try to explain what I mean by using the RDB entries for Majora's Mask.

This:

```
[9FC385E5-3ECC05C7-C:50]
Good Name=The Legend of Zelda - Majora's Mask (E) (M4) (Debug Version)
Internal Name=ZELDA MAJORA'S MASK
Status=Compatible
Cote Note=(see GameFAQ)
Plugin Note=[video] errors:day transistion (see GameFAQ)

[E97955C6-BC338D38-C:50]
Good Name=The Legend of Zelda - Majora's Mask (E) (M4) (V1.0)
Internal Name=ZELDA MAJORA'S MASK
Status=Compatible
Cote Note=(see GameFAQ)
Plugin Note=[video] errors:day transistion (see GameFAQ)
Clear Frame=0
Culling=1
Emulate Clear=1
Primary Frame Buffer=0
RDRAM Size=8
Self Texture=1

[0A5D8F83-98C5371A-C:50]
Good Name=The Legend of Zelda - Majora's Mask (E) (M4) (V1.1)
Internal Name=ZELDA MAJORA'S MASK
Status=Compatible
Cote Note=(see GameFAQ)
Plugin Note=[video] errors:day transistion (see GameFAQ)
Clear Frame=0
Culling=1
Emulate Clear=1
Primary Frame Buffer=0
RDRAM Size=8
Self Texture=1

[5354631C-03A2DEF0-C:45]
Good Name=The Legend of Zelda - Majora's Mask (U)
Internal Name=ZELDA MAJORA'S MASK
Status=Compatible
Cote Note=(see GameFAQ)
Plugin Note=[video] errors:day transistion (see GameFAQ)
Clear Frame=0
Culling=1
Emulate Clear=1
Primary Frame Buffer=0
RDRAM Size=8
Self Texture=1

[6AECEC4F-F0924814-C:50]
Good Name=The Legend of Zelda - Majora's Mask - Collector's Edition (E) (M4) (GC)
Internal Name=ZELDA MAJORA'S MASK
Status=Compatible
Cote Note=(see GameFAQ)
Plugin Note=[video] errors:day transistion (see GameFAQ)
Clear Frame=0
Culling=1
Emulate Clear=1
Primary Frame Buffer=0
RDRAM Size=8
Self Texture=1

[B443EB08-4DB31193-C:45]
[...]

PR #1025, opened 2016-03-14
Set CF=1 for the Ganbare Goemon games

oddMLan:
Fixes the out-of-sync intro
Compared against a real N64

Also made Culling=1 consistent across all the games that support it for small speedup using Jabo's D3D8


oddMLan:
I still think all the Jabo settings should be separated in their own RDB
#856 

If the RDB was region-independant, all these redundant settings across roms would be simplified out into two entries, see #235  for reference.

oddMLan:
This also corrects the cutscene timings which were a bit off before

AmbientMalice:
Good find.

Issue #1028, opened 2016-03-14
[RSP] Automobili Lamborghini crashes


purplemarshmallow:
With the PJ64 RSP and LLE gfx it's not possible to get ingame. It crashes with SP DMA WRITE error.
It's the same problem as #1021 and #864 , removing the hack fixes the crash.

correct with cxd4's RSP plugin

Issue #1021, opened 2016-03-04
[RSP] Goldeneye wrong blood


purplemarshmallow:
With the PJ64 RSP the blood in Goldeneye's intro is wrong. RSP Recompiler/Interpreter makes no difference. Correct with cxd4 RSP
![1](https://cloud.githubusercontent.com/assets/7283660/13525704/25dae93a-e203-11e5-9aed-a4803673e958.jpg)

LegendOfDragoon:
This seems to be due to the mfc0 exit implementation.

cxd4:
@purplemarshmallow are you using either one of:  HLE audio or HLE graphics?

purplemarshmallow:
No
LLE gfx, LLE audio

cxd4:
So, if you switch to enable HLE audio does that affect the issue at all?  As weird as that sounds.

purplemarshmallow:
Looks different but still incorrect
![1](https://cloud.githubusercontent.com/assets/7283660/13536323/81e7d008-e23f-11e5-8e09-1c05be99ce71.jpg)

cxd4:
And, if you can compile with this hack removed:
https://github.com/project64/project64/blob/master/Source/RSP/Interpreter%20Ops.c#L313

Change it to some sort of false condition, like `if (0)` or whatever.

Does the modified RSP with the task reset hack removed still have the glitch, with audio HLE?

purplemarshmallow:
I removed the hack and now the glitch is gone (audio HLE, RSP interpreter)

cxd4:
I suppose I could PR my preferred implementation of task short-circuiting over to @project64's RSP, but in principle I have not really decided whether my method is worse or better.

I reconstructed my implementation for more technical problems I was facing than bugs in games.

LegendOfDragoon:
I definitely think your method is better overall. I myself just write game specific code to deal with games that require this mfco exit code to run.

cxd4:
It technically could be that his method is better, depending on how similarly other titles outside of Boss Games may also be particular about timing in response to signal and semaphore changes.  My method has the disadvantage of not doing this subtle (even where sometimes invisible) corrections to the order of updates made to the state machine.

Issue #864, opened 2016-01-07
Super Mario 64 LLE gfx issues with PJ64 RSP


purplemarshmallow:
happens with both PJ64 RSP interpreter and recompiler. z64gl and GLideN64 LLE work correctly with cxd4's interpreter

z64gl
![1](https://cloud.githubusercontent.com/assets/7283660/12171902/afe9ae78-b54d-11e5-9513-d042724c5412.jpg)

GLideN64 LLE
![gliden64_super_mario_64_001](https://cloud.githubusercontent.com/assets/7283660/12171906/bbf4d4b8-b54d-11e5-9fc6-3972a77edbaa.jpg)

LegendOfDragoon:
I cannot reproduce the issue with z64gl. What does it look like, with Angrylion's?

purplemarshmallow:
No angrylion's plugin looks ok. Jabo's plugin also has problems. Some geometry is flickering. I compiled the latest RSP from git. No problems with cxd4's interpreter.
![super mario 64 u snap0000](https://cloud.githubusercontent.com/assets/7283660/12185121/71de8418-b599-11e5-953b-2f01d1e8cc2c.jpg)

LegendOfDragoon:
Can you upload the binary for the RSP?

purplemarshmallow:
https://drive.google.com/file/d/0B7Y6r4SpC_QQVllYODk0TjBEWms/view?usp=sharing

LegendOfDragoon:
I'm still not able to reproduce the issue. Hopefully we can hear feedback from others.

LegendOfDragoon:
I'm wondering if this is a core issue. Can you try testing with cpu interpreter?

purplemarshmallow:
cpu interpreter does not help. Still the same bug

vgturtle127:
@purplemarshmallow @LegendOfDragoon I knew about the rest of those problems, except for the fact that LLE graphics were broken in Super Mario 64.

Let me test it today with the latest build and see what I can come up with.

Have both of you tried modifying the per-game settings (counter factor, etc.)?

purplemarshmallow:
I'm testing with the default settings.
Yes changing the counter factor from 2 to 1 fixes the problem. So it's a timing realted bug

LegendOfDragoon:
Weird, I still couldn't reproduce the issue on counter factor 2. But I just realized that I haven't tried the latest exe. I did test with latest rsp though.

[...]

PR #1033, opened 2016-03-20
introduce new per-game RSP settings

### LINKED ISSUE #139: RSP Recompiler has issues with Rogue Squadron and Battle For Naboo
Basically those 2 games won't run with LLE gfx enabled. It has something to do with CPU<->RSP sync. Raising the count to a high number like 0x4000, allows the game to run. I will spend some time investigating it today. I highly doubt it's an opcode issue, unless it's a branching problem.

It would help a lot, if I knew exactly what PJ64 does for CPU<->RSP sync.

purplemarshmallow:
This fixes #139 #1028 #1021 #864 and other RSP issues and also improves RSP performance

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -2469,7 +2469,7 @@ Good Name=Indiana Jones and the Infernal Machine (E) (Unreleased)
 Internal Name=Indiana Jones
 Status=Only intro/part OK
 Core Note=Camera issue; can't play
-Plugin Note=[rsp] interpreter only [video] HLE not supported
+Plugin Note=[Glide64] errors:various
 32bit=No
 AudioResetOnLoad=Yes
 Counter Factor=1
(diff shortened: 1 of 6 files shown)


LegendOfDragoon:
I have a few suggestions. For the semaphore exit, it's better to do the check at line 1705 instead of https://github.com/purplemarshmallow/project64/blob/RSP_Settings/Source/RSP/Recompiler%20Ops.c#L1710
so line 1705 can be ```if (AudioHle || GraphicsHle || SemaphoreExit == 0)```

Also, for https://github.com/purplemarshmallow/project64/blob/RSP_Settings/Source/RSP/Recompiler%20Ops.c#L1686 , it's supposed to be ```JbLabel8("label", 10);```  Only CompConstToX86reg should use Mfc0Count.

LegendOfDragoon:
Why does Battle Phoenix need RSP-Mfc0Count=10? Are you sure it's necessary? Stunt Racer 64 needs RSP-Mfc0Count=10 and so does Pokemon Stadium 2 for GB Tower.

purplemarshmallow:
All fixed, I also removed the interpreter only RDB entries. I think this can be merged.

I never added RSP-Mfc0Count=10 to Battle Phoenix. It might be misleading, github previews are not really helpful with rdb changes

PR #1038, opened 2016-03-21
Port @Gonetz Bomberman 2 blend mode to Glide64.

AmbientMalice:
I haven't tested this personally, since I can't load the save states because PJ64 has broken save state loading. But it should work. I personally want to see Glide64 replaced as soon as possible, but there's no harm porting such an obvious bug fix.


LegendOfDragoon:
Save States should work if you created it with a recent build.

AmbientMalice:
>Save States should work if you created it with a recent build.

That's the problem. The save states were made with 2.2, and I don't know enough about Bomberman to find the affected area on my own.

PR #1051, opened 2016-03-27
[Glitch64] compiles now on Linux w/o warnings or errors

cxd4:
Let me know if I did the first commit incorrectly.  I was confused by this particular case of `#ifdef WIN32` in that it appears to have predominantly placed there by the original authors before zilmar even forked the plugin, but also in that it applies whether to name height g_height or just height.

The second commit...I imagine I did not do incorrectly.  It just restores the `#include <SDL.h>` that zilmar removed for WIN32 builds and for ANDROID builds.  I am more than able to remove the Linux dependency on SDL in this plugin with context creation (not the gamma function though...not sure what's up with that), but I figure this is faster as a preface to the gradual process of removing it for there as well.  Besides, this way it can get the plugin compiling for Macintosh or whatever as well.

### DIFF
--- a/Source/Glitch64/OGLglitchmain.cpp
+++ b/Source/Glitch64/OGLglitchmain.cpp
@@ -1126,7 +1126,7 @@ int                  nAuxBuffers)
         glCompressedTexImage2DARB = (PFNGLCOMPRESSEDTEXIMAGE2DPROC)dummy_glCompressedTexImage2D;
 #endif
 
-#ifdef _WIN32
+#ifndef ANDROID
     glViewport(0, viewport_offset, g_width, g_height);
     viewport_width = g_width;
     viewport_height = g_height;
--- a/Source/Glitch64/OGLglitchmain.cpp
+++ b/Source/Glitch64/OGLglitchmain.cpp
@@ -29,6 +29,21 @@
  */
 #include <Settings/Settings.h>
 
+#if !defined(_WIN32) && !defined(__ANDROID__) && !defined(ANDROID)
+/*
+ * Do not include <SDL.h> for Windows or Android builds.
+ * Most other platforms tend to include it by default, though.
+ *
+ * Just a few dependencies to SDL remain in the code:
+ *     1.  SDL_SetGammaRamp(), in CorrectGamma()
+ *     2.  SDL_SetGammaRamp(), in grGetGammaTableExt()
+ *     3.  SDL_Surface* m_pScreen = SDL_SetVideoMode(...), to create context
+ * Context creation can be done fine without SDL, and eventually it will be
+ * nice to remove it.  However, Mac & modern Linux tends to come with SDL.
+ */
+#include <SDL.h>
+#endif
+
 struct ResolutionInfo
 {
     unsigned int dwW, dwH, dwF;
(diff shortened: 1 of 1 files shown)


Source/Glitch64/OGLglitchmain.cpp
```diff
@@ -29,6 +29,21 @@
...
+ * Most other platforms tend to include it by default, though.
+ *
+ * Just a few dependencies to SDL remain in the code:
+ *     1.  SDL_SetGammaRamp(), in CorrectGamma()
+ *     2.  SDL_SetGammaRamp(), in grGetGammaTableExt()
+ *     3.  SDL_Surface* m_pScreen = SDL_SetVideoMode(...), to create context
+ * Context creation can be done fine without SDL, and eventually it will be
+ * nice to remove it.  However, Mac & modern Linux tends to come with SDL.
+ */
+#include <SDL.h>
```
project64: what exactly do you need sdl for?

cxd4: You have remaining dependencies to SDL in your fork of Glitch64 at:

* [here](https://github.com/project64/project64/blob/0d3d42c7f9cd335f32279d71560d9983054d23f2/Source/Glitch64/OGLglitchmain.cpp#L2803),
* [here](https://github.com/project64/project64/blob/0d3d42c7f9cd335f32279d71560d9983054d23f2/Source/Glitch64/OGLglitchmain.cpp#L2838),
* and ... [here](https://github.com/project64/project64/blob/e5c6fa245bdb5a67c9192d18e7f45acce72de700/Source/Glitch64/OGLglitchmain.cpp#L1041-L1109) (which later you deleted, which just does not work)

Does that answer your question?

project64: it does but I would prefer to remove the SDL functionality then add SDL back in unless there is something that can not be done for it. Cause your calling functions there on SDL, but has it even been initialized or being used at this point.

cxd4: I understand neither of us wants SDL.  I don't either; maybe you fear I'm trying to restore it.

I was not adding SDL functionality back in, so much as just trying to fix the compile errors as directly as possible.  I don't mean to call or add any SDL functions, but removing them is too immediate of an approach to me when I can just debug the smaller code it after the SDL works.

Putting this into perspective--think back to when you helped port Project64 to 64-bit x86.  You didn't _immediately_ delete/replace all of the inline assembly and other features causing 64-bit Project64 to not build; you used breakpoints or macro'd them out.  If you immediately tried replacing the 32-bit x86 inline assembly with x64 code you would have no way to debug it without getting it to compile first.

So in this case the quickest way to get it to compile is include SDL.h, and then remove SDL.

So I can take the long road and immediately replace your leftover SDL lines, but this will:
a) Not be debuggable until after everything compiles and starts to run here on Linux.
b) Take longer.  I can replace the SDL with some extra time but it will be a while before I can have this PR updated.

cxd4: I suppose c) not work with Macintosh, I have no idea how to initialize OpenGL on a Mac.

But my intention was not to get SDL working, just shut up the errors and then remove the SDL.

cxd4:
@project64 I just have added a couple more commits now.

If you check the diff it should look slightly less confusing.  There's no SDL include this time, but I'd still like to postpone implementing the non-SDL solution for when the plugin actually runs so it can be debugged.

PR #1056, opened 2016-04-07
[PropertySheets] Omit SmallerTypeCheck MSVC setting.

JunielKatarn:
The latest update to MSVC deprecates the `/RTCc` switch, labeling as non standards-conformant:
```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\yvals.h(71): error C2338: /RTCc rejects conformant code, so it isn't supported by the C++ Standard Library. Either remove this compiler option, or define _ALLOW_RTCc_IN_STL to
 acknowledge that you have received this warning. [D:\Extra\pj64\jk\Source\Common\Common.vcxproj]
```

This switch was only enabled on the `Debug` configuration anyways, and didn't add any value to the debugging experience or performance.

### DIFF
--- a/PropertySheets/Debug.props
+++ b/PropertySheets/Debug.props
@@ -11,7 +11,6 @@
       <EnableFiberSafeOptimizations>false</EnableFiberSafeOptimizations>
       <WholeProgramOptimization>false</WholeProgramOptimization>
       <BasicRuntimeChecks>EnableFastChecks</BasicRuntimeChecks>
-      <SmallerTypeCheck>true</SmallerTypeCheck>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
     </ClCompile>
     <Midl>
--- a/PropertySheets/Debug.props
+++ b/PropertySheets/Debug.props
@@ -11,6 +11,7 @@
       <EnableFiberSafeOptimizations>false</EnableFiberSafeOptimizations>
       <WholeProgramOptimization>false</WholeProgramOptimization>
       <BasicRuntimeChecks>EnableFastChecks</BasicRuntimeChecks>
+      <SmallerTypeCheck Condition="'$(PlatformToolsetVersion)'&lt;140">true</SmallerTypeCheck>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
     </ClCompile>
     <Midl>
--- a/PropertySheets/Platform.props
+++ b/PropertySheets/Platform.props
@@ -61,11 +61,12 @@
       <StringPooling>true</StringPooling>
       <MinimalRebuild>true</MinimalRebuild>
       <ExceptionHandling>Async</ExceptionHandling>
+      <BasicRuntimeChecks>Default</BasicRuntimeChecks>
       <StructMemberAlignment>Default</StructMemberAlignment>
       <BufferSecurityCheck>false</BufferSecurityCheck>
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <EnableEnhancedInstructionSet Condition="'$(Platform)'=='Win32'">NoExtensions</EnableEnhancedInstructionSet>
-      <EnableEnhancedInstructionSet Condition="'$(Platform)'=='Win32' AND ('$(PlatformToolset)'=='v90' OR '$(PlatformToolset)'=='v100')">NotSet</EnableEnhancedInstructionSet>
+      <EnableEnhancedInstructionSet Condition="'$(Platform)'=='Win32' AND $(PlatformToolsetVersion)&lt;110">NotSet</EnableEnhancedInstructionSet>
       <FloatingPointModel>Precise</FloatingPointModel>  
       <FloatingPointExceptions>false</FloatingPointExceptions>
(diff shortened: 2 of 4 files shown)


project64:
any particular reason to remove it instead of fixing the code that causes the check ?

JunielKatarn:
Yes. According to the error documentation, this check (again, not even used in Release) rejects code existing in the standard library. I think we want the compiler to accept standard library code and, worst case, reject non-standards only.

While I didn't thoroughly search for the error, if you see the message I posted above, the line triggering such error is the following:
`C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\yvals.h(71)`, which is not something we can modify.

JunielKatarn:
More insight here:
https://www.reddit.com/r/cpp/comments/46mhne/rtcc_rejects_conformant_code_with_visual_c_2015/

cxd4:
Though it's not supposed to be used in Release.  It's better-suited to Debug only, as that's what it's for.

>According to the error documentation, this check (again, not even used in Release) rejects code existing in the standard library. I think we want the compiler to accept standard library code and, worst case, reject non-standards only.

Having just done a little research through links like what you pasted, I am seeing that this is relevant to the C++ standard *template* library but not the C standard library.  So the fix here would be to not involve C++ language dependencies and just use C rules.

JunielKatarn:
> the fix here would be to not involve C++ language dependencies and just use C rules
I believe it's easier to disregard the switch than to port all the code to plain C.
What's wrong with using the standard C++ library?

As a side note, "template" is a deprecated denomination.

Do you guys know of any use case that would make us miss the /RTCc switch while debugging?

project64:
what is the case where this is actually causing you an issue at the moment?

the case that would make us miss the /RTCc switch while debugging is where new code is added that is overflowing cause it writing to a smaller container. It is to help detect problems before they are ever checked in.

JunielKatarn:
The latest MSVC compiler update deprecates the switch out of the box.
The line in the header I pointed out literally does a `static_assert(false ...`.
In short, Project64 won't build for anyone using Visual Studio 2015 Update2 and anything that is released later.

To meet you halfway, I'll make my change exclusive to the affected MSVC version.

cxd4:
>I believe it's easier to disregard the switch than to port all the code to plain C.

Yes it is _easier_ to disregard it, but your issue is you want to _remove_ it, which isn't really the same. :)

>What's wrong with using the standard C++ library?

Like you addressed, it's not really a C++ linguistic issue, just with the standard templates library.

JunielKatarn:
@cxd4 I'll rephrase the last question.
What's wrong with relying on the C++ Standard Library?
Forcing C rules on a C++ code base doesn't sound very useful to me.

JunielKatarn:
@project64 I have made my change exclusive to Visual Studio 2015 and later versions (which has the `static_assert(false)` for the `/RTCc` compiler switch.

PropertySheets/Platform.props
```diff
@@ -61,8 +61,6 @@
       <StringPooling>true</StringPooling>
       <MinimalRebuild>true</MinimalRebuild>
       <ExceptionHandling>Async</ExceptionHandling>
-      <BasicRuntimeChecks>Default</BasicRuntimeChecks>
```
project64: please add the condition instead of removing it.

JunielKatarn: It's actually not removed. Please see both Debug.props and Release.props.
It is defined on both files with different values.
Its presence on Platform.props was redundant and always overwritten.

PropertySheets/Release.props
```diff
@@ -2,15 +2,15 @@
...
       <Optimization>MinSpace</Optimization>
       <InlineFunctionExpansion>Default</InlineFunctionExpansion>
       <IntrinsicFunctions>false</IntrinsicFunctions>
       <FavorSizeOrSpeed>Size</FavorSizeOrSpeed>
       <OmitFramePointers>false</OmitFramePointers>
       <EnableFiberSafeOptimizations>false</EnableFiberSafeOptimizations>
       <WholeProgramOptimization>false</WholeProgramOptimization>
-      <PreprocessorDefinitions>NDEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <SmallerTypeCheck>false</SmallerTypeCheck>
+      <BasicRuntimeChecks>Default</BasicRuntimeChecks>
```
JunielKatarn: @project64 Added it here.

PropertySheets/Debug.props
```diff
@@ -11,7 +11,7 @@
       <EnableFiberSafeOptimizations>false</EnableFiberSafeOptimizations>
       <WholeProgramOptimization>false</WholeProgramOptimization>
       <BasicRuntimeChecks>EnableFastChecks</BasicRuntimeChecks>
```
JunielKatarn: @project64 It was already here.

cxd4:
It's actually the C standard library for the most part, just with C++ templates included.  Rather, that's one of the smarter decisions concerning C++ as a language:  to make it mostly backwards compatible.

JunielKatarn:
@project64 I updated the PR and added the platform check instead of removing the setting.

PR #1059, opened 2016-04-12
Use sizeof() to eliminate respecification of array bounds.

cxd4:
I really just noticed this so far in `FlashRam.cpp` only.  There might be others to apply this to.

### DIFF
--- a/Source/Project64-core/N64System/Mips/FlashRam.cpp
+++ b/Source/Project64-core/N64System/Mips/FlashRam.cpp
@@ -29,7 +29,7 @@ CFlashram::~CFlashram()
 
 void CFlashram::DmaFromFlashram(uint8_t * dest, int32_t StartOffset, int32_t len)
 {
-    uint8_t FlipBuffer[0x10000];
+    uint8_t FlipBuffer[1 << 16];
     switch (m_FlashFlag)
     {
     case FLASHRAM_MODE_READ:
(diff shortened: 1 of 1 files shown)


Source/Project64-core/N64System/Mips/FlashRam.cpp
```diff
@@ -29,21 +29,21 @@ CFlashram::~CFlashram()
 
 void CFlashram::DmaFromFlashram(uint8_t * dest, int32_t StartOffset, int32_t len)
 {
-    uint8_t FlipBuffer[0x10000];
+    uint8_t FlipBuffer[1 << 16];
```
project64: I prefer 0x10000 over 1 << 16. It might be just me but at a quick looking I have no idea how much is 1 << 16. I have to spend time thinking and trying to calculate it to work out the size, instead of my attention of looking at the code.

The rest looks good tho

cxd4: I think you're right; I only changed it from `0x10000` to `1 << 16` just to make GitHub zealously highlight that line to show you where the array was being declared.

It was more to highlight the safety of the change, not really make a better change.  So scratch that.

cxd4:
Yeah that was just to highlight the scope of the diff to show non-breaking-ness of the change.

It's redacted.  I don't really prefer the `1 << 16` way except for showing address space layout in bits.

PR #1062, opened 2016-04-14
Further simplified property sheets.

JunielKatarn:
Shrinked VCXPROJ files even more.
Now, it is only required to import an entry point `PropertySheets\Project.props` and all default project properties will be resolved.

This also hides MSVC-specific imports from project files, so an eventual multi-platform build gets easier.


vgturtle127:
:+1: Cool!

PR #1066, opened 2016-04-21
Port Common to compile on Unix systems

smlu:
This port uses new c++14 standard as much as it can to compile on Ubuntu, Mac OS X and Android.

No Windows code base was modified, the port was added via #if #else macros to not break current codebase or create compiler issue with older msvc compilers. The source code in classes `CriticalSection`, `SyncEvent` and some parts of class `CPath` could be merget to use common code across different  platforms. `CThread` uses pthread library on Unix port but it can be ported to use std::thread.

The code was tested and it works as expected. The only problem was found with `CPath::DirectoryCreate()` method, when relative path is used no new directory is created. This is because method `GetDriveDirectory()` returns empty path. This is expected?


JunielKatarn:
The latest release of MSVC is C++14-compliant. I'll take a look at the code, but maybe you don't need to if-def out all of it.

Source/Common/CriticalSection.cpp
```diff
@@ -1,16 +1,28 @@
...
 
 CriticalSection::CriticalSection()
 {
-	m_cs = new CRITICAL_SECTION;
-	::InitializeCriticalSection((CRITICAL_SECTION *)m_cs);
+#ifdef _WIN32
+    m_cs = new CRITICAL_SECTION;
+    ::InitializeCriticalSection((CRITICAL_SECTION *)m_cs);
+#else
+    m_cs = static_cast<void*>(new std::mutex);
```
JunielKatarn: For example: here.
Visual Studio 2014 (at least that version) can do static_cast.

project64:
I will have a look and see how it compares to what I have done so far to make common more linux compatible.

CriticalSection.cpp - https://gist.github.com/project64/7c592e50d0fdb9a6cc6acd9de86b545c
SyncEvent.cpp - https://gist.github.com/project64/b1bd33cc5f157ee2c10b2ea2f9ff8fd2
path.cpp - https://gist.github.com/project64/118c04a230a289f34cd320d16f3c6a91
MemoryManagement.cpp - https://gist.github.com/project64/d8e4967bf4ba9bac923e7ffd38a5a95c

project64:
I also need the code to be able to compile on Android NDK

smlu:
I've fixed Android port  and I should work now. I've also added Makefile and Makefile-Android scripts to simplify compilation process. There is also `test`  exe to test this port on Unix. For now it only tests multi-threading and part of `CPath` class. 

In IMO the design of classes `Cthread`, `SyncEvent` and `CriticalSection` should be redesigned to use c++ standard classes. By that I mean that they should be implemented to use `std::thread`, `std::std::condition_variable` and `std::mutex` instead of native code. Less code to handle later on.

Source/Common/CriticalSection.cpp
smlu: I will fix that when (if) this commit is merged and tested

project64:
I will look at starting to commit my changes to common then look at if I can use the code you have put here.

project64:
The common should now be able to compile on linux now. Is there any changes you think should be added?

PR #1071, opened 2016-04-28
Updated Spanish Translation.

MELERIX:
for: https://github.com/project64/project64/commit/02d7ac014e17df0fd1b928393a28f5ae456c5f32

### DIFF
--- a/Lang/Spanish.pj.Lang
+++ b/Lang/Spanish.pj.Lang
@@ -234,6 +234,7 @@
 #465#  "Mostrar velocidad"
 #466#  "Velocidad de la pantalla:"
 #467#  "Comprobar si Project64 ya está en ejecución"
+#468#  "Directorio Único de Guardados de Juego"
 
 //ROM Browser Tab
 #480#  "Máx # de ROMs recordadas (0-10):"


Lang/Spanish.pj.Lang
```diff
@@ -234,6 +234,7 @@
 #465#  "Mostrar velocidad"
 #466#  "Velocidad de la pantalla:"
 #467#  "Comprobar si Project64 ya está en ejecución"
+#468#  "Directorio Único de Guardados de Juego"
```
JunielKatarn: I would add a noun instead. "Guardados" is mostly used as an adjective.

Suggestion: Analogously to "save states" in English, you could go for "estados guardados".

Also, as opposed to English, Spanish does not capitalize words as much ("Únicos" instead of "únicos", "guardados" instead of "Guardados", "juego" instead of "Juego").

MELERIX: ok, but this is used for "save states" or is for native "save games" aka nvram ?

JunielKatarn: Not sure. This diff doesn't show the English version for entry 468.
Doesn't matter, though. It's merged. It can be improved in a future PR, if needed.

PR #1082, opened 2016-05-08
[GlideHQ] undefined symbol:  fxt1_encode()

cxd4:
I'm assuming there must have been some reason why the `extern "C"` was there.

Somebody on Windows try testing this commit before it gets merged, because I don't know if it will cause any linkage to break on there or not.  All I know is right now, with GNU it solves this run-time link fault:
```
bash-4.2$ ./mupen64
Couldn't load plugin './plugins/PJ64Glide64.so': ./plugins/PJ64Glide64.so: undefined symbol: fxt1_encode
```

Again, maybe it fixes for one platform and breaks on another.  This C-to-C++ namespace mess is rather hard to do a time-efficient investigation of, so if anyone understands a better solution, do say.

### DIFF
--- a/Source/GlideHQ/TxUtil.h
+++ b/Source/GlideHQ/TxUtil.h
@@ -31,9 +31,6 @@
 #include <string>
 
 #ifndef DXTN_DLL
-#ifdef __cplusplus
-extern "C"{
-#endif
 void tx_compress_dxtn(int srccomps, int width, int height,
                       const void *source, int destformat, void *dest,
                       int destRowStride);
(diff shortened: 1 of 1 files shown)


project64:
fxt1_encode is here https://github.com/project64/project64/blob/master/Source/GlideHQ/tc-1.1%2B/fxt1.c#L1127

since it is in a .c file it has c naming style, the extern "c" basically means do not use c naming formatting, but use C style.

so with out the lines you removed it will not be able to link, unless the file get treated as a c++ file

JunielKatarn:
MSVC projects can be set to treat all source files as C++ by default.
Does Pj64 need any files to be treated as C-only?

cxd4:
@project64 I would then like to close this PR, except, I have no current ideas to replace it with.

What other fixes are there?  Rename the .c files to .cpp ones?  Use some kind of extern "C++" feature (so to speak, not sure if one even exists), or add an extern "C" in one of the prototypes where it should have been added but was forgotten?

project64:
I generally try removing c in general, probably best to rename the c files to cpp files, but that needs to be tested.

cxd4:
https://github.com/project64/project64/pull/1083

PR #1083, opened 2016-05-09
[GlideHQ] fixed undefined symbol errors caused by C-C++ mismatch

cxd4:
This branch may or may not need relevant testing by somebody running Microsoft Windows.


JunielKatarn:
I test-built. It fails with the following linkage error:
```
error LNK2019: unresolved external symbol "void __cdecl tx_compress_dxtn
```

cxd4:
Something definitely is wrong here, yeah.

With the original `TxUtil.h`:
```c
extern "C" {
void tx_compress_dxtn(int srccomps, int width, int height, const void *source, int destformat, void *dest, int destRowStride);
int fxt1_encode(int width, int height, int comps, const void *source, int srcRowStride, void *dest, int destRowStride);
}
```
Error:  `./plugins/PJ64Glide64.so: undefined symbol: fxt1_encode`

If I change `TxUtil.h` to only extern 1 of them:

```c
extern "C" {
void tx_compress_dxtn(int srccomps, int width, int height, const void *source, int destformat, void *dest, int destRowStride);
}
int fxt1_encode(int width, int height, int comps, const void *source, int srcRowStride, void *dest, int destRowStride);
```
Error:  `./plugins/PJ64Glide64.so: undefined symbol: tx_compress_dxtn`

If I remove that extern and move it to the other function instead:

```c
extern "C" {
void tx_compress_dxtn(int srccomps, int width, int height, const void *source, int destformat, void *dest, int destRowStride);
}
int fxt1_encode(int width, int height, int comps, const void *source, int srcRowStride, void *dest, int destRowStride);
```
Error:  `./plugins/PJ64Glide64.so: undefined symbol: _Z16tx_compress_dxtniiiPKviPvi`

Finally, just removing both extern C's gives the same exact error as the case above.

cxd4:
https://github.com/project64/project64/pull/1084

cxd4:
This was a waste of time.

Unlike as with past cases, the whole problem here instead was a static fold of standard C++ dependency code being injected due to invoking the GNU C++ toolchan, `g++`, instead of cc / gcc.

PR #1085, opened 2016-05-09
[GlideHQ] fixed -Woverflow large integer conversion warnings

cxd4:
The `cc` is of type `dword *`, where `types.h` has that typedef'd to `unsigned int`.

Anyway, the problem here was that `~0UL` could be `0xFFFFFFFFFFFFFFFF` instead of just `0xFFFFFFFF`, when it's clear that the target is meant to store only 32-bit elements.  Saying `0xFFFFFFFFul` instead of `~0ul` prevents interpreting values of the (unsigned long) data type which on other ABIs would exceed the limit of what would fit into the (unsigned int) type.

```
Compiling GlideHQ library sources for Glide64...
./../../GlideHQ/tc-1.1+/fxt1.c: In function 'fxt1_quantize_MIXED1':
./../../GlideHQ/tc-1.1+/fxt1.c:758:2: warning: large integer implicitly truncated to unsigned type [-Woverflow]
  cc[0] = ~0UL;
  ^
./../../GlideHQ/tc-1.1+/fxt1.c:792:2: warning: large integer implicitly truncated to unsigned type [-Woverflow]
  cc[1] = ~0UL;
  ^
./../../GlideHQ/tc-1.1+/fxt1.c: In function 'fxt1_quantize':
./../../GlideHQ/tc-1.1+/fxt1.c:1112:2: warning: large integer implicitly truncated to unsigned type [-Woverflow]
  cc[0] = cc[1] = cc[2] = ~0UL;
  ^
Assembling GlideHQ library sources...
```


cxd4:
Oh, and don't merge because a bunch of unrelated files also got staged with it. :) Time to sleep.

PR #1086, opened 2016-05-10
[GlideHQ] fixed -Woverflow large integer conversion warnings

cxd4:
https://github.com/project64/project64/pull/1085

Now with a pinch of human sleep added.

### DIFF
--- a/Source/GlideHQ/tc-1.1+/fxt1.c
+++ b/Source/GlideHQ/tc-1.1+/fxt1.c
@@ -755,7 +755,7 @@ fxt1_quantize_MIXED1 (dword *cc,
     /* left microtile */
     if (maxColL == -1) {
 	/* all transparent black */
-	cc[0] = ~0UL;
+	cc[0] = 0xFFFFFFFFUL;
 	for (i = 0; i < n_comp; i++) {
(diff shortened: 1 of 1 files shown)


Source/GlideHQ/tc-1.1+/fxt1.c
```diff
@@ -755,7 +755,7 @@ fxt1_quantize_MIXED1 (dword *cc,
     /* left microtile */
     if (maxColL == -1) {
 	/* all transparent black */
-	cc[0] = ~0UL;
+	cc[0] = 0xFFFFFFFFUL;
```
JunielKatarn: Does this actually change the value that will be set to `cc[0]`, or is it only a style thing?

cxd4: It's a warning fix.  C warnings are not generally about style.

As explained in the #1085 copy of this pull request:
```
~0UL on 32-bit:  0000 0000 FFFF FFFF
~0UL on LP64 :  FFFF FFFF FFFF FFFF
```

As `cc[0]` is of type `typedef unsigned int dword;`, ~0UL risks overflow.

PR #1095, opened 2016-05-23
[64DD] 64DD disk conversion fix

LuigiBlood:
Fixes Randnet DD

### DIFF
--- a/Source/Project64-core/N64System/N64DiskClass.cpp
+++ b/Source/Project64-core/N64System/N64DiskClass.cpp
@@ -290,7 +290,7 @@ void CN64Disk::ConvertDiskFormat()
         atrack = 0;
         for (track = 0; track < ZoneTracks[zone]; track++)
         {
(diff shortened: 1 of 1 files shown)


Source/Project64-core/N64System/N64DiskClass.cpp
```diff
@@ -290,7 +290,7 @@ void CN64Disk::ConvertDiskFormat()
         atrack = 0;
         for (track = 0; track < ZoneTracks[zone]; track++)
         {
-            if (track == SystemData[0x20 + zone * 0xC + atrack])
+            if (atrack < 0xC && track == SystemData[0x20 + zone * 0xC + atrack])
```
JunielKatarn: Out of curiosity, what does '0xC' imply?
Would it be worth to create a constant to name the value meaningfuly?

LuigiBlood: Amount of alternate tracks. Perhaps a constant would be useful but we'll see that later when we'll do development disk support.

PR #1103, opened 2016-06-03
Disable 32-bit Engine for NBA Live 2000 and NFL Blitz 2000

LegendOfDragoon:


pcca-matrix:
Must be disabled for NBA In the Zone 2000  also.
without it crash with Unhandled R4300i opcode at: 80023994

PR #1105, opened 2016-06-03
[UI-Settings] Add 64DD IPL ROM path to General Settings UI

LuigiBlood:
Also Enable Disk setting is now temporary. It's useless to have it saved.


LuigiBlood:
The new setting in action:
image

project64:
I like it, but wouldn't this more be in advanced settings page?

cxd4:
LOL, yeah.

LuigiBlood:
I don't think it's a good idea to put it in Advanced Settings, because it isn't very advanced. I'm not a fan of putting this in a page that's hidden by default, also.

cxd4:
Who wouldn't?

Good luck finding users on here who even know what "64DD" means.  Seeing 64DD stuff right smack in front of their face when they go to open Project64's generic settings just looks confusing and random.

Now, if the emulator were titled "Project64DD", **then** people might expect advanced things like DD support to be right on the front page of everything.

I'm sure you could always create a new page for it:  64DD or Advanced/64DD.

LuigiBlood:
Making a new page just for one option (and perhaps another for saving which does make sense on the Advanced page) would be a bit of a mess.

cxd4:
Less of a mess than putting it on the front page and scaring the shit out of all the little kids who've never heard of a DD.

In fact all other 64DD options added in the future would belong there.

LuigiBlood:
...Well for now I'll put it on Advanced and we'll see about a 64DD page later.

LuigiBlood:
Currently how it looks like now:
image

cxd4:
Looks better to me.

Yes I'm sure as new 64DD settings are added a new page can progressively be more appropriate.  I don't really care about it right now.  Advanced is fine atm.

PR #1107, opened 2016-06-04
Begin compiling past most fatal errors immediately terminating compilers.

cxd4:


project64:
Are you trying to compile the projec64 project?

it is only meant for windows

cxd4:
Your -core module compiles as a static library and needs an entry point from something like the `Project64` (UI) project to interface with it.  You have this on Windows but not other platforms.

If not that, then are you proposing I create a new UI folder separate from the `Project64` one for running the program on non-Windows platforms?  Or does your reply entail leaving Project64 Windows- and Android-only with no other ports?

project64:
Each OS will need to provide a native UI. You could do a simple command line UI to get started. Something like https://gist.github.com/project64/5fc3f0a93c22ef6b9160883a31ec88a6

PR #1118, opened 2016-06-06
Fix Project64.sln mappings for Android projects.

JunielKatarn:
Win32 is failing to build, and x64 is not correctly mapped.
All config,platform combinations work fine now.


project64:
the projects could just have been disabled for 64bit as I doubt they will be ever used

PR #1127, opened 2016-06-25
[Common] Make WriteTrace() #define more stable #ifdef _DEBUG.

cxd4:
With the high traffic in Project64's interface over the core, the state of my current working tree with local modifications to silently ignore/bypass each one just to wait until finding out what the real cause is has become rather overtaxing.  It becomes too difficult to keep local changes out of even my own master.

So why not have the macro fixed for _DEBUG compiles?

### DIFF
--- a/Source/Common/Trace.h
+++ b/Source/Common/Trace.h
@@ -37,9 +37,19 @@ private:
 };
 
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


cxd4:
closed in favor of https://github.com/project64/project64/pull/1133

PR #1150, opened 2016-07-09
Remove warnings due to variable re-declarations.

JunielKatarn:
Less warnings, happier builds.

### DIFF
--- a/Source/Common/IniFileClass.cpp
+++ b/Source/Common/IniFileClass.cpp
@@ -54,24 +54,24 @@ void CIniFileBase::fInsertSpaces(int Pos, int NoOfSpaces)
     if (NoOfSpaces < 0)
     {
         int ReadPos = Pos + (NoOfSpaces * -1);
-        int WritePos = Pos;
+        int localWritePos = Pos;
 
         do {
             SizeToRead = end - ReadPos;
             if (SizeToRead > fIS_MvSize) { SizeToRead = fIS_MvSize; }
             m_File.Seek(ReadPos, CFileBase::begin);
             m_File.Read(Data, SizeToRead);
-            m_File.Seek(WritePos, CFileBase::begin);
+            m_File.Seek(localWritePos, CFileBase::begin);
             m_File.Write(Data, SizeToRead);
             ReadPos += SizeToRead;
-            WritePos += SizeToRead;
+            localWritePos += SizeToRead;
         } while (SizeToRead > 0);
 
-        m_File.Seek(WritePos, CFileBase::begin);
+        m_File.Seek(localWritePos, CFileBase::begin);
         stdstr_f SpaceBuffer("%*c", (NoOfSpaces * -1), ' ');
         m_File.Write(SpaceBuffer.c_str(), (uint32_t)SpaceBuffer.length());
 
-        m_File.Seek(WritePos, CFileBase::begin);
+        m_File.Seek(localWritePos, CFileBase::begin);
         m_File.SetEndOfFile();
         m_File.Seek(0, CFileBase::begin);
     }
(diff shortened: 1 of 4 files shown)


cxd4:
Those are _"shadowing"_ declarations, not re-declarations.

Unlike as in the case of a couple changes (which here, again, are not on-topic to the idea of the PR and would belong to their own commit), they are classified under a very, very high level of warning that not even the modern `cc` with `-Wall` warns about anymore.  (You would have to pass `-Wshadow`.)  They do not at all "re-declare" anything and basically just remove the ability to access global variables with the same name (which zilmar tries to classify under a convention of removing/moving global variables to either inside a struct or prefixed with `g_*`).

If MSVC is calling them re-declarations, it is quite wrong; they are two separate, unchanged decl's.

JunielKatarn:
OK, vocabulary aside, I believe it's better to not overlap variable naming within the same context, accurate warnings or not.
Code will be easier to understand if developers don't need to expect a variable to "seem to mean two things" in the same scope.

cxd4:
It appears as though you may(?) be implying they're in the same scope or context.  They are not.

A situation like
```c
int variable;

int
main(void)
{
    int variable;

    variable = 5;
    return 0;
}
```
... evidently has `variable` declared in two separate scopes, with no similarity of context.  You may make the case that the context is similar if they are both named inside of a function (which rarely enough does happen, as in the case of one or two of the changes in your commit).  The vast majority of shadowing however is with clear, visible local variables shadowing hidden or externally visible global variables (which should not need to be accessed anyway).

In that scenario, I would like to propose that the clear, well-named local variables not be renamed.  Either move the shadowed global variables to a class or a structure, rename them (usually by prefixing with `g_` if you follow the normal convention @project64 is slowly integrating), or just ignore the warnings in VC2015.  Up to you.  I can't look at the issue myself because I am busy with life and will not be able to commit or PR to this repository for a while now.

JunielKatarn:
Sounds reasonable. However, I'm not 100% sure all those re-declared (or "shadowed", if you want) variables start in the global scope.
I'll take a look and update the PR accordingly.

cxd4:
I don't really "want" anything; I simply presumed that what you want is to not look inept to everyone else by using under-informed descriptions.

Now, if you had
```c
int
func(void)
{
    short variable;
    long variable;
```
...then **that** would be a re-declaration, which surely wouldn't be legal enough to be called a warning.

But as all of the instances replaced in your commit tend to warnings with different variables under a different scope but with a coinciding name, it's quite clear that no re-declarations possibly exist here.

Source/Common/IniFileClass.cpp
```diff
@@ -54,24 +54,24 @@ void CIniFileBase::fInsertSpaces(int Pos, int NoOfSpaces)
     if (NoOfSpaces < 0)
     {
         int ReadPos = Pos + (NoOfSpaces * -1);
-        int WritePos = Pos;
+        int localWritePos = Pos;
```
project64: just remove the int and use the global one

Source/Common/md5.cpp
```diff
@@ -107,15 +107,15 @@ void MD5::update(const uint1 *input, uint4 input_length)
 
 void MD5::update(FILE *file)
 {
-    unsigned char buffer[1024];
+    unsigned char localBuffer[1024];
```
project64: this should stay as buffer, the member variable should be starting with m_ better to change that one then the local one

cxd4:
Pretty much what pj64 said--g_*, m_*, that other stuff I have trouble remembering (and wanting to use).

The gist of where I was getting at is that it's better to rename (better still:  delete) global variables, than it is to rename well-named and securely organized local variables.  This can be debatable in the difference of two local variables in separate scope but sharing the same name, which zilmar commented one approach to.

PR #1153, opened 2016-07-10
Add 0x04600000-0x0460000C to LW.

Nekokabu:
Add 0x04600000-0x0460000C to LW.


project64:
I am not sure if reading the actual values is the same on a reason n64, but most likely is.

PR #1155, opened 2016-07-11
Fix x64 build due to Android project mappings.

JunielKatarn:
Re-introduced this small configuration fix to unblock x64 builds.
(See 49939eae0bf6468626d64354ef7f1c512b0118b1).


Source/Android/Bridge/Bridge.vcxproj
```diff
@@ -5,10 +5,18 @@
       <Configuration>Debug</Configuration>
       <Platform>Win32</Platform>
     </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
```
project64: this project does not need to be in 64bit

Source/Android/PluginInput/PluginInput.vcxproj
```diff
@@ -5,10 +5,18 @@
       <Configuration>Debug</Configuration>
       <Platform>Win32</Platform>
     </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
```
project64: this project does not need to be in 64bit

Source/Android/Bridge/Bridge.vcxproj
JunielKatarn: @project64 

I believe it doesn't hurt to have it (there could eventually be an x64 Android variant).
Plus, without this change, the overall build fails for that platform.

Another alternative would be to fully leave out Android projects on x64 builds.

Which do you think is best?
In any case, we need to submit a change or builds will keep failing.

project64:
how is it failing, I do not get any errors when I compile it

JunielKatarn:
I'm only getting the errors when running from the command line (using `MSBuild`).
It looks like Visual Studio resolves the configuration mappings differently, safely ignoring the Android projects in x64.

MSBuild is relevant for automated and unattended builds (meaning, not needing to open the IDE just to build).
Would it hurt to include the Android plugins in x64? If so, I'll keep researching how to have both IDE and command line builds working.

project64:
what is the command line you using

JunielKatarn:
> what is the command line you using

`MSBuild Project64.sln /p:Configuration=Release /p:Platform=x64`
(Note, this behavior exists at least in VS2012 and later).

JunielKatarn:
@project64 
I have updated the CR. It now skips the plugins on x64, instead of explicitly building them.
Both IDE-based and command line-based builds succeed.

PR #1156, opened 2016-07-11
Add some of the 64DD Registers to SW_Const

Nekokabu:


LuigiBlood:
Wouldn't that mess up this code though?
https://github.com/Nekokabu/project64/blob/5ceea0f2c2138f2aaed8b023a4ef68f9d558f07b/Source/Project64-core/N64System/Recompiler/x86/x86RecompilerOps.cpp#L11385

Nekokabu:
Replaced it, thanks:)

PR #1176, opened 2016-08-19
Fix install directory

Frank-74:
This changes the default install directory to Users\USERNAME\Project64 2.3.

Windows 8, 8,1 and 10 don't have permission for programs to write files in the Program Files directory.

I tried to use OnlyBelowVersion and MiniVersion parameters, but couldn't get it to work for DefaultDirName.  I'm not sure which directory it will point at for XP or 7.  Needs to be tested on Windows 7/XP.

### DIFF
--- a/Source/Installer/Installer.iss
+++ b/Source/Installer/Installer.iss
@@ -5,7 +5,7 @@
 AppId={{BEB5FB69-4080-466F-96C4-F15DF271718B}
 AppName=Project64
 AppVersion={#AppVersion}
-DefaultDirName={pf}\Project64 2.3
+DefaultDirName={userdocs}\..\Project64 2.3
 VersionInfoVersion={#AppVersion}
 OutputDir={#BaseDir}\Bin\{#Configuration}
 OutputBaseFilename=Setup Project64 2.3


Frank-74:
Another more complete (complicated) solution would be to query the OS version and if above Windows 7, install Config folder to appdata.  Not sure how to do this though.

project64:
I know what your getting at and this might be what has to be done, but I dislike it, most programs install in the program files directory.

I changed the code a long time ago that writing to files was never in the root dir and just to sub dirs

in the installer I change the permissions for the sub dirs so they can be written to.
```
Name: "{app}\Config"; Permissions: users-modify
Name: "{app}\Logs"; Permissions: users-modify
Name: "{app}\Save"; Permissions: users-modify
Name: "{app}\Screenshots"; Permissions: users-modify
Name: "{app}\Textures"; Permissions: users-modify
```

maybe instead of **user-modify** it should be **everyone-modify** but I do not have windows 10 to test with

project64:
the main .exe is designed that it can be portable, that you can take the folder and move it any where and will work, having the data folders in appdata folder would either need modification to the .cfg file in the install process to point to the new location, or inline code that checks is if the execution path is inside program files and change the default location of the sub dirs is in program data not relative to the execution path

cxd4:
I always just did C:\Users\(me).

I never use Program Files.  I could maybe understand it if I was administering to a multi-user computer.

Frank-74:
Visual Studio 2008 installs the binaries to Program Files (x86), and all the config/settings files to Users/Documents, on Windows 8.1 (x64) here.

All the Visual Studio 2008 x64 binaries (debugger etc) go to Program Files.

To keep Project64 portable, I thought the Users/name directory was the easiest solution.
I think on XP, this PR should point to C:\Project64 2.3.  As it points to Users\Documents, and goes up one level to the Users directory, on XP one level up from My Documents is the root drive.

cxd4:
Oh I'm sure things like Visual Studio installed themselves to Program Files.

Anything I conscientiously kept/wanted, though, and had a choice, I just put under /Users instead of /Program Files.  If it defaulted to asking me to put under /Program Files and I was afraid of wasting my time with a failed installation or something maybe I tl;dr'd it to /Program Files for those cases as well.

Frank-74:
I've tried copying the project64 folder to C:\, on running Project64.exe, I get the Exception.
capture
Only users directory works.

Frank-74:
I've found the solution.

Permissions: everyone-full
Install to {pf32}\Project64 2.3

It's working.  I'll close this PR and make a new one.

PR #1177, opened 2016-08-19
[Installer] Fix permissions - desktop icon option

Frank-74:
Fixes permissions for Windows 7/8/10.  And correct Program Files (x86) folder when running on 64bit windows.

### DIFF
--- a/Source/Installer/Installer.iss
+++ b/Source/Installer/Installer.iss
@@ -40,6 +40,7 @@ Name: "{app}\Logs"; Permissions: everyone-full
 Name: "{app}\Save"; Permissions: everyone-full
 Name: "{app}\Screenshots"; Permissions: everyone-full
 Name: "{app}\Textures"; Permissions: everyone-full
+Name: "{app}"; Permissions: everyone-full
 
 [Icons]
 Name: "{commondesktop}\Project64"; Filename: "{app}\Project64.exe"; Tasks: desktopicon
--- a/Source/Installer/Installer.iss
+++ b/Source/Installer/Installer.iss
@@ -5,7 +5,7 @@
 AppId={{BEB5FB69-4080-466F-96C4-F15DF271718B}
 AppName=Project64
 AppVersion={#AppVersion}
-DefaultDirName={pf}\Project64 2.3
+DefaultDirName={pf32}\Project64 2.3
 VersionInfoVersion={#AppVersion}
 OutputDir={#BaseDir}\Bin\{#Configuration}
 OutputBaseFilename=Setup Project64 2.3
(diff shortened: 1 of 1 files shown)


Frank-74:
I think Project64 needs to set the security/access permissions directly as well.  So that it can be unzipped to any folder and run.

Been reading up, and maybe need to implement this: https://msdn.microsoft.com/en-us/library/windows/desktop/aa379579(v=vs.85).aspx

project64:
the Project64.exe would have to be run as an administrator for the Project64.exe to  change the permissions.. 

This really just applies if you are in the program files directory so for a zip package does not matter so much

Frank-74:
Not on Windows 8+.  I just downloaded and extracted the v220 zip from the beta forum.  It doesn't run under any directory other than below Users directory.

Even changing permissions for the folder and running Project64.exe as an administrator, it gives an Exception caught in main.cpp, line 99.  Unless under the Users directory.

project64:
if you run as administrator does it work correctly?

i suspect that at startup it should check if it running as admin, in which case change the permissions of the dir, if not then check what the permissions of the directories are and if they are not correct then do a prompt and exit.

Frank-74:
I only have myself on this computer.  I login as an administrator.  I set project64.exe to run as administrator.  I change the folder permissions to full control.  Still doesn't run under any directory outside of Users.

project64:
I mean right click and choose run as administrator

Frank-74:
> I mean right click and choose run as administrator

Doesn't work...

Frank-74:
Still one weird problem.

Project64.exe doesn't run if in "C:\Project64 2.3", but it does run if in, "C:\Project64 2.3\Project64 2.3"
With the zip and the installer.

It needs another directory layer for some strange reason.

Anything like C:\N64\Project64 2.3 or, C:\Games\Project64 2.3 etc. is fine.

Frank-74:
Used to have 
https://github.com/project64/project64/blob/71cc6def7b35f0d533d2b8a108834823834ac288/Source/Project64/main.cpp#L40

Frank-74:
I've found the bug that causes the exception in main.cpp if run from C:\Project64.

https://github.com/project64/project64/commit/b2545c108bf597e8d4d8455381f73c5e365d6c06#diff-49d517f4abbfa5c428ffaea9a380bbfcL454

I added this back and it's now working in C:\folder.

project64:
ok commited, does the code now solve all the issue with permissions, or there is still some cases that need more work?

I am thinking of maybe doing a 2.3.1 with these changes.

Frank-74:
Eveything solved as far as I've tested.

I've extracted the zip to C:\project64, and it works fine now.  I've tested it on a USB stick with no folder, H:\Project64.exe, works fine now.  I've run the installer, fine installed in Program Files, Users, C:\.

2.3.1 release is a good idea.  Surprised this bug went undetected since April.

project64:
not really surprising, most likely cause people were using zip not the installer and if they were building it was up multiple levels.

maybe it could be done easily to add an "install as portable" which just basically would not add the registry options for uninstall.

Frank-74:
There's an example in the inno setup help.

> [Setup]
Uninstallable=not IsTaskSelected('portablemode')
[Tasks]
Name: portablemode; Description: "Portable Mode"

project64:
yes something exactly like that :)

Frank-74:
Need to add IsTaskSelected to this line as well I think.

Name: "{commonprograms}\Project64 2.3\Uninstall Project64 2.3"; Filename: "{uninstallexe}"; Parameters: "/LOG"

I'll give it a go.

Frank-74:
Ok, i've done it.  Virtually just cut and paste from inno help's Uninstallable option.
https://github.com/project64/project64/pull/1182

PR #1181, opened 2016-08-21
Fix up Glide dialog, spellings

Frank-74:


project64:
I think I cut and pasted from the original, but spelling is never my thing

PR #1182, opened 2016-08-22
[Installer] Add 'Portable Mode' option.

Frank-74:
It's unchecked by default, doesn't create an uninstall shortcut icon if selected.


project64:
the Flags: createonlyiffileexists on the uninstall icon, will that mean it is never created on a clean install?

Frank-74:
No, the uninst000.exe is installed when portable mode is unchecked it, so the file exists.  The install copies the files before the icons.

PR #1199, opened 2016-09-02
[RDB] 64DD IPL USA added

LuigiBlood:
Can't make a difference between 64DD IPL (JPN) and 64DD TOOL IPL (JPN) though.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -8157,3 +8157,10 @@ Fixed Audio=0
 Good Name=64DD IPL (JPN)
 RDRAM Size=8
 Status=Compatible
+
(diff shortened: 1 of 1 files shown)


project64:
[00000000-00000000-C:45]

the crc's are 0 ... hmmm ... is the 64dd headers any different from normal 64 roms.. I wonder if there is any other fields I should refer to in the headers.

LuigiBlood:
64DD IPL checksum is part of the IPL3 code instead, a bunch of LUI and ORI instructions and compares with those.
And instead of two 32-bit words for the checksum, there's 6 of them at 0x608, 0x618, 0x628, 0x638, 0x648, 0x658.

PR #1201, opened 2016-09-07
Jabo_Direct3D8 no d3d syslog

Frank-74:
Patched to remove logging, which can use a lot of memory on some roms.


Frank-74:
P.S.
I didn't make a change to Installer.iss.

LuigiBlood:
I randomly tried it on Sim City 64, and it fixes some of the graphical problems I had when I tried this game with it. It's still slow, unfortunately.

Frank-74:
> I randomly tried it on Sim City 64, and it fixes some of the graphical problems I had when I tried this game with it. It's still slow, unfortunately.

Interesting that it would fix some graphic problems.  I had no idea it could do that.  I just patched 1 byte, a relative jump, so it doesn't create an error to write to the log.

I think that's what I did.  I made a guess, as I was just reading the win32 disassembly, and made a guess at what it was doing, going back from the create file export.

It's been merged while typing this reply.

Frank-74:
Just realised I've also disabled dumping textures.

@project64 
Can you revert it, or shall I do another PR to put back the original?

LuigiBlood:
Honestly I don't think it's a huge issue. You shouldn't use Jabo anymore.

LegendOfDragoon:
>You shouldn't use Jabo anymore.

Glide64 is too slow for some people.

>Can you revert it, or shall I do another PR to put back the original?

I would try seeing if you can specifically disable the dumping code, before reverting.

oddMLan:
who dumps textures with Jabo anyways
for that use either Rice or Glide64

PR #1218, opened 2016-11-21
Fix Uncompressed Saves.

Frank-74:


cxd4:
Wrong, it's not deleting line endings; it's adding a missing EOF line break that zilmar's automated (and crappy) code syntax cleaner keeps deleting from the end of the files.  The files should have line ends at the very end of every line; what you're seeing is files with abruptly terminated lines of text being corrected by GitHub's online web editor.

PR #1219, opened 2016-11-21
Only delete old save file being saved.

Frank-74:
Saving compressed file deletes uncompressed save and vice versa.

Also if uncompressed file exists, it loads uncompressed save instead of compressed save.

### DIFF
--- a/Source/Project64-core/N64System/N64Class.cpp
+++ b/Source/Project64-core/N64System/N64Class.cpp
@@ -1518,11 +1518,6 @@ bool CN64System::SaveState()
         SaveFile.DirectoryCreate();
     }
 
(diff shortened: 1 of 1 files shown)


cxd4:
I hope this fixes that buggy accidental deletion of save states if I save and reload too much with F5 and F7 (quick zip-compressed save and restore).  But that was a few years ago and I can't test now without wine or something.

oddMLan:
I used to run into that issue too.

Frank-74:
To make it bullet proof.  It should write to a temp file and only delete the previous save if the temp file write is succesful.

oddMLan:
Sounds good to me 👍

PR #1222, opened 2016-11-24
Make SyncToAudio() counter use ScreenHertz setting.

Frank-74:
I think there's a slight improvement for NTSC games when using Sync using audio.


Frank-74:
Somethings wrong with SyncToAudio().

It only works if Fixed audio timing is enabled.

I put a writetrace into the SyncToAudio function above, and with Sync using audio checked and fixed audio unchecked, it never writes to the log.  Switch Fixed audio on and it does write to the log.

PR #1241, opened 2017-01-03
Fix Glitch64 build.

JunielKatarn:
`fog_enabled` was removed from glitchman.h. This caused build errors:
```
Source\Glitch64\OGLgeometry.cpp(383):Source\Glitch64\OGLgeometry.cpp(383,0): Error C2065: 'fog_enabled': undeclared identifier
Source\Glitch64\OGLgeometry.cpp(385):Source\Glitch64\OGLgeometry.cpp(385,0): Error C2065: 'fog_enabled': undeclared identifier
...
```

because it is stil used in OGLgeometry.
Added the variable statically.


JunielKatarn:
I see this has been fixed already.

PR #1246, opened 2017-01-10
[RDB] Use best possible settings for Super Mario 64

Fanatic-64:
Starting work on the RDB. First things first: Super Mario 64. This is like the most popular game for the console, and probably what most people will first play on the emulator, so it should run as good as possible. Explanation of these settings:

1. Set Counter Factor to 1. Makes the title screen with Mario's disembodied head far smoother (runs at a steady 30 DL/s NTSC/25 DL/s PAL), and should prevent any slowdowns. Also makes transitions between maps slightly quicker.
2. Disabled all Self-Modifying Code Methods. 99% certain this game doesn't need them. Should increase core speed, not that this game is resource-intensive but it doesn't hurt to optimize.

Tested these settings very thoroughly with the (U) ROM (collected all 120 stars, did about every action possible, even tried the hidden level select and debug displays), and fairly extensively with the (E), (J) and Shindou Edition ROMs.

I was tempted to disable Advanced Block Linking (also tested fairly extensively with it Off) to prevent microstuttering, but I think that's something that should be considered on a more global level.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -5541,24 +5541,44 @@ Plugin Note=[Glide64] splitscreen/picture in picture problem
 Good Name=Super Mario 64 (E) (M3)
 Internal Name=SUPER MARIO 64
 Status=Compatible
+Counter Factor=1
+SMM-Cache=0
+SMM-FUNC=0
+SMM-PI DMA=0
+SMM-TLB=0
 Culling=1
 
 [4EAA3D0E-74757C24-C:4A]
 Good Name=Super Mario 64 (J)
 Internal Name=SUPER MARIO 64
 Status=Compatible
+Counter Factor=1
+SMM-Cache=0
+SMM-FUNC=0
+SMM-PI DMA=0
+SMM-TLB=0
 Culling=1
 
 [635A2BFF-8B022326-C:45]
 Good Name=Super Mario 64 (U)
 Internal Name=SUPER MARIO 64
 Status=Compatible
+Counter Factor=1
+SMM-Cache=0
+SMM-FUNC=0
+SMM-PI DMA=0
+SMM-TLB=0
 Culling=1
 
 [D6FBA4A8-6326AA2C-C:4A]
 Good Name=Super Mario 64 - Shindou Edition (J)
 Internal Name=SUPERMARIO64
 Status=Compatible
+Counter Factor=1
+SMM-Cache=0
+SMM-FUNC=0
+SMM-PI DMA=0
+SMM-TLB=0
 Culling=1
 
 [66572080-28E348E1-C:4A]


project64:
From what I remember mario needed cache self mod on. You need to be running with sync core running to help with testing it.

AmbientMalice:
I'd be really cautious about tampering with counter factor, too. The main game doesn't have any obvious speed issues, and counter factor can have all sorts of unexpected side effects. Also, raising the counter factor actually raises the system requirements, so if you're concerned about performance, it's rather counterproductive to be disabling possibly important self mod methods while raising the counter factor.

cxd4:
I don't remember mario needing cache or any self-modifying methods, except if I turned off Use TLB.  But that sounds like I'm mis-remembering, haha.  Since Corn played it fine all the way through on its static recompiler I doubt any self-mod is needed to play the game till the end if Fanatic says it isn't.

LegendOfDragoon:
If it works in Corn then I guess cache isn't needed. I never really found out what Cache did anyway and it was hard to tell a difference in performance with  it on vs off, so I always enabled it.

>I was tempted to disable Advanced Block Linking (also tested fairly extensively with it Off) to prevent microstuttering, but I think that's something that should be considered on a more global level.

I think it's usually better to leave it on.

cxd4:
It's supposed to be faster with it On, but for some reason I always turned it off for most games I played.  I don't remember why.  If it's faster with it on but audio micro-stutters then maybe the results from the optimization are just wrong.

Fanatic-64:
I've tested all versions of the game pretty thoroughly on the latest debug build (compiled with VS2008) using Sync Cores (side note: does anyone else have trouble getting ROMs to load with that setting? I have to set Recompiler, load the ROM, then set Sync Cores and hard reset), and found no issues with the proposed settings. Also tested most of the game with Advanced Block Linking off and didn't find any issues either.

@AmbientMalice I've tested the game very thoroughly with Counter Factor=1 (in fact, I've using that setting with Super Mario 64 (U) for _years_) and can pretty much guarantee that it doesn't cause any negative side effects (timers run fine, enemies behave fine, everything works). While it technically increases CPU resource usage, Super Mario 64 is a lightweight game anyway (I've played it at full speed with that setting on a shitty AMD laptop from 2007), and the setting does give a tangible improvement. The SMCMs were disabled because they offer no benefit and are a free performance improvement that should at least partially offset the performance penalty of CF=1 (to allow for faster uncapped speed, or for those with really old clunkers).

@cxd4 Older versions of Project64 (at least up to 1.6) have an issue in some games (including SM64) where if Advanced Block Linking is turned off, the Self-Modifying Code Method must be set at least to Cache (it is mentioned in the old PJ64 user manual). The current version of PJ64 doesn't appear to have this issue from what I've tested. Regarding the effects of Advanced Block Linking, the stuttering is more like momentary pausing of the game (audio and video), the old user manual again explains the effects of that setting:
> Advanced Block Linking is one of Project64's speed optimisation techniques, [...] this control is a performance option for the recompiler, where setting On gives more speed (higher maximum and overall average speed) and setting Off gives better "smoothness" (higher consistency of speed, less variation from one part of a game to another, fewer jerks and slowdowns, but a lower overall speed). [...] During the early stages of Project64 development performance was a major concern, but we expect Project64 to perform better on future PCs without ABL.
So, using Advanced Block Linking is faster but does have a trade-off. The ABL code has since been rewritten, probably reducing some of these effects, but they still exist.

cxd4:
It doesn't have any greater trade-off than using the interpreter instead of the dynarec.  The dynarec's primary speed optimization is register caching.  I usually turned ABL off when I didn't need it for full speed.

If I needed *maximum* speed (like when getting Virtual Chess 64, set to maximum difficulty level) to hurry up and play a move on the chess board, using a dynarec with ABL=On was far from the only thing I did to make emulation faster.  CF=4 or =5, use Rice's Daedalus 5.2.0, frame skip, ... plenty of things that could be done if you just want the "smoothness" of over-optimizing something that's already full-speed anyway.

>Older versions of Project64 (at least up to 1.6) have an issue in some games (including SM64) where if Advanced Block Linking is turned off, the Self-Modifying Code Method must be set at least to Cache (it is mentioned in the old PJ64 user manual).

Could be.  I never would have noticed it since all the games except Mario64 require a minimum of "Cache" for the self-modifying code method anyway; the only lower setting is None at all, which would be just a few demos and maybe Mario64 and that's it.

PR #1248, opened 2017-01-19
Upgrade zlib to version 1.2.11

JunielKatarn:


project64:
seems fine, does this fix the save state issue on XP when using the latter VS versions?

I will merge this after I do the next release just cause there not been enough testing/time with the new code.

PR #1253, opened 2017-01-24
Update French.pj.Lang

BenjaminSiskoo:
small fix : La version\ Android u2122 --> La version \u2122 Android


baptiste0602:
@BenjaminSiskoo Still not good, u2122 is Unicode for Copyright caracter, so it should be after "Android" not before

BenjaminSiskoo:
@baptiste0602 Ok changed. Hope it's right now :) Thanks

PR #1260, opened 2017-01-25
Change Z_BUFSIZE to fix zlib behavior on Windows XP.

JunielKatarn:
Project64 failed to save states on Windows XP when built with Visual Studio 2010 or later.
This is likely a compiler issue related to memory layout.

To work around it, changed the value of `Z_BUFSIZE` in ZLib's makefile (zlib.vcxproj) to a lower one (from `64KB` to about `45KB`).


JunielKatarn:
@oddMLan, @Frank-74, @project64 This fixes XP savestate issues. If all XP issues are gone, it should be safe to drop support for VS2008.

PR #1262, opened 2017-01-26
Update German.pj.Lang

daarky:
language file has not been loaded


cxd4:
I'm not understanding what this is....

PR #1264, opened 2017-01-27
Fix crash in 1080 Snowboarding

Frank-74:
Game crashes with ABL on when pressing A+B together in Trick Attack or Contest modes.


project64:
not really fix .. just hide :P

PR #1267, opened 2017-01-29
Fix GoodName missing from Project64.cfg

### LINKED ISSUE #1263: Good Names missing from .cfg.
In 2.3.2-200, .cfg has Good Names.  But Good Names are disappearing in .cfg with 2.3.2-202, if I Edit game settings.  Good Name is no longer written at all to .cfg.

Guessing it's caused by https://github.com/project64/project64/commit/bf0ba50dc494f21b91a353c3ab9a926a5699d0d8.

Frank-74:
Fix #1263.

### DIFF
--- a/Source/Project64-core/Settings/SettingsClass.cpp
+++ b/Source/Project64-core/Settings/SettingsClass.cpp
@@ -171,7 +171,7 @@ void CSettings::AddHowToHandleSetting(const char * BaseDirectory)
     AddHandler(Game_File, new CSettingTypeTempString(""));
     AddHandler(Game_UniqueSaveDir, new CSettingTypeTempString(""));
     AddHandler(Game_GameName, new CSettingTypeTempString(""));
-    AddHandler(Game_GoodName, new CSettingTypeGame("Good Name", Rdb_GoodName));
+    AddHandler(Game_GoodName, new CSettingTypeGame("Good Name", Game_GameName));
     AddHandler(Game_TempLoaded, new CSettingTypeTempBool(false));
     AddHandler(Game_SystemType, new CSettingTypeTempNumber(SYSTEM_NTSC));
     AddHandler(Game_EditPlugin_Gfx, new CSettingTypeGame("Plugin-Gfx", Default_None));


project64:
I need to work out a proper solution, maybe this is correct, but it looks wrong ..

the setting config is saying read it out of the cfg, file if not found then read out of rdb
this change says read out of cfg, if not found read out of cfg ..

Frank-74:
> this change says read out of cfg, if not found read out of cfg ..

I thought it was, read out of cfg, if not found, read out of tempstring.
It's what the Rdb_GoodName does.

project64:
Goodname is reading from cfg, if not existing read from rdb .. 

maybe a better solution is to just read from rdb, so rename Rdb_GoodName to Game_GoodName
and the current Game_GoodName to Cfg_GoodName.

It does mean you can not override it in the cfg file, but I do not think any one would be doing that anyway. It is just more useful there for manual editing. 

That way saving to the Cfg_GoodName will have an empty default and it will not match the default and get deleted. So it will fix it automatically by doing this.

Frank-74:
I tried renaming, but it caused Config, window title and savestates to have blank Good Name.  Maybe I did something wrong, or missed something, there are over 20 renames to be done.  Renaming Game_GoodName to Cfg_GoodName and Rdb_GoodName to Game_GoodName.

The tempstring Game_GameName, is just "", until a ROM is loaded or edited.  Using Game_GameName as default, was the only way I managed to get it to work without breaking anything else.

I also tried if (m_DefaultSetting != Rdb_GoodName) condition on the delete setting, it still deletes it.  Tried using Game_GoodName for condition as well.

project64:
Should be like:
* Rename Game_GoodName to Cfg_GoodName
* Change all usages of Game_GoodName to Rdb_GoodName except for the case where writing to cfg file
* Set handler for Cfg_GoodName to be like AddHandler(Cfg_GoodName, new CSettingTypeGame("Good Name", ""));

Frank-74:
Ok, I've got it working.  Just need to do a new PR.

PR #1269, opened 2017-01-31
Fix GoodName missing from Project64.cfg

### LINKED ISSUE #1263: Good Names missing from .cfg.

Frank-74:
Fix #1263.

### DIFF
--- a/Source/Project64/UserInterface/SettingsConfig.cpp
+++ b/Source/Project64/UserInterface/SettingsConfig.cpp
@@ -251,7 +251,7 @@ void CSettingConfig::ApplySettings(bool UpdateScreen)
     if (!GameIni.empty())
     {
         stdstr GoodName;
-        if (!g_Settings->LoadStringVal(Rdb_GoodName, GoodName))
+        if (g_Settings->LoadStringVal(Rdb_GoodName, GoodName))
         {
             g_Settings->SaveString(Cfg_GoodName, GoodName);
         }
--- a/Source/Project64/UserInterface/SettingsConfig.cpp
+++ b/Source/Project64/UserInterface/SettingsConfig.cpp
@@ -80,7 +80,7 @@ LRESULT	CSettingConfig::OnInitDialog(UINT /*uMsg*/, WPARAM /*wParam*/, LPARAM /*
 
     if (!GameIni.empty())
     {
-        ConfigRomTitle.Format("Config: %s", g_Settings->LoadStringVal(Game_GoodName).c_str());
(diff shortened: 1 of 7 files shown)


Source/Project64/UserInterface/SettingsConfig.cpp
```diff
@@ -251,9 +251,9 @@ void CSettingConfig::ApplySettings(bool UpdateScreen)
     if (!GameIni.empty())
     {
         stdstr GoodName;
-        if (!g_Settings->LoadStringVal(Game_GoodName, GoodName))
+        if (!g_Settings->LoadStringVal(Cfg_GoodName, GoodName))
```
Frank-74: Wasn't sure if this should be Rdb_GoodName.

project64: should be 

if (g_Settings->LoadStringVal(Rdb_GoodName, GoodName))

Since Cfg_GoodName does not inherit from Rdb_GoodName it would be always writing an empty string to the cfg file

project64: the way the setting class works is that it will load it from the source, if it does not exist it will read if from the default.

it will return true if read from the source, false if from the default.

So if this returned false before it read it from rdb_goodname so had the goodname now in the variable, and then could just write it.

Looks good, so basically all reads come from rdb_goodname and all writes are to cfg_goodname

Frank-74: Ok, changed it to read from Rdb, write to Cfg.  I should have done what I thought in the first place.

Good to go?

Source/Project64/UserInterface/SettingsConfig.cpp
```diff
@@ -251,7 +251,7 @@ void CSettingConfig::ApplySettings(bool UpdateScreen)
     if (!GameIni.empty())
     {
         stdstr GoodName;
-        if (!g_Settings->LoadStringVal(Cfg_GoodName, GoodName))
+        if (!g_Settings->LoadStringVal(Rdb_GoodName, GoodName))
```
project64: Remove the !

project64: it can be just Load then save do not even need the if statement

Frank-74: Removed the !.

PR #1272, opened 2017-02-05
Remove pjutil::DynLibCallDllMain()

Frank-74:


cxd4:
Jabo's 1.5.2 and 1.6 plugins (particularly video) are still useful and fix some things that 1.6.1 and 1.7 don't.

It's a WIN32 hack from an otherwise portable module, yes, so removing it is beneficial in that respect.  Bear in mind though that Jabo's older plugins have served good use for people where 1.6+ weren't good.  They're not open-source, though, so it's not really worth fighting over.  Especially since I don't use Windows anymore anyway.

Frank-74:
@cxd4 
I tested Jabo's 1.5.2, 1.6, 1.6.1 and 1.7 gfx plugins.  They all work fine without this WIN32 hack.  I think this hack was for an even older version than 1.5.2.

cxd4:
Oh.  Then I don't care at all personally. :)

PR #1284, opened 2017-03-10
Donkey Kong 64 (U) Polygon tearing fix

Frank-74:


oddMLan:
imagen
confirmed fixing the polygon tearing

PR #1285, opened 2017-03-12
Donkey Kong 64 (E)(J) Polygon tearing fix

Frank-74:
Original (U) fix by Isotarge.  Simplified to a single code and E and J regions by Retroben.


weinerschnitzel:
There wasn't any side effects to this? Can I get a link to where I can read more on Isotarge's hack fix?

Thanks in advance.

Frank-74:
@weinerschnitzel 
I played for about an hour, and never noticed any side effects.

I don't know where to get info about the original fix by Isotarge, retroben knows more.  From this thread:
http://forum.pj64-emu.com/showthread.php?t=7559

PR #1317, opened 2017-05-30
Multiplayer timing fix for E and J regions

Nekokabu:
Add Mario Kart Multiplayer timing fix cheat for E and J regions.


loganmc10:
Are the E and J versions really the same? How were these calculated?

loganmc10:
Also these are half the size of the U version. Does this fix 2 and 3/4 player?

Nekokabu:
>Are the E and J versions really the same? How were these calculated?

Original code? is here: http://tasvideos.org/GameResources/N64/MarioKart64/Patches.html
I disassemble and compare them.

>Also these are half the size of the U version. Does this fix 2 and 3/4 player?

Unnecessary code was deleted, so it got shorter. 
Including 2 and 3/4 player fix

Nekokabu:
https://github.com/project64/project64/pull/607

Issue #1318, opened 2017-06-04
Project64 Video Aspect Ratio setting cannot be changed.


Frank-74:
Unable to change from 4:3 to any other setting (force 16:9, stretch, original).

Frank-74:
Think I found a fix.  Removing this line fixes it.
https://github.com/project64/project64/blob/bf5dc3517df2cc3c93d455616b7e4effc3aa2bdc/Source/Project64-video/Settings.cpp#L916

Frank-74:
Fixed with https://github.com/project64/project64/commit/a9683b36fb064b1a0779587bca1990e5021298a3.

PR #1319, opened 2017-06-05
Fix Aspect Ratio setting (#1318)

Frank-74:
Fixes https://github.com/project64/project64/issues/1318


project64:
while this might fix it for window, it will break it on android.

basically I need a flag when inside CSettings::WriteSettings, when that function is executing, SettingsChanged should not execute.  So set flag at start of CSettings::WriteSettings, clear it at the end. And check at the start of settingsChanged if set, if so just return.

project64:
does https://github.com/project64/project64/commit/a9683b36fb064b1a0779587bca1990e5021298a3 fix it?

Frank-74:
Yes, it's fixed now.  Tested in the first minute of the commit.

Cheers.

PR #1321, opened 2017-06-08
Skip Android modules on Windows builds.

JunielKatarn:
Currently, doing a full Windows build also compiles Android plugins, unnecessarily increasing build time.

This change excludes those projects when building for Windows x86 or x64.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,10 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
-
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


project64:
if something changes as you changes code, then a build should build android to make sure android is not broken to.

How much does it add to build times? It should only add to build times when you rebuild all/first build.

JunielKatarn:
It doesn't add significant time. It was more about not building non-required components for the given workflow.
Makes sense if you want to leave them for validation purposes.

PR #1335, opened 2017-07-10
Wait for previous process to terminate

LegendOfDragoon:
The new process will not have read/write access to config files if it
does not wait until previous processes have been terminated.

### DIFF
--- a/Source/Common/Util.cpp
+++ b/Source/Common/Util.cpp
@@ -107,6 +107,7 @@ bool pjutil::TerminatedExistingExe()
                     if (TerminateProcess(hHandle, 0))
                     {
                         bTerminated = true;
+                        WaitForSingleObject(hHandle, 30 * 1000);
                     }
                     else
                     {


Source/Common/Util.cpp
```diff
@@ -107,6 +107,7 @@ bool pjutil::TerminatedExistingExe()
                     if (TerminateProcess(hHandle, 0))
                     {
                         bTerminated = true;
+                        WaitForSingleObject(hHandle, 30 * 1000);
```
cxd4: This looks like highly inefficient code.

It would be 30 times more optimized to just say `WaitForSingleObject(hHandle, 30000);`.

loganmc10: Wouldn't any decent compiler do that itself?

cxd4: Yep, but he likes figuring out things for the compiler so I just felt like being an ass.

PR #1341, opened 2017-07-29
Fix #1340 black sky in OoT

### LINKED ISSUE #1340: [Project64-Video] Another regression.
Just discovered OoT has a problem with the sky after the logo is displayed.
glide64_the_legend_of_zelda_04
The sky appears in the next camera change.

Going back to Project64-Video from the overclock branch, it's half broken at the same place.
glide64_the_legend_of_zelda_01

It's fine before it became Project64-Video.
I'll try to find the commit that started it over the weekend.

Frank-74:


project64:
good find

Frank-74:
It looks like there's more to be done in terms of optimizing this function.  The else condition could probably be removed.

PR #1342, opened 2017-07-30
Added missing include for x64 build.

JunielKatarn:
MSVC 14.1 and later require <intrin.h> to define `__cpuid` (See https://docs.microsoft.com/en-us/cpp/intrinsics/cpuid-cpuidex).
* Tested in both Visual Studio 2015 and Visual Studio 2017.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,8 @@ Thumbs.db
 /Plugin64/GFX/pdb
 /Plugin64/GFX/PJ64Glide64.dll
 /Plugin64/GFX/PJ64Glide64_d.dll
+/Plugin64/GFX/Project64-Video.dll
+/Plugin64/GFX/Project64-Video_d.dll
 /Plugin64/Input/lib
 /Plugin64/Input/map
 /Plugin64/Input/pdb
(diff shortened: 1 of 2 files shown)


cxd4:
Lumping the include under an existing `#ifdef _WIN32` here is going to cause more problems.

I would suggest that you either not include it at all and simply fix the `__cpuid` code to not be needed for 64-bit builds or at the very least put the include under something that checks the value of `_MSC_VER` to avoid breaking builds with WDK versions, older versions of Visual Studio, or non-Visual-Studio IDEs.

JunielKatarn:
Fair enough.

cxd4:
Eventually @project64 the inline asm can be removed to just use the intrinsic version.

The only thing is that the SSE2 recompiler features used in the RSP are already unconditionally supported by any 64-bit-capable CPU.  So if you build in _WIN64 or something it makes sense to just not do any CPUID computation at all, since maybe there are some WDK versions of MSVC that don't include the header with 64-bit versions either.

The only reason that I don't do a PR for that is that I don't know the syntax for AMD and Intel features when passing 1 to __cpuid(), but anything you're doing in the RSP recompiler a 64-bit CPU already supports.  So if you avoid inline asm because 64-bit compiles in MSVC don't support inline asm you really could eliminate needing to use the CPUID instruction at all.

LegendOfDragoon:
>So if you build in _WIN64 or something it makes sense to just not do any CPUID computation at all,

I agree. No point in checking for features if you already know for sure the user's machine supports it :smile: .

PR #1347, opened 2017-07-30
Added ClickOnce deployment scripts.

JunielKatarn:
Adding the packaging and deployment scripts for the ClickOnce installer (already used in my [buildbot](http://hyvart.com/download/project64/)).

### DIFF
--- a/Source/Script/PowerShell/Launcher/Launcher.cs
+++ b/Source/Script/PowerShell/Launcher/Launcher.cs
@@ -14,7 +14,7 @@ namespace Project64.ClickOnce
 			{
 				Environment.SpecialFolder folder = Environment.SpecialFolder.LocalApplicationData;
 				string localPath = Environment.GetFolderPath(folder);
-				string localAppPath = Path.Combine(localPath, "Hyvart Software", "Project64Dev");
+				string localAppPath = Path.Combine(localPath, "Project64", "Project64Dev");
 				string configPath = Path.Combine(localAppPath, "Config");
 
 				// If local config folder exists, offer to keep/overwrite.
--- a/Source/Script/PowerShell/Launcher/Launcher.cs
+++ b/Source/Script/PowerShell/Launcher/Launcher.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using System.Windows.Forms;
+using System.Deployment.Application;
+
+namespace Project64.ClickOnce
+{
+	class Launcher
+	{
+		public static void Main()
+		{
+			try
+			{
+				Environment.SpecialFolder folder = Environment.SpecialFolder.LocalApplicationData;
+				string localPath = Environment.GetFolderPath(folder);
+				string localAppPath = Path.Combine(localPath, "Hyvart Software", "Project64Dev");
+				string configPath = Path.Combine(localAppPath, "Config");
+
+				// If local config folder exists, offer to keep/overwrite.
+				if (ApplicationDeployment.CurrentDeployment.IsFirstRun && Directory.Exists(configPath))
+				{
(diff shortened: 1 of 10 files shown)


Source/Script/PowerShell/Launcher/Launcher.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+	class Launcher
+	{
+		public static void Main()
+		{
+			try
+			{
+				Environment.SpecialFolder folder = Environment.SpecialFolder.LocalApplicationData;
+				string localPath = Environment.GetFolderPath(folder);
+				string localAppPath = Path.Combine(localPath, "Hyvart Software", "Project64Dev");
```
project64: what does Hyvart Software have to do with pj64

JunielKatarn: Nothing! Outdated strings. Fixing...

Source/Script/PowerShell/Tfs/Archive.ps1
```diff
@@ -0,0 +1,16 @@
...
+    [SecureString]
+    $FtpPassword,
+    $FtpHost
+)
+
+mkdir $(Agent.BuildDirectory)\zip;
+
+Add-Type -Assembly 'System.IO.Compression.FileSystem';
+
+[System.IO.Compression.ZipFile]::CreateFromDirectory("$(Agent.BuildDirectory)\artifacts\drop\", "$(Agent.BuildDirectory)\zip\$(Build.BuildNumber).zip\");
```
project64: build number, I assume is from the build system is not great, it should use the information from git to determine build description

project64:
what exactly does this add? How is it any different/better then what is already there?

JunielKatarn:
@project64 Allows for packaging the app as a ClickOnce application, tracking updates automatically without requiring an installer. It doesn't interfere at all with the current scripts.

project64:
so an alternative installer, that is just a link to the latest version?

Source/Script/PowerShell/Tfs/Archive.ps1
JunielKatarn: This build number is orthogonal to the emulator version.
The Pj64 binaries still keep their metadata and manifest info.

Again, it's for ClickOnce packaging, and doesn't I terrero with the traditional installer or zip.

JunielKatarn:
@project64 not a link. It downloads the binaries from wherever the installer is placed.
Just an alternative (and quicker) way to distribute the builds.

project64:
Does it pull down the installer and run that? 

I do not see any advantage of this? Just seems like more compilications

JunielKatarn:
Not quite. See http://hyvart.com/download/project64/project64dev.application
It tracks the latest version and updates the necessary binaries. Also doesn't require admin privileges.
It's also independent from the regular installer. It's meant to run only by us who want to package it that way. Users don't interact with it.

Do you see any issues?

project64:
You have spent time on it, I understand adding the scripts to main line helps you out with them there, I want to like it cause someone else is paying attention and adding to project64. 

There is just a few issues really, I do not like click once, so you have a huge up hill battle there. The other thing if I add it then I generally want to use it or see that I will use it. I just do not see me using it. It just seems an alternative to the installer. 

Ok it does not use admin privileges, the only reason the installer does it cause it installs in to program files, yes it could install in to the app data dir as the the example you have above. But that is not the common place applications are installed. I like that the installer at least gives the option of where to install to. 

Ok granted that most users are just going to press next and not care, but I still much prefer giving them that option.

I just see this as a replacement for the installer or an alternative downloader for the installer, which again I just do not see the benefit.

JunielKatarn:
Alright, I won't push the issue further.
I can always just have these scripts in a private repository.

project64:
I should have the builds on the site soon as well, tho I am making it just downloadable for patreon members

PR #1357, opened 2017-08-18
Add debug features

shygoo:
From https://github.com/shygoo/project64


project64:
I will merge and any fall out we can fix after

PR #1373, opened 2017-09-16
Some RDB improvements

AmbientMalice:
I've added some rom hacks/fan translations, plus tweaked a few incorrect entries.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -767,13 +767,13 @@ Status=Compatible
 Clear Frame=1
 
 [DF6FF0F4-29D14238-C:4A]
-Good Name=Bomberman 64 (J)
+Good Name=Bomberman 64 - Arcade Edition (J)
 Internal Name=BOMBERMAN64
 Status=Compatible
 Culling=1
 
 [9780C9FB-67CF6B4A-C:45]
-Good Name=Bomberman 64 (J) [T]
+Good Name=Bomberman 64 - Arcade Edition (J) [T]
 Internal Name=BOMBERMAN64
 Status=Compatible
 Culling=1
@@ -1011,7 +1011,7 @@ Plugin Note=[input] disable Memory Pak
 Good Name=Chameleon Twist (J) [T]
 Internal Name=Chameleon Twist
 Status=Compatible
-Plugin Note=[input] disable Memory Pak
+Plugin Note=
 
 [6420535A-50028062-C:45]
 Good Name=Chameleon Twist (U) (V1.0)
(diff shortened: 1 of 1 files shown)


AmbientMalice:
I removed CF=1 from Indiana Jones to bring it in line with mupen64plus. The game appears to work perfectly in mupen64plus now, but PJ64 still has freezes at the end of each level.

Config/Project64.rdb
```diff
@@ -150,6 +150,15 @@ Bad ROM?.AutoFullScreen=False
...
 // ---- END OF HEADER, START OF ROM ENTRIES --------------------------
 //================  0-9  ================
+[B2242748-FFBD61DA-C:45]
+Good Name=007 - Goldfinger (Beta)
+32bit=No
+Counter Factor=1
+Fast SP=0
+Good Name=GOLDFINGER
+Linking=0
+RDRamSize=8388608
```
Frank-74: RDB has different syntax compared to .cfg.
RDRAM Size=8
Linking=Off

AmbientMalice: Fixed it. Thanks.

LegendOfDragoon:
I'm not sure if removing `HLE GFX=No` is a good idea. Sure it benefits GLideN64 users, but it does the opposite for people who use Jabo's or PJ64's Video plugin.

I could be wrong, but I think the RDB is meant to prioritize 1st party plugins.

AmbientMalice:
HLE GFX=No was added to these games with the accurate at that time assumption that there was no way for them to be emulated in HLE. That is no longer the case, and that assumption is causing a huge usability issue for the emulator.

PJ64-Video's LLE mode is completely broken. Jabo's isn't all that much better. So what you're going to get is:

Person tries Rogue Squadron.
It's broken.
So they download GLideN64 and install it.
It's still broken.
They google the problem.
Fixing it is complicated and involved heaps of submenus and enabling Advanced Mode.
At this point, a vast majority of casual users are going to either give up or go download mupen64plus instead.

Additionally, not that it'll actually much good because it's so broken, there's nothing stopping the HLE ucode implementations being backported to PJ64-Video.

AmbientMalice:
>”- Arcade Edition ” is not necessary.
>It is not included in the title screen, box art etc.
It's an unofficial title that distinguishes the game from Bomberman 64. I alternatively considered labeling it Bomberman 64 (2001). It's rare for two completely different N64 games to have the exact same name, and this causes some confusion.

cxd4:
Yeah the (J) Bomberman 64 from 1997 was called Baku Bomberman or whatever.  Still I was quite confused to see Bomberman 64 (J) and think it wasn't an alternative arcade editition when in fact it was.

Issue #1374, opened 2017-09-18
[Request] 16 MB RAM size


Tarnadas:
The developers from Super Mario 64 Online need it, so I was wondering, if this is possible.
 My C++ and Emulator knowledge is limited, so it would be easiest to just ask.

Frank-74:
It's already supported.  But not selectable from the options.  I added the option back in February, but never did a PR for it.  You can see what is needed to change in these commits:
https://github.com/Frank-74/project64/commit/305a2b2db3c81f136643c7d817e46142280b58de
https://github.com/Frank-74/project64/commit/c032d1fd865914f59bf4d7b17feaedce915a8494
https://github.com/Frank-74/project64/commit/63e7c5cb4f6c6b8f2a8e59f943824872c77dbc78

For now, you can add RDRamSize=16777216, under the Super Mario 64 entry in Project64.cfg.

Frank-74:
I've just done a PR to add 16MB option.

MelonSpeedruns:
Frank, thanks for your help! I've tried injecting custom code int he 16mb area and it works! However, trying to loads 3D models freezes the game... Any idea what could be the cause?

Frank-74:
> However, trying to loads 3D models freezes the game... Any idea what could be the cause?

@MelonSpeedruns
Sorry, no idea.  Best uneducated guess is if they're new 3D models, there might be a limit to amount of polygons.  N64 3D models are all very low polygon count.

cxd4:
Also a lot of stuff in Project64's RSP (and even some HLE graphics plugins) assume a maximum RDRAM address of 0x7FFFFF which is good for only 8 MB and not 16, so adding the support in Project64's core doesn't fix the plugins.

OffSetEyeBrow:
hi, i downloalded the edited file, but is the android edition?

project64:
Not going to make it a visible option, closing for now

whateverlior:
> Not going to make it a visible option, closing for now

well thats badass

Issue #1031, opened 2016-03-19
RDRam size missing in the plugin spec


purplemarshmallow:
The gfx plugin needs to know the size of RDRam for correct RDP emulation. The plugin spec does not provide it and there are no simple and not platform specific ways to detect it.

Maybe it's possible to make an optional extenson to the plugin spec without breaking compatibility with older plugins. Or bump the version of the spec.

Currectly most plugins try to detect it with causing an access violation and then they assume it's either 4 or 8 MB. PJ64 allows setting it to any amount so this solution is problematic

cxd4:
Definitely, but I don't know of any solutions there.  Any optional extension to the specs would require adding to the PLUGIN_INFO (i.e. RSP_INFO etc.) struct's in the headers, which changes the size of the struct and can and will go out of bounds for the memory info that the emulator core's implementation of the specs expects.

The only compatible solution I know of would be to make a new official release of the plugin specs, like Gfx 1.4, Audio 1.2, Controller 1.2 and RSP 1.2, where the plugin info struct's supply a `size_t max_DRAM_address` or the like.

purplemarshmallow:
The Gfx 1.4, Audio, 1.2 Controller 1.2 and RSP 1.2 spec already exists but I'm not sure what is changed and only PJ64 supports it

Glide64 for PJ64 already uses the 1.4 spec. It could be changed to 1.5 and this ugly code could be removed
https://github.com/project64/project64/blob/master/Source/Glide64/Main.cpp#L1620

I think an official release of the new spec with new headers would be a good idea.

cxd4:
>The Gfx 1.4, Audio, 1.2 Controller 1.2 and RSP 1.2 spec already exists but I'm not sure what is changed and only PJ64 supports it

They actually kind of do not.

That's because that's just Project64-ization garbage and not really legitimate versions of the spec.  zilmar mostly used those version numbers as a means of keeping people from using the plugins on other emulators, primarily as far as the donations lifespan of Project64 1.7 was concerned.

>Glide64 for PJ64 already uses the 1.4 spec.

Again that really is just because of personal custom extensions zilmar did to the specs for Project64.  Said custom extensions were custom to begin with, so I stand by what I said about adding this `size_t RDRAM_size` member of the plugin info structs you suggest in a Gfx 1.4, Audio 1.2, Controller 1.2 and RSP 1.2 of the specs.

[...]

PR #1375, opened 2017-09-18
Add 16MB RAM to options

Frank-74:
Needed for Super Mario 64 Online mod community.  #1374

### DIFF
--- a/Source/Project64-core/Multilanguage.h
+++ b/Source/Project64-core/Multilanguage.h
@@ -345,6 +345,7 @@ enum LanguageStringID
     //RDRAM Size
     RDRAM_4MB = 580,
     RDRAM_8MB = 581,
+    RDRAM_16MB = 582,
 
     //Advanced Block Linking
     ABL_ON = 600,
--- a/Source/Project64-core/Multilanguage/LanguageClass.cpp
+++ b/Source/Project64-core/Multilanguage/LanguageClass.cpp
@@ -313,6 +313,7 @@ void CLanguage::LoadDefaultStrings(void)
     //RDRAM Size
     DEF_STR(RDRAM_4MB, "4 MB");
     DEF_STR(RDRAM_8MB, "8 MB");
+    DEF_STR(RDRAM_16MB, "16 MB");
 
     //Advanced Block Linking
     DEF_STR(ABL_ON, "On");
--- a/Source/Project64/UserInterface/Settings/SettingsPage-Game-General.cpp
+++ b/Source/Project64/UserInterface/Settings/SettingsPage-Game-General.cpp
@@ -53,6 +53,7 @@ CGameGeneralPage::CGameGeneralPage(HWND hParent, const RECT & rcDispay)
         ComboBox->SetTextField(GetDlgItem(IDC_MEMORY_SIZE_TEXT));
         ComboBox->AddItemW(wGS(RDRAM_4MB).c_str(), 0x400000);
         ComboBox->AddItemW(wGS(RDRAM_8MB).c_str(), 0x800000);
+        ComboBox->AddItemW(wGS(RDRAM_16MB).c_str(), 0x1000000);
     }
 
     ComboBox = AddModComboBox(GetDlgItem(IDC_SAVE_TYPE), Game_SaveChip);


oddMLan:
Hm. Would a better solution be to add a "custom" option in the combobox selection? Where you can set it as high as you want, easier for development.

I know the RDB parameter is there for that, but it's not documented anywhere. I think it's time for a wiki (or document whatever's missing in the one we got here). At least the UI would be more discoverable, due to lack of documentation.

Frank-74:
@oddMLan 
Not sure it's necessary, though there was once a custom build called Fruity or something, that allowed 20 or 24 MB if I recall.  Not sure why though.

cxd4:
If they really want it to be custom I don't see why not just edit the CFG file.

4, 8, and 16 MB are the only significant amounts.  You could go up to 512 MB for weird semi-DD stuff, but nothing in the RCP goes past the 16 MB barrier so it would be inconsistent.

You could even have 2 MB of RDRAM installed to the N64 console; I do not see that 4 MB is the minimum.
From what I remember in the address map there's a range barrier around every 2 MB.

cxd4:
Fruity was a very fruity project.

project64:
I am not sure about this, mostly cause you can not do it on a real n64, as you said there is the ability to do it with the config .. I think it should only be there in advanced mode .. and even then I am not sure.

cxd4:
>mostly cause you can not do it on a real n64

A few peeps on #n64dev would like to have a word with you about that.

It's been done and that's been proven false.  It only wasn't possible to do it on a real N64 in the late 90s because (for all the expensiveness of RAM) Nintendo had no official accessories priceable at that time for having 16 MB globally available to consumers.  Saying "you can not do it on a real n64" is a failed reiteration of that circumstance.

project64:
yes it might be possible to hack the hardware to do it .. but I mean for a real n64 with official accessories you can not.

cxd4:
Why would you think it has anything to do with hacking?

I'm no manufacturer, but even before Nintendo added an R4300 on to it, SGI's RCP was developed and tested with different multiples of 2 MB of RAM.  Did they have to hack their own hardware to do it?  Did Nintendo have to hack the N64 hardware to release an 8 MB expansion pak "hack"?

No, it just takes someone who can install RAM chips of different sizes in to the system, and Nintendo wasn't the only manufacturer of that.  Because they, themselves, did not release a 16 MB one doesn't mean it couldn't be done without a hack; it was just too expensive in the 90s.

Frank-74:
I'll close this PR.  It will need #1031 to be implemented first anyway, then maybe have this option in the enhancement branch.

project64:
yes, as an option to the enhancement build it might be possible .. there will never be fixing jabo's plugin .. but rsp and video should be able to get the size of the rdram from the emulator. Tho again not sure if that helps Gliden64 but that is a different issue.

cxd4:
It's not an "enhancement" if the actual N64 system supports it.

You may as well call any size RAM an enhancement because the base N64 console doesn't come with any RAM at all built in; an external 4 MB is just put in there when selling the thing.  Do you call that an enhancement?

oddMLan:
Also consider that adding this feature would not only aid the development of hacks, but by helping them (SM64 hacks are always popular on YouTube) it would attract tons of potential patrons to your project. Project64 strives for accuracy but also enhancements over the original hardware; I see this more as an enhancement rather than a hack. There's plenty of other hacks for other consoles that don't require extended memory addresses, but I guess in those situations hardware limits weren't such a big deal there, on the other hand, the N64 is severely limited in memory. Which I guess, would not allow a SM64 online multiplayer with extra characters to happen under just 8MB RDRAM

Frank-74:
Adding RDRAM size to spec would help gonetz and GLideN64, he says here: https://github.com/gonetz/GLideN64/issues/1577#issuecomment-330739310

cxd4:
The only way it would be an enhancement is if it was a hack, and installing RAM to the N64 isn't hacking.

It may be valid to consider it as an enhancement to the games themselves, which are not purposely designed to exploit the availability of 16 MiB RDRAM.  It then follows that it is an enhancement for the enhancement build of Project64 if the intended function of Project64 is to play games, which zilmar denies by calling it the best "N64 emulator", which is not the same thing as a game player.  If it was a game player we wouldn't need it; many of these games have already or can be ported to a PC version for Windows defeating the need to have a dynarec or an interpreter for them.  An emulator has unique purpose; a game-centric plug-and-play ROM loader thing does not.

cxd4:
It's a hack and an enhancement for the commercial games, not the hardware.

PR #1378, opened 2017-09-22
Updated "Aidyn Chronicles - The First Mage" entries in rdb/rdx..

MelchiorGaspar:
Updated "Aidyn Chronicles - The First Mage" entries in rdb/rdx.. 

Good Name=, 
AND added added missing  (U)[!](V1.1) postfix in the rdx file 
and updated the release dates.

### DIFF
--- a/Config/Project64.rdx
+++ b/Config/Project64.rdx
@@ -173,17 +173,25 @@ Players=4
 ForceFeedback=No
 
 [2DC4FFCC-C8FF5A21-C:50]
-Good Name=Aidyn Chronicles - The First Mage (E) [!]
+Good Name=Aidyn Chronicles - The First Mage (E)[!]
 Developer=H2O Entertainment
-ReleaseDate=2001/03/20
+ReleaseDate=2001/08/03
 Genre=RPG
 Players=1
 ForceFeedback=No
 
 [E6A95A4F-BAD2EA23-C:45]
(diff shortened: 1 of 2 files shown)


MelchiorGaspar:
Huwayy! my COMMIT has been accepted!

Config/Project64.rdx
```diff
@@ -173,17 +173,25 @@ Players=4
 ForceFeedback=No
 
 [2DC4FFCC-C8FF5A21-C:50]
-Good Name=Aidyn Chronicles - The First Mage (E) [!]
+Good Name=Aidyn Chronicles - The First Mage (E)[!]
 Developer=H2O Entertainment
-ReleaseDate=2001/03/20
+ReleaseDate=2001/08/03
```
cxd4: It's funny.
Actually looking through the RDX file, it looks like almost every game in there has the same release date copy pasta'd across regions.  So Mario U/E/J all have the same release date in the RDX (even though they don't).

Hardly anything to create an issue over (since it has nothing to do with N64 emulation), but the RDX database turns out to be rather sloppy.

MelchiorGaspar: :P yeah yeah.. so I fixed it for one game ;) :P  maybe I will fix it for another.. and another... maybe...
THAT WILL TAKE FOREVER!

PR #1380, opened 2017-09-23
spaces

cxd4:
Fixes a serious, catastrophic regression from #1378.


cxd4:
You write like a five-year-old.  You're hardly qualified to tell me what's easier to read, and my opinion doesn't matter either because it's GoodN64's standards.  The official Good Tools names have the spaces in between them, so it's their specification to invent arbitrary grammar rules on, not yours.

PR #1383, opened 2017-09-27
Fix Mischief Makers slowdown.

AmbientMalice:
Mischief Makers has a problem where the game essentially runs at half speed on level 2-3. (30fps instead of 60fps, and with literally halved speed.) Setting Counter Factor to 1 appears to solve it, but there may potentially be other side effects, so some testing is probably ideal. As discussed here:

https://github.com/mupen64plus/mupen64plus-core/issues/409

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -3504,18 +3504,21 @@ Good Name=Mischief Makers (E)
 Internal Name=MISCHIEF MAKERS
 Status=Compatible
 Plugin Note=[Glide64] some missing graphics
+Counter Factor=1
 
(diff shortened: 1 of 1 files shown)


LegendOfDragoon:
This reminds me of Top Gear Rally. I remember getting slowdowns in certain parts with CF=2.

Frank-74:
NASCAR 99 slows down badly at title screen with CF=1, and Monaco Grand Prix menu's with CF=1.  Both are fine with CF=2.  I set Monaco Grand Prix to CF=2 and OC=2.

PR #1384, opened 2017-09-27
Update Audio.rdb

Frank-74:
Added settings I found, only (U) versions added for now.

### DIFF
--- a/Config/Audio.rdb
+++ b/Config/Audio.rdb
@@ -1,1 +1,433 @@
 ﻿// ============ Audio RDB for Project64  ====================================
+[82DC04FD-CF2D82F4-C:45]
+Good Name=A Bug's Life (U)
+BufferDivider=60
+BufferLevel=2
+SyncAudio=1
+
+[E6A95A4F-BAD2EA23-C:45]
+Good Name=Aidyn Chronicles - The First Mage (U) [!] (V1.0)
+BufferDivider=60
+BufferLevel=2
+SyncAudio=1
+
+[112051D2-68BEF8AC-C:45]
+Good Name=Aidyn Chronicles - The First Mage (U) [!] (V1.1)
+BufferDivider=60
+BufferLevel=2
+SyncAudio=1
+
+[1FB5D932-3BA9481B-C:45]
+Good Name=Armorines - Project S.W.A.R.M. (U)
+SyncAudio=0
+
+[4C52BBB2-CEAB0F6B-C:45]
+Good Name=Army Men - Air Combat (U)
+SyncAudio=0
+
+[B20F73B6-2975FC34-C:45]
+Good Name=Army Men - Sarge's Heroes 2 (U)
+SyncAudio=0
+
+[A4BF9306-BF0CDFD1-C:45]
+Good Name=Banjo-Kazooie (U) (V1.0)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[CD7559AC-B26CF5AE-C:45]
+Good Name=Banjo-Kazooie (U) (V1.1)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[C2E9AA9A-475D70AA-C:45]
+Good Name=Banjo-Tooie (U)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[6AA4DDE7-E3E2F4E7-C:45]
+Good Name=BattleTanx (U)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[75A4E247-6008963D-C:45]
+Good Name=BattleTanx - Global Assault (U)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[EDF419A8-BF1904CC-C:45]
+Good Name=Beetle Adventure Racing! (U) (M3)
+BufferDivider=60
+SyncAudio=1
+
+[7C647C25-D9D901E6-C:45]
+Good Name=Blast Corps (U) (V1.0)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[7C647E65-1948D305-C:45]
+Good Name=Blast Corps (U) (V1.1)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[5326696F-FE9A99C3-C:45]
+Good Name=Body Harvest (U)
+BufferDivider=95
+SyncAudio=1
+
+[F568D51E-7E49BA1E-C:45]
+Good Name=Bomberman 64 (U)
+BufferDivider=60
+SyncAudio=1
+
+[F35D5F95-8AFE3D69-C:45]
+Good Name=Castlevania (U) (V1.0)
+BufferDivider=60
+BufferLevel=2
+SyncAudio=1
+
+[F35D5335-B7667CB7-C:45]
+Good Name=Castlevania (U) (V1.1)
+BufferDivider=60
+BufferLevel=2
+SyncAudio=1
+
+[4BCDFF47-AAA3AF8F-C:45]
+Good Name=Castlevania (U) (V1.2)
+BufferDivider=60
+BufferLevel=2
+SyncAudio=1
+
+[30C7AC50-7704072D-C:45]
+Good Name=Conker's Bad Fur Day (U)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[46A3F7AF-0F7591D0-C:45]
+Good Name=Cruis'n Exotica (U)
+
+[FF2F2FB4-D161149A-C:45]
+Good Name=Cruis'n USA (U) (V1.0)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[5306CF45-CBC49250-C:45]
+Good Name=Cruis'n USA (U) (V1.1)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[B3402554-7340C004-C:45]
+Good Name=Cruis'n USA (U) (V1.2)
+BufferDivider=30
+BufferLevel=2
+SyncAudio=1
+
+[DFE61153-D76118E6-C:45]
+Good Name=Cruis'n World (U)
+BufferDivider=60
+SyncAudio=1
+
+[DEE584A2-0F161187-C:45]
+Good Name=Destruction Derby 64 (U)
+BufferDivider=60
+SyncAudio=1
+
+[53D440E7-7519B011-C:45]
+Good Name=Diddy Kong Racing (U) (M2) (V1.0)
+BufferDivider=30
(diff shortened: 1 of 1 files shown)


theboy181:
every single (U) game? What about games that will be written in the future?

Only Joking. 

I'm still puzzled as to why N64 audio emulation needs per game settings. How do other emulators get audio right?

Another question I have about per game settings is what if a games FPS/DL's change does affect the audio, does it also change the needs for the audio settings? Are some games using dynamic audio solution to increase performance?

MelchiorGaspar:
maybe audio or some other advanced timings are coded in the ROM some where??
probably the only way... the N64 would know... but idk...

LegendOfDragoon:
> How do other emulators get audio right?

pro-tip, they don't :smile: 

>Are some games using dynamic audio solution to increase performance?

Some games do change fps in certain parts, but idk if it's for performance reasons or what.

loganmc10:
> pro-tip, they don't 😄

That's not really true. Try https://m64p.github.io (download the latest version and try it out, I implemented a better resampler this week, so whatever bit of crackling might have existed before should be gone). It gets the audio right 99% of the time. I think Hey You! Pikachu is the only game I know of that has terrible audio with mupen64plus. There are no per-game audio settings. There aren't even per-game VIRefresh settings anymore.

I think relying on massive RDB files with specific settings for each game is the wrong way to go. It's no secret that mupen64plus has benefited by implementing fixes from Project64. I think Project64 could benefit from looking at audio emulation in mupen64plus.

theboy181:
@loganmc10 is your calculation of 99% based on actual math,or an expression. I see a load of developers who make claims based on thought. 

Anyway, I am glad to hear that you have improved audio. Any improvement in N64 audio is a step in the right direction. 

To me audio has always been a bunch of unknown, and people who don't claim to be experts of it. Brings me back to why does the N64 audio scene lack good audio emulation.

loganmc10:
> @loganmc10 is your calculation of 99% based on actual math,or an expression. I see a load of developers who make claims based on thought.

Well I provided the link so you can try for yourself. I would call it an expression of actual math ;)

oddMLan:
Yeah, the RDB shouldn't even be necessary except for general purpose speedups I guess, like 32-bit mode and such.

Audio shouldn't have hardcoded values to run properly, a general-purpose solution should exist, that should estimate the correct values on-the-fly, without needing user interaction or a database with hard coded values to read from. There is a huge amount of N64 games and software, setting this on a per game basis would be insane. Wouldn't mind as much if it was just a few oddball games that needed fine-tuning.

oddMLan:
Also, don't know why you dislike FAT so much, it's a general purpose solution that works fantastically for me and fixes audio pops and such for the games I play (dunno about Clay Fighter, Goldeneye... games I don't usually open).

A general purpose solution that works 90% of the time (and its heuristics could always get improved) it's better than having hardcoded values. What if the timing in the emulator becomes more accurate, and these values no longer work? Maintenance burden.

Frank-74:
@loganmc10 I've tried it out.  Tested 3 games.  Body Harvest, crackles and video stutters, video/audio go way out of time with each other.  Goldeneye video stutters badly, crackles like hell because of it, then froze with black screen during after intro, just before level demo starts.  DK64, crackles, video stutters badly.
Recordings:
https://www.dropbox.com/s/tdvwu9l2zn0vqit/m64p_goldeneye.mp3?dl=0
https://www.dropbox.com/s/52bb9z11tfrd56l/m64p_bodyharvest.mp3?dl=0
https://www.dropbox.com/s/p6v2n7ilbv3tzrc/m64p_dk64.mp3?dl=0

@oddMLan Try my last build of Project64-Audio with built in settings for a number of games, list is in zip.  It also has a number of fixes not commited to git yet.  Turn FAT off.  DK64 needs Counter Factor set to 2 for correct timing.
https://www.dropbox.com/s/s4di6nqwxj8m132/Project64-Audio-Frank74.zip?dl=1

loganmc10:
> @loganmc10 I've tried it out. Tested 3 games. Body Harvest, crackles and video stutters, video/audio go way out of time with each other. Goldeneye video stutters badly, crackles like hell because of it, then froze with black screen during after intro, just before level demo starts. DK64, crackles, video stutters badly.

Obviously there is a lot more going on there than the audio. Does GLideN64 work for you in Project64? Somehow you and @LegendOfDragoon are the only people that can't seem to get it working properly...

Anyway doesn't matter, whether you two can figure out how to work your computers or not, what I'm talking about is the audio emulation code in mupen64plus, which does work well with on a decent computer/user combo. There seems to be more interest in tweaking RDB files than improving emulation over here, so I would say go ahead and merge it!

LegendOfDragoon:
>@loganmc10 is your calculation of 99% based on actual math,or an expression. I see a load of developers who make claims based on thought.

When people use estimates in the 90 percentile range, I've learned not to trust the claim :smile: . I would never claim something like that, until I actually tested enough to factually back the claim.

>It gets the audio right 99% of the time.

I wish it did, but it doesn't even come close on my end. I got crackling in a few games like WDC, Gauntlet, etc. Then I had stuttering in other games like Top Gear Rally and Deadly Arts (not sure if audio, gfx, or core is to blame though).

>There is a huge amount of N64 games and software, setting this on a per game basis would be insane.

I'd rather spend 2 minutes tweaking settings for each game, than to have to deal with crackling or stuttering. Ideally it would be nice to not have to do that, but there is no current better option.

Frank-74:
Project64 with my Project64-Audio, FAT off, latest WIP3 GLideN64 default settings.
There's one interruption in the audio when I got a message notification from Discord.
https://www.dropbox.com/s/xlwp48ispscqaz7/pj64-audio_dk64.mp3?dl=0
Goldeneye and Body Harvest are perfect too.

LegendOfDragoon:
>Somehow you and @LegendOfDragoon are the only people that can't seem to get it working properly...

What? GLideN64 works on my machine.

>what I'm talking about is the audio emulation code in mupen64plus, which does work well with on a decent computer/user combo.

So basically you're just going to assume it's somehow the user's fault?

>There seems to be more interest in tweaking RDB files than improving emulation over here

How is it "improving emulation", if it still sounds worse than other emulators?

oddMLan:
@Frank-74, if that's the case, I'm looking forward to the new timing method.

@loganmc10  I find that odd too. Probably because they are using laptop hardware with a crappy Intel IGP. Never really had an issue with the current audio timing personally, except for a couple of oddballs.

Really not a big fan of per-game hacks personally. One should work from what's accurate and then focus how to make it faster, not work with something already not accurate enough and focus on making it even more innacurate because it doesn't run good enough with your hardware. How about asynchronous audio? There would be no stutter.

At that point, why not writing a custom emulator/wrapper customized for every game?

Enhancements are a different story though, Widescreen and 60FPS are enhancements that are optional. Wonder if there's a way to make the jumbotron refresh at a higher rate with a cheat code instead of a plugin hack.

LegendOfDragoon:
>Probably because they are using laptop hardware with a crappy Intel IGP. 

My laptop only supports OGL 2.1, so I can't even test on that if I wanted to. So that's not the reason.

>One should work from what's accurate

Nothing is accurate for audio.

>At that point, why not writing a custom emulator/wrapper customized for every game?

That would actually be great, but would require much more time than tweaking audio settings.

>Really not a big fan of per-game hacks personally. One should work from what's accurate and then focus how to make it faster

Neither am I, but I will take a better experience over a worse experience any day.  A few mins spent tweaking, is much better than hearing crackling on a regular basis.

oddMLan:
I get what you mean, of course non-crackling audio > crackling audio, but at what cost? It feels like an incentive to keep hacking around things instead of actually solving them. Which might even break later on with other changes to the audio plugin or internal emulator timing.
Imagine if someone thought that Glide64 was already good enough with all its hacks and tried to hack the rest of the games to "improve" it (instead of reversing the uCodes and improving framebuffer emulation for a general purpose). Would GLideN64 be a thing now? Improving things on a surface level might be enough to fool end users but, what if you can't find the time to keep maintaining the project, and someone complained a particular game was broken at a certain point, which you have never played? Would you jump in and work around it? Tweaking the settings would work for you in particular, but doubt the end-user would try to tweak the values manually if the defaults didn't work good enough for them. So this kind of approach would make Project64 more of a simulator rather than an emulator.

oddMLan:
I guess I can compromise with these values but only if they are meant as temporary, but still I am not sure... gonetz has implemented hacks in GLideN64 and then replaced them with a general purpose solution later on.

Also they would need more testing, since these are hardcoded, these might not work at all points where the games change FPS. Or under a different hardware configuration. Can't test right now, no access to my computer till the next couple weeks.

LegendOfDragoon:
I have to say, this plugin seems to work pretty good for Smash Bros on my so-called "toaster" laptop :smile: . I'm not even using these RDB setings yet (doesn't seem necessary for SSB). Looks like Project 64's audio has good potential.

>Would you jump in and work around it?

I go out of my way to help end users :smile: . Even if I didn't have the time to personally tweak it for them, i could tell them what to tweak. Essentially, i can teach end users how to fish.

Tbh i'm not sure if all of Frank's changes are necessary. I have found that per-game settings are not necessary for your average game (with the right setup). Only reason I'd tweak a lot of games is to enhance my experience (minimizing latency and adjusting for possible issues like performance). For instance, I have to raise latency when using Angrylion's plugin for some games (Mario kart comes to mind).

>Which might even break later on with other changes to the audio plugin or internal emulator timing.

That's more of a problem for FAT. With FAT disabled, that's less likely to be an issue.

>It feels like an incentive to keep hacking around things instead of actually solving them. 

Literally nobody in this scene knows how the N64's audio truly works. People have tried their best, in order to come up with the current results. Getting audio to work right, is far more challenging than getting graphics to work. For graphics, you can refer to documentation and you're set for the most part. Audio doesn't have much documentation.

Frank-74:
> I have to say, this plugin seems to work pretty good for Smash Bros on my so-called "toaster" laptop :smile: . I'm not even using these RDB setings yet (doesn't seem necessary for SSB).

It needs buffer change for the plugin audiosync to keep around 60 with frame limiter off.  Without buffer changes, it goes above 60 which is potentially a crackle with frame limiter on.

theboy181:
@frank I’m not by a computer now, but when I tested last all the games would run faster or slower than 60vi with some of the settings. Has that been addressed?  Can you unlock the frame lint with this plugin?

Frank-74:
@theboy181 yes, audiosync is off by default in main branch at the moment.
You can turn it on/off with configure audio > game settings, and check/uncheck audiosync.
I'm just adding some of my improvements to a branch I just created at my fork.

Frank-74:
Done, you can view my changes here: https://github.com/project64/project64/compare/master...Frank-74:Project64-Audio-improvements
Test build here: https://www.dropbox.com/s/airfp9vy23w8sf4/Project64-Audio.zip?dl=1

PR #1385, opened 2017-09-28
Fix volume control

Frank-74:
DS volume is -10000 minimum volume, to 0 maximum volume.

### DIFF
--- a/Source/Project64-audio/AudioSettings.cpp
+++ b/Source/Project64-audio/AudioSettings.cpp
@@ -25,7 +25,7 @@ CSettings::CSettings() :
     m_AudioEnabled(true),
     m_advanced_options(false),
     m_debugger_enabled(false),
-    m_Volume(100),
+    m_Volume(0),
     m_BufferDivider(90),
     m_BufferLevel(4),
     m_SyncAudio(false)
@@ -74,7 +74,7 @@ void CSettings::RegisterSettings(void)
     m_Set_log_dir = FindSystemSettingId("Dir:Log");
 
     SetModuleName("Audio");
-    RegisterSetting(Set_Volume, Data_DWORD_General, "Volume", "Settings", 100, NULL);
+    RegisterSetting(Set_Volume, Data_DWORD_General, "Volume", "Settings", 0, NULL);
     RegisterSetting(Set_Logging_MD5, Data_DWORD_General, "MD5", "Logging", g_ModuleLogLevel[TraceMD5], NULL);
     RegisterSetting(Set_Logging_Thread, Data_DWORD_General, "Thread", "Logging", g_ModuleLogLevel[TraceThread], NULL);
     RegisterSetting(Set_Logging_Path, Data_DWORD_General, "Path", "Logging", g_ModuleLogLevel[TracePath], NULL);
--- a/Source/Project64-audio/ConfigUI.cpp
+++ b/Source/Project64-audio/ConfigUI.cpp
@@ -64,7 +64,7 @@ public:
         m_btnMute.SetCheck(!g_settings->AudioEnabled() ? BST_CHECKED : BST_UNCHECKED);
 
         m_Volume.Attach(GetDlgItem(IDC_VOLUME));
-        m_Volume.SetPos(100 - g_settings->GetVolume());
+        m_Volume.SetPos(g_settings->GetVolume());
         m_Volume.SetTicFreq(20);
         m_Volume.SetRangeMin(0);
         m_Volume.SetRangeMax(100);
@@ -74,7 +74,7 @@ public:
     bool OnApply()
     {
         g_settings->SetAudioEnabled(m_btnMute.GetCheck() != BST_CHECKED);
-        g_settings->SetVolume(100 - m_Volume.GetPos());
+        g_settings->SetVolume(m_Volume.GetPos());
         FlushSettings();
         return true;
     }
--- a/Source/Project64-audio/Driver/DirectSound.cpp
+++ b/Source/Project64-audio/Driver/DirectSound.cpp
@@ -132,10 +132,11 @@ void DirectSoundDriver::SetFrequency(uint32_t Frequency)
 
 void DirectSoundDriver::SetVolume(uint32_t Volume)
 {
-    /*DWORD dsVolume = ((DWORD)Volume * -25);
+    LPDIRECTSOUNDBUFFER & lpdsb = (LPDIRECTSOUNDBUFFER &)m_lpdsb;
+    DWORD dsVolume = ((DWORD)Volume * -25);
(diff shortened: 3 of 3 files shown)


project64:
I do find it strange that the setting in the cfg file has volume =0 as max volume, while volume = 100 as mute strange.

Frank-74:
I'll work it out so it's 0 min, 100 max.  I did it for XAudio2 volume before, but Azimer changed it to be same as other backends.

project64:
I still think it is backwards, but I will merge and fix later if it is

cxd4:
0 is the max volume in the DirectSound API, but the minimum is a negative number and not positive.  I guess plugins like this and AziAudio used +100 because the parameter is a DWORD?

>The volume is specified in hundredths of decibels (dB). Allowable values are between DSBVOLUME_MAX (no attenuation) and DSBVOLUME_MIN (silence). These values are currently defined in Dsound.h as 0 and -10,000 respectively. The value DSBVOLUME_MAX represents the original, unadjusted volume of the stream. The value DSBVOLUME_MIN indicates an audio volume attenuated by 100 dB, which, for all practical purposes, is silence. Currently DirectSound does not support amplification.

https://msdn.microsoft.com/en-us/library/mt708939(v=vs.85).aspx

Frank-74:
@cxd4 Yes, that's right, I had to use float before.
Cheers.

Frank-74:
I'll close and redo this after some sleep, it's too late (early!) to do now.  6:30am...

cxd4:
That would simplify down to `dsVolume = -100.0f * volume;`, and yeah that's correct.
(The uint32_t volume can be implicitly increased to float precision without any warnings, I think.)

That works for DirectSound, but I don't know what OpenSL ES uses.  I do seem to remember OpenAL using the same geometric kind of vector coefficient for audio as what you would expect from OpenGL.

So this all works well and good, but do we really want the Project64 audio plugin's central `volume` parameter to be on the 0 (max) to 100 (min) scale?  I forgot why AziAudio does this...why not just pass -100 instead of 100 when we want it muted?

So
```c
 void DirectSoundDriver::SetVolume(long decibels)
 {
    long dsVolume;
    LPDIRECTSOUNDBUFFER & lpdsb = (LPDIRECTSOUNDBUFFER &)m_lpdsb;

    dsVolume = 100 * decibels; /* (GUI_volume * -25) */
    if (dsVolume < DSBVOLUME_MIN)
        dsVolume = DSBVOLUME_MIN;
    if (dsVolume > DSBVOLUME_MAX)
        dsVolume = DSBVOLUME_MAX;

    if (lpdsb != NULL) {
        lpdsb->SetVolume(dsVolume);
    }
}
```

cxd4:
Your original statement:

>`float dsVolume = 1.0f - ((float)volume * 100.0f) -1;`

Simplified:
```c
float dsVolume = 1.0f - ((float)volume * 100.0f) - 1;

float dsVolume = 1.0f - 1 - ((float)volume * 100.0f);

float dsVolume = 0 - ((float)volume * 100.0f);

float dsVolume = -(float)volume * 100.0f;
```

Frank-74:
Just had a play with it.  *100.0f is way too much, volume is inaudible at half way mark.

This works well:
void DirectSoundDriver::SetVolume(uint32_t Volume)
{
	LPDIRECTSOUNDBUFFER & lpdsb = (LPDIRECTSOUNDBUFFER &)m_lpdsb;
	float dsVolume = - (float)Volume * 33.0f;
	if (Volume == 100) dsVolume = DSBVOLUME_MIN;
	if (Volume == 0) dsVolume = DSBVOLUME_MAX;
	if (lpdsb != NULL) lpdsb->SetVolume((DWORD)dsVolume);
}

With volume at 1, it's just about audible.

cxd4:
My function takes an input of dB decibels.

Seems the "volume" passed to the Project64 audio function is of a different metric...GUI units in some Win32 slider?  Hence the * 25 instead of * 100.  It looks like after converting it by * 100 you have to divide it again by 4 to account for the precision in whatever GUI thing is selecting this volume.

Either way, Microsoft specifically states the DirectSound function to set the volume reads from hundredths of a dB, so you take decibels times 100.  Clearly decibels isn't the unit Project64 is inputting.

I guess what zilmar would like is percentage.  So SetVolume(unsigned int per_cent) and if per_cent is 0 it comes out to -100 dB = DSBVOLUME_MIN; if per_cent is 100 it comes out to 0 or DSBVOLUME_MAX.

Frank-74:
Think it's because dB is a logarithmic scale.  So decreasing volume using a linear scale, lowers the audible volume too quickly.  Gui slider is from 0 to 100.

cxd4:
If that is true then there are so many things wrong with keeping `DWORD dsVolume = ((DWORD)Volume * -25);` in the function.   GUI slider is from 0 to 100 so wtf does multiplying anything by -25 accomplish?  Then we have checks if the original volume is 0 make it DS_MAX, or 100 make it DS_MIN...when * -25 should be doing that for us.  OWAIT!  lolwut...I need some sleep too now.

Frank-74:
https://en.wikipedia.org/wiki/Decibel

cxd4:
But the slider is from 0 to 100 and zilmar thinks 100 should be max, so he wants percentage.

Therefore the coefficient of -25 is still retarded.

Yes it should be negative because DirectSound volumes are always in hundredths of non-positive decibels (0 being unaltered plain volume, -100 dB being the maximally silenced volume), but it shouldn't be 25 because the DirectSound range is -10000 to 0.  -25 might account partially for the logarithmic scale that dB uses which you addressed, but it's abrasive to the two `if` statements trying to say 100 should be -10000.

cxd4:
As far as account for the exponential nature of adjusting by decibels, that's a fundamentally different question altogether.  Multiplying by either -25 or -100 don't tackle that issue because multiplying by a constant is still linear adjustment and not logarithmic.  Raise it to a power or something maybe...but either way, -25 is dumb.

Frank-74:
Found a good article to get something more accurate.
http://www.playdotsound.com/portfolio-item/decibel-db-to-float-value-calculator-making-sense-of-linear-values-in-audio-tools/

Frank-74:
This works ok:
```cpp
void DirectSoundDriver::SetVolume(uint32_t Volume)
{
	LPDIRECTSOUNDBUFFER & lpdsb = (LPDIRECTSOUNDBUFFER &)m_lpdsb;
	int32_t dsVolume = - 10 ^ (Volume * 25);
	if (Volume == 100)
	{
		dsVolume = DSBVOLUME_MIN;
	}
	if (lpdsb != NULL)
	{
		lpdsb->SetVolume((DWORD)dsVolume);
	}
}
```

PR #1386, opened 2017-09-28
Fix Volume control

Frank-74:
0 muted volume, 100 max volume.

### DIFF
--- a/Source/Project64-audio/Driver/DirectSound.cpp
+++ b/Source/Project64-audio/Driver/DirectSound.cpp
@@ -132,10 +132,16 @@ void DirectSoundDriver::SetFrequency(uint32_t Frequency)
 
 void DirectSoundDriver::SetVolume(uint32_t Volume)
 {
-    /*DWORD dsVolume = ((DWORD)Volume * -25);
-    if (Volume == 100) dsVolume = (DWORD)DSBVOLUME_MIN;
-    if (Volume == 0) dsVolume = DSBVOLUME_MAX;
-    if (lpdsb != NULL) lpdsb->SetVolume(dsVolume);*/
+    LPDIRECTSOUNDBUFFER & lpdsb = (LPDIRECTSOUNDBUFFER &)m_lpdsb;
+    int32_t dsVolume = - 10 ^ ((100 - Volume) * 33);
+    if (Volume == 0)
+    {
+        dsVolume = DSBVOLUME_MIN;
+    }
+    if (lpdsb != NULL)
+    {
+        lpdsb->SetVolume(dsVolume);
+    }
 }
 
 void DirectSoundDriver::SetSegmentSize(uint32_t length, uint32_t SampleRate)


Source/Project64-audio/Driver/DirectSound.cpp
```diff
@@ -132,10 +132,16 @@ void DirectSoundDriver::SetFrequency(uint32_t Frequency)
 
 void DirectSoundDriver::SetVolume(uint32_t Volume)
 {
-    /*DWORD dsVolume = ((DWORD)Volume * -25);
-    if (Volume == 100) dsVolume = (DWORD)DSBVOLUME_MIN;
-    if (Volume == 0) dsVolume = DSBVOLUME_MAX;
-    if (lpdsb != NULL) lpdsb->SetVolume(dsVolume);*/
+    LPDIRECTSOUNDBUFFER & lpdsb = (LPDIRECTSOUNDBUFFER &)m_lpdsb;
+    int32_t dsVolume = - 10 ^ ((100 - Volume) * 33);
```
cxd4: I recognize that you intend to use the logarithmic solution from that article you linked to.

But you do realize that `-10 ^ ((100 - volume) * 33)` is not the same as `pow(-10, (100 - volume) * 33)`?

I seriously doubt you intended to use the bit-wise XOR operation in C to simulate a logarithmic scale effect, because that's what the `^` operator in C does.  XOR would just completely scramble the result of 33*(100 - v) by the binary mask 1010 and has no mathematical significance towards the result's exponentiation.

cxd4: I recognize that you intend to use the logarithmic solution from that article you linked to.

But you do realize that `-10 ^ ((100 - volume) * 33)` is not the same as `pow(-10, (100 - volume) * 33)`?

I seriously doubt you intended to use the bit-wise XOR operation in C to simulate a logarithmic scale effect, because that's what the `^` operator in C does.  XOR would just completely scramble the result of 33*(100 - v) by the binary mask 1010 and has no mathematical significance towards the result's exponentiation.

cxd4:
I'm trying to think of a solution that does both the logarithmic algorithm you want while exploiting the entire legal range of values for DirectSound volumes in decibels.

Maybe this is the problem with my approach, is that the two objectives are conflicting.  The way DirectSound has 0 dB as the max volume (no silencing) and -100 dB (or -10000 hundreds of a dB) as a minimum suggests that -100 dB is severely lower than what's necessary to mute the audio in a Nintendo 64 game.  If something like even -10 dB is sufficient to mute quiet N64 audio, then that would explain why my earlier suggestion of just doing `-100 * dB` made the slider behave too insensitively.

cxd4:
It's a shame this wasn't OpenAL, because then this would be a cinch.
```c
void change_volume(signed int per_cent)
{
    ALfloat master_volume;
    ALfloat min = 0.0, max = 1.0;

    master_volume = per_cent / 100.0f;
#ifdef CLAMP_MASTER_TO_SOURCE_CHANNEL
    alGetSourcef(source, AL_MIN_GAIN, &min);
    alGetSourcef(source, AL_MAX_GAIN, &max);
    if (master_volume < min)
        master_volume = min;
    if (master_volume > max)
        master_volume = max;
#endif
    alListenerf(AL_GAIN, master_volume);
}
```

cxd4:
```
f(0) =  0 =  1 - 1
f(1) =  1 =  2 - 1
f(2) =  3 =  4 - 1
f(3) =  7 =  8 - 1
f(4) = 15 = 16 - 1
...
f(100) = 10000 // abs(DSBVOLUME_MIN)
...
f(x) = 2^x - 1
```

(Of course, here, my usage of `^` is intended to represent raising to a power, as virtually all ASCII math writing on the Internet does.  It's just that we can't represent exponents in C that way.)

Implementing the fact that all nonzero volumes are silencers from 0 being the maximum volume:
```
f(x) = -(2^x - 1) = 1 - 2^x
```

Implementing the fact that zilmar wants 100 to represent max volume and 0 to represent min:
```
f(x) = 1 - 2^(100 - x)
...
f(87) = 1 - 2^13 = -8191
...
f(0) = -2^100 + 1 = a very, very long number
g(0) = -10000 # clamped to DSBVOLUME_MIN
```

cxd4:
This becomes disadvantageous as well because you see base 2 quickly exceeding 10,000 after only 13 iterations of incremental input, out of 100 values for the GUI slider in Project64.

With that being said, the solution of using base 10 in Frank74's logarithm will of course exceed that limit only even faster, so using base 2 for the exponent is the best way to cover as many possible values between 0 and -10,000 while maintaining the logarithm approach.

In short, this type of logarithm approach is severely flawed and will never work.

cxd4:
I suppose you could solve:
```
f(0) = 1 - b^100
// We want f(0) in PJ64's GUI slider to be -10000, the full muteness DirectSound allows.
1 - b^100 = -10000
b^100 - 1 = 10000
b^100 = 10001
```
That would give you the right base--quite a small number in between 1.0 and 2.0--that would have logarithmic functionality with the decibel unit's exponential nature in the audio slider GUI.  It wouldn't necessarily be comfortable; I still propose just having a linear scale where you can just take -100 * x and call it a day.  That's what PulseAudio's volume slider in Ubuntu Linux did...even though decibels are logarithmic, the GUI slider was a linear scale where the volume in dB was dragged through a plain ratio.

Frank-74:
If you make the volume decrease to -10000, it will be silent by a third of the way down.....
-10000 is Directsound's mute value of -100dB.  -33 dB is quite enough just above the bottom.  That's why we have the if Volume == 0 to mute.

I changed it to -20dB and it's nice and smooth.  It works.
```cpp
	int32_t dsVolume = -((100 - (int32_t)Volume) * 20);
	if (Volume == 0)
	{
		dsVolume = DSBVOLUME_MIN;
	}

cxd4:
True, but if we cap the minimum to something like -4000 instead of -10000, then the power used in the approach of -4000 = b^100 will be even weaker than the solution power in -10000 = b^100, causing a greater duplication of the same integer-rounded values for the volume across a 100-unit GUI slider.  But like you said, -10000 is too silent to drag all the remaining values all the way out to, so you can't win either way with the exponent approach.

Yeah, the linear approach is probably best.  Something like what you just wrote, except that caps the minimum to -2000 instead of -10000 (which isn't bad necessarily...I know you're going to do a final `if` statement to override 0% to DSBVOLUME_MIN or -10000, but the slider value right before -2000 might be more audible in other games).

Another thing to consider is the actual source volume in the games you are testing.  Most N64 ROMs have quiet audio (in terms of dB) compared what you'd hear in death metal MP3s.  You have fewer decibels to subtract from to begin with if the audio in the N64 game you are testing is quiet.

cxd4:
If we wanted to do anything else but the linear style you just instantiated, it would have to be some kind of polynomial function that somehow decelerates the change in dB the lower the volume gets (for minimum coverage of extremely negative values, to account for the logarithmic nature of dB units) and accelerates the change in dB the higher the volume gets (for maximum coverage of the negative values closest to 0 or max volume, again to account for the insignificance in change between decibels close to zero).

PR #1393, opened 2017-10-14
[RDB] Add CF=1 to Pokemon Snap

Enker:
This fixes an issue that is only visible with angrylion's plugin. With CF=2, the red dot will eventually keep appearing/beeping when you are holding Z and looking at the ground or a wall. It happens in every level. In the first level, it starts once you pass the first Meowth. I tested the (E)(J)(U) versions.

I also fixed the save type for two other games.


project64:
there is some changes to Save Type=16kbit Eeprom was that intended?

Enker:
Yes, I added those in the commit since they are small changes. I tested both games.

PR #1396, opened 2017-10-17
[RDB] Save type fixes

Enker:
I checked each of these games and none of them save with FlashRam. NBA Showtime uses the Controller Pak, and the other games use 4kbit Eeprom.

I also fixed Kuru Kuru Fever's CRC1/2 and title.


project64:
why is Guru - Kuru Kuru Fever (J) [ALECK64] in the H section.

Also how did you test it to know that it does not use it?

Enker:
It looks that way in the preview, but it's actually in the K section :)
くるくるフィーバー is on the title screen, Guru ( ぐる) is not present.

Enker:
I added two more fixes. Now both games can save to all four files instead of just file 1.

Enker:
I added fixes for two more games. Those should be the last changes for this PR.

Issue #1290, opened 2017-03-19
FAT breaks RNG on boot


theboy181:
Fixed Audio Timing breaks RNG in PJ64. I have been reading that people are saying FAT is more accurate but this is just not true.

Tested Games with FAT = 0

SSB64  - Character pulled from the toy box will always remain the same every load of the game.
MK64  - Items boxes are not random, a high amount of shrink items given during multi player sessions. (investigating)
SWE1R - Sebulba is in the intro cut scene each boot, instead of Anikin.

I have a feeling that using FAT is just nonsense when it comes to RNG, and a disclaimer should warn its users. FAT on by default is bad for the uninformed.

Frank-74:
FAT has been made worse by optimizations since FAT was first introduced,  Now, it's no benefit.  Because timing has been changed by optimizations.

Latest 2.4, I cannot play Body Harvest at all.  It slows down to 30-40VI's and crackles like hell.  My 2.3 build with #984 reverted is ok.  It runs full speed mostly.  I don't know why my latest 2.4 build performs so bad.  Not much changed.

theboy181:
Have you gone back to check this claim with prebuilds? or is this an assumption..

oddMLan:
You clearly have no idea what you are talking about. FAT is actually _more accurate_ emulation.

theboy181:
@oddMLan you clearly do.. enjoy accuracy.. 

But instead of arguing please see that RNG is killed when FAT is enabled, and that is a deal breaker.  If you still think that this isn't an issue I would love to debate this. I want to hear how you come out looking like I'm not correct..

I would expect FAT fantasies to stop being reverential to accuracy or audio emulation goals.

I have tested, and I  am sure with testing you will have to agree that RNG is needed to function properly for enjoyable game play in MK64 MP.

oddMLan:
Sure I do see your points in the OP, but correlation doesn't always equal causation. There are many timing hacks in the core that prevent proper diagnosis of these issues. For intance, I'm not saying FAT is _entirely_ accurate, but it is definitely more accurate than no FAT cause it fixes so many timing issues. I'm thinking a game could change the AI value at a particular moment, or that the AI value is "close enough" but off by some margin that causes differences in emulation than in the console. This seems to be a more sound hypothesis on why the RNG seems to be affected by it.

[...]

PR #1397, opened 2017-10-17
Randomization for SI and PI interrupts

Azimer:
Change basic PI and SI read interrupts to be random.  Should help issue #1290


project64:
how this works in the pull request will break the sync core .. 

if it is to be done, at start up create an array of random values. Send that to the main core and the sync core so they have the same values. Then have an index in the array as you use each move to the next one. If you reach the end, loop back to the start. 

So given the same random seeds + inputs would result in the exact same behaviour. It also means if netplay is ever put in then those values would have to be shared and each version use the same seeds.

PR #1407, opened 2017-10-26
Added UpdateVersion PowerShell script.

JunielKatarn:
PowerShell alternative to the CMD-based UpdateVersion script.

Advantages:
- Does not require `sed`, or any other external dependencies.
- Easier to read and maintain.

### DIFF
--- a/Source/Script/UpdateVersion.ps1
+++ b/Source/Script/UpdateVersion.ps1
@@ -1,22 +1,22 @@
 Param
 (
-	$SolutionDir = $(Get-Location),
+	$SolutionDir = (Split-path (Split-Path ($PSScriptRoot))),
 	[Parameter(Mandatory=$true)]
 	[string] $Version,
 	[string[]] $Files = (
-		'Source\Project64-core\version.h',
-		'Source\nragev20\version.h',
-		'Source\RSP\version.h',
-		'Source\Project64-video\version.h'
+		"$SolutionDir\Source\Project64-core\version.h",
+		"$SolutionDir\Source\nragev20\version.h",
+		"$SolutionDir\Source\RSP\version.h",
+		"$SolutionDir\Source\Project64-video\version.h"
 	)
 )
 
 $versionArr = $Version.Split('.')
 foreach ($file in $Files) {
-	(Get-Content "$SolutionDir\$file") `
+	(Get-Content $file) `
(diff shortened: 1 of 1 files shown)


project64:
yes cmd processing is very weak, something like power shell might be better. 

you use Get-Location for the solution dir, but where is the location set?
where is the call to git describe, how is this meant to be used?

JunielKatarn:
Yeah, just noticed that is a bug. I'll update in a few minutes.

project64:
one of the suggestion was to run update version on building, if it is a power script not sure it is going to work on most pc right out of the box where a cmd file will.

project64:
when I try to run I get "UpdateVersion.ps1 cannot be loaded because running scripts is disabled on this system. For more information, see about_Execution_Policies at http://go.microsoft.com/fwlink/?LinkID=135170."

Sure this will be fine for the build bot, I can set up permissions .. the running as part of the build process not so great

JunielKatarn:
That you can fix with a one-off setting.
From an Administrator (elevated) PowerShell, run this:
```
Set-ExecutionPolicy -ExecutionPolicy Bypass -Scope CurrentUser
```
This allows your user to run any PowerShell script.

> if it is a power script not sure it is going to work on most pc right out of the box where a cmd file will.

PowerShell works on all PCs with Windows XP or later (XP/2003 need to install it, though).

project64:
I am really not sure about PowerScript .. Yes something better then a cmd that has to use external tools .. maybe I should write a C++ app to do it .. Maybe a VBS script could work. 

tho I suspect I am probably better off  writing an app and have a cmd that works something like:
FOR /F "tokens=1 delims=" %%A in ('git describe --tags --long') do SET current_tag=%%A

call %base_dir%\bin\version_setter.exe "%base_dir%\Source\Project64-core\version.h" !current_tag!
call %base_dir%\bin\version_setter.exe "%base_dir%\Source\nragev20\version.h" !current_tag!
call %base_dir%\bin\version_setter.exe "%base_dir%\Source\RSP\version.h" !current_tag!
call %base_dir%\bin\version_setter.exe "%base_dir%\Source\Project64-video\version.h" !current_tag!

JunielKatarn:
Whatever works.
I favor PowerShell because it's included in the OS, and it can do anything the .NET Framework can do (it's essentially implemented on top of that framework).

Your choice, but this script shell could save you quite some time (I know it's been my case).

PR #1411, opened 2017-11-13
fixed 64-bit memory pointer warnings

cxd4:
Cast pointers to integers only of the same size as the integer type big enough to hold a pointer address.


cxd4:
The `#include` stddef for size_t and typedef of size_t was supposed to go under the `#ifdef` for _MSC_VER and not the `#else` because stdint already determines the definition of uintptr_t.

Did a force push to overwrite that commit.  It was 2 AM and I was playing with moving the emulation of uintptr_t outside of the existing if/else and using the C99 limits.h UINTPTR_MAX macro instead.

project64:
closing old pull requests

PR #1414, opened 2017-11-28
Ensure UpdateVersion is built before dependencies.

JunielKatarn:
Updates build logic to have UpdateVersion.exe built before UpdateVersion.cmd is invoked.


LaLoLZ:
Big crash fix:

image

PR #1416, opened 2017-12-17
Added conditional switch to UpdateVersion.

JunielKatarn:
Added a switch to run UpdateVersion (or not).
Problem: For fork-based clones of Project64 (i.e. `git@github.com:<other user>/project64`) the command `git describe --tags --long --dirty` yields the following:
```
fatal: No names found, cannot describe anything.
```

This makes UpdateVersion.cmd fail inevitably.

This change adds the MSBuild property 'UpdateVersion' (defaulting to false) to control wether run the tool or just copy the Version.h.ini contents.
To build running the tool again, use the following command line:
```
msbuild /p:UpdateVersion=true
```


project64:
you are getting that error cause you have no tags/releases. When I forked gliden64 (https://github.com/project64/GLideN64/releases). I have all there releases. Did you do something that deleted the tags/releases or was there an option that needed to be selected to take the tags as well?

JunielKatarn:
I did nothing related to tags. I'll check if this also happens to other contributors' forks.

project64:
If I clone from the project64 github page it pulls down the tags .. tho if I push to a new repository, I have an option to push the tags as well, it might not push the tags with out specifying to

JunielKatarn:
Yes. That works and solves the missing tags issue. I gues I forked your repo way before you had the 2.x tags.
That said, I still think there should be a way to override the versioning mechanism and set custom version numbers.
For instance, I want to keep my own versioning scheme in my builds.
I'll work further on this and re-submit this PR.

project64:
in part there was like emucr and other people download the source and compiling and getting 2.4.0.999 builds so you could not tell where the commit that build came from, which is why I forced the building of the version number in to the build

Issue #1412, opened 2017-11-15
Build fails.


JunielKatarn:
As of commit 1178b5c73fa054c329c11d769374e5d50928eaea, a regular (x86|Debug) build fails with the following message:
```
"D:\Extra\pj64\buildme\Project64.sln" (default target) (1) ->
       "D:\Extra\pj64\buildme\Source\nragev20\NRage_Input_V2.vcxproj" (default target) (8) ->
       (PreBuildEvent target) ->
         C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.CppCommon.targets(123,5): error MSB3073:
        The command ""D:\Extra\pj64\buildme\Source\Script\UpdateVersion.cmd" "Debug" "D:\Extra\pj64\buildme\Source\nragev20\Version.h.in" "D:\Extra
       \pj64\buildme\Source\nragev20\Version.h"\r [D:\Extra\pj64\buildme\Source\nragev20\NRage_Input_V2.vcxproj]
       C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.CppCommon.targets(123,5): error MSB3073: :
       VCEnd" exited with code 1. [D:\Extra\pj64\buildme\Source\nragev20\NRage_Input_V2.vcxproj]
```

It may be related to this program: `UpdateVersion.exe`.

project64:
hopefully the last few commits fix it

JunielKatarn:
Some errors were fixed, but seems like Android now breaks:
```
"D:\Build\pj64\buildme\Project64.sln" (default target) (1) ->
"D:\Build\pj64\buildme\Source\Android\Bridge\Bridge.vcxproj" (default target) (4) ->
(ClCompile target) ->
  jniBridge.cpp(19): fatal error C1083: Cannot open include file: 'Project64-core/Version.h': No such file or directory [D:\Build\pj64\buildme\Source\Android\Bridge\Bridge.vcxproj]
```

As a side note, if Android is actually built with another set of tools (Android Studio, right?), why not just leave it out of the Visual Studio build? It would make things faster and easier to maintain on the Windows side.

project64:
sounds like build order, need to make sure Bridge.vcxproj has a dependency on Project64-core so that the version is created first before bridge is built

project64:
[...]

PR #1417, opened 2017-12-18
Allow manual version settings for plugins or main app.

JunielKatarn:
Allow custom version numbers to be set as MSBuild properties for each of the following:
- Audio plugin
- Video Plugin
- Input Plugin
- RSP Plugin
- Main App

Setting a custom version number the UpdateVersion.cmd and instead passes the version segments (major, minor, etc) as preprocessor values.
Example usage:
```
msbuild /p:RspVersion=2.3.5.0001
```

The command above sets the RSP plugin version to `2.3.5.0001`, but the rest of the plugins are versioned by `UpdateVersion.cmd`.

Also useful for builds that require their own build number.

Going forward, this approach could be followed by the "official" builds. Setting the version numbers implicitly, but without the need of copying/writing headers.


JunielKatarn:
Serves as a possible workaround for #1412.

PR #1460, opened 2018-05-30
RDB: HLE support for Indiana Jones and Battle for Naboo

Lithium64:
GlideN64 now supports HLE graphics for Indiana Jones and Star Wars: Battle for Naboo, i've also made other minor improvements on RDB.

http://gliden64.blogspot.com/2018/05/hle-implementation-of-microcodes-for.html


Frank-74:
I've just experimented with Indiana Jones (U) settings.  Now no slowdowns, crackling audio or crashes, audio/video in sync.  The demo's cycle to the end, instead of crashing at Sophia's tent.  Set Recompiler's Function lookup to Virtual table and turn off Protect Memory.  Change VI to 1400 and Overclock=2.
FuncFind=2
VIRefresh=1400
OverClockModifier=2
Remove SMM-Protect=1

PR #1481, opened 2018-08-07
Use 8MB RDRAM by default

shygoo:
Less hassle for romhackers and romhack players

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -181,6 +181,7 @@ Plugin Note=[Glide64] framebuffer:screen
 Culling=1
 Emulate Clear=1
 Linking=Off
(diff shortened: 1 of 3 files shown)


oddMLan:
I was thinking of about this. Some hacks will require more than 8MB. It would also be nice to make the cfg file read the RDRAM size parameter in MB instead of bytes. And the only way to edit that currently is through the cfg. It is a little confusing to have the RDB read the RDRAM size in MB while the cfg in bytes.

oddMLan:
Maybe instead of adding redundant RDRAM Size=4 to the RDB, I guess something better would be to change the logic to: "if there's no entry in the ROM Database, default RDRAM Size to 8MB".

PR #1484, opened 2018-08-28
Added Kaillera support and Goldeneye/Perfect Dark Mouse Injector support

click4dylan:


theboy181:
Can these changes be be implemented into the plug, with minimal changes to PJ64 being needed?

click4dylan:
no it cannot, unfortunately

ZaxLofful:
This doesn't sound like a wise idea IMO. If it could be a plugin, that would be nice. I think Zilmar should weigh in

PR #1509, opened 2018-12-04
[Audio] Code cleanup and slightly smaller buffer.

Frank-74:
NTSC consists of ​30 / 1.001 (approximately 29.97) interlaced frames of video per second. Each frame is composed of two fields, a field refresh frequency of ​60 / 1.001 Hz (approximately 59.94 Hz).

I added +1 to the dacrate in the buffer_size = audio clock / dacrate calculation.


project64:
is this ready to merge now?

Frank-74:
I've tested with several games.  Slight improvement.  Seems ok to merge now.
Increased timer resolution has certainly helped a lot.  Star Wars games need audio rdb updating.  But I'll do that separately.

PR #1517, opened 2018-12-09
[RomBrowser] Show Internal Name instead of Bad ROM warning

oddMLan:
When a ROM is not in the Rom Database, Project64 would display under "Good Name" the following text: "Bad ROM? Use GoodN64 & check that the RDB is up-to-date."

Now it shows the ROM's Internal Name instead.
Additionally, I moved the Bad ROM warning to the "Notes (core)" column.

Now, the Bad ROM warning will serve the same purpose as before, while making ROMs which are not in the RDB more easily identifiable.


oddMLan:
Before:
image

After:
image

PR #1529, opened 2018-12-25
Removed nagware from the Windows interface

Clownacy:


CrashOveride95:
Why though

theboy181:
Less work to just donate to an awesome emulator developer. Happy Xmas zilmar!

CrashOveride95:
I have no issue with the nagware, don't see why it needs to be removed

theboy181:
image
So sad

cxd4:
I buy Himalayan pink salt.

Has the following benefits over white table salt:

* pricier but mined from the Himalayas where warm sea water (for sea salt) mercury contamination is out of the question
* more mineral nutrients and more _of_ them with pink salt than white salt
* stronger concentration of flavor per unit of Himalayan salt per less sodium consumed...so harder to exceed daily recommended sodium intake when getting the same flavor

Iodized sea salt may be the next best thing if you can't afford the pink salt, but water is contaminated with inorganically present minerals all the same, even if you buy it purified inside some cheap plastic bottle off the store shelf.  So just buy a home distiller...or reverse osmosis purifier may be cheaper if you're one of those opponents of drinking distilled water.

Issue #1531, opened 2018-12-27
Latest Build of PJ64 Debugger Woes 12/27/2018 (PR freeze please)


theboy181:
I have noticed that almost every PR that has been accepted recently, has broke something, I use PJ64 for debugging mostly, so it has been quite disruptive lately. I have created a NEW PJ64 directory to see if its my config.

I have been reluctant to reporting on GIT lately as @project64 has not been that active, but if we are accepting PRs I think we really need to start looking at the bugs they are creating before accepting anymore. I would like to suggest a PR freeze until its stable again.

@shygoo  and @project64 can you test this one? 

On latest PJ64 Build v2.4.0-758-geddd90b. (fresh folder)

1.) Open Project64.exe 
2.) Options>Settings> Uncheck "Hide advance settings"
3.) Options>Advanced> "Enable debugger", and "Always use Debugger"
4.) File>"Choose ROM Directory..." (choose your path)
5.) Open a ROM (I opened CyberTiger (U) please use this to do testing )
6.) Debugger>"Breakpoint..."
7.) Change the Memory address to "80020354" (note if you copy this and past it you will see that PJ64 will not jump to the address, you will need to scroll Command list first then paste, or manually type it in) -Bugged for a while.
8.) Right Click the Instruction @ 0x80020354>"Edit code"
9.) Observe that you can no longer edit ASM.

This is very sad news.

There are many other issues with the ROM Browser, and Audio, and config since the latest PR's and I will do my best to report them. 

I do hope that we can get things back to normal soon.

theboy181:
Looks like https://github.com/project64/project64/commit/d7ad7be6cb38546c477f6743c2d7bb9dd2b33421 is the culprit for the ASM edit breakage..

project64:
can I close this now?

PR #1532, opened 2018-12-27
[Debugger] Fix opcode editor in the commands window

shygoo:
Addresses #1531


theboy181:
:) Thanks Shygoo. Were you able to fix the copy past issue too?

PR #1535, opened 2018-12-29
[Rom Browser] Name column - Use Filename in absence of Good Name & hide file extensions

oddMLan:
@theboy181 brought up a good point that the internal name more often than not isn't helpful enough to distinguish a ROM when an user has many hacks of the same ROM, OR the Internal Name is empty (the case of special ROMs such as Aleck64 and iQue).

Additionally, provided an option in the UI to show the extension at the end of the filename, disabled by default. Rom Browser refreshes as it should when toggling this option.

Before
imagen

After
imagen

imagen


theboy181:
Thumbs up! Thanks.

cxd4:
>OR the Internal Name is empty (the case of special ROMs such as Aleck64 and iQue).

That's not a special case.  The 20-char name inside the ROM header is part of Nintendo's market registration data for tracking and isn't really supposed to be set necessarily for anything that's not a commercial game.  ROMs without an "internal name" as you call it are exceedingly common.

Wherefore I chiefly used the "File Name" column from the ROM browser configuration settings and only kept "Good Name" or "Internal Name" columns for database and sorting purposes.  Not sure what you guys did to that since I last looked at the source to this repository when it was 3 separate columns for like 20 years instead of being bound to just a single generalized "Name" column that I see in the screenshots.

theboy181:
I had mine setup to show the files name in a separate colum before too. It’s nice to have this option as it reduces the columns needed. 

Great addition and this improves something that has not been ideal for year while leaving the ability to have it the old way if you still wish. 

Hats off to you sir

PR #1536, opened 2018-12-29
Add Unaligned DMA options to Defaults and Game Settings

oddMLan:
After f0c9ea887e62712e411133e3b3c9e4629db5d9fe was merged many Romhacks stopped working. The commit was a needed accuracy fix but many play these romhacks using older versions of PJ64 because they don't work in the latest. This introduces an optional hack that allows these games to work again.

Before
imagen

After
imagen

imagen

imagen


oddMLan:
Cleaned the conflict by rebasing so there's no ugly "merged master into this branch"

oddMLan:
@cxd4 I agree that would be better, but not sure if these romhacks will be updated again. Then again, one could make the argument that PJ64 aims to be as accurate as possible and shouldn't cater to romhacks which don't even run on an actual console.

I just see two other things that ultimately inclined me to do this Pull Request:
- We have 32-bit mode and Fast SP which are already hacks, not to mention RDB hacks, patching annoyances such as Subscreen Delay in Zelda, Animal Crossing, etc. Which, needless to say, bypass accurate emulation
- People are staying with outdated versions of PJ64 because they can't play these romhacks anymore, missing many improvements from the newer versions. You could say that's on them, but I myself don't like switching between an older and a newer version, and rather have it all in one place.

I asked Kaze on Discord whether he intends to go back and update his romhacks to make them console compatible, and therefore, having the win-win situation you mentioned. 

He said that he's willing but just for the hacks that were meant to be within the N64 polygon limits. He also mentioned that there's a new tool for SM64 romhacks called SM64 Rom Manager which produces console-compatible code, so future hacks might not suffer from the same issues.

So ultimately this Pull Request might be pointless, I just need to hear from @project64 and what he has to say. If he isn't crazy about introducing yet another hack into PJ64, then I can close this PR.

LegendOfDragoon:
>Wouldn't it be more worthwhile to create a win-win situation by patching the ROM hack projects to not assume the preservation of the trailing alignment bit in the DMA address instead of patching Project64 to produce inaccurate results with inaccurate N64 code?

Agreed. I don't think we should go out of our way to support rom hacks that don't work on console.

theboy181:
I think having a mode for compatibility for ROM Hack is needed, but I also think that @kaze needs to work towards making things more in alignment with the HW. 

I get that using emulators to exceed the poly limits, is the perfect place for emulators.

gamemasterplc:
Exceeding polygon limits will probably just make the game unreasonably slow on console if properly implemented.

LegendOfDragoon:
Exceeding N64's limitations are is one thing, but bandaid patches for bugs is not good.

theboy181:
I can agree to that, but having a compatibility option is something that will stop users from using 1.6 (hopefully) While allowing people to enjoy, older hacks that will not see updates from their authors.

Maybe a disclaimer when you enable the option that stated the purpose of this option. 

" The following option is used to get around ROM hackers exploiting a bug in older versions of PJ64" After all, the HACK was bugged due to PJ64 1.6 allowing in in the first place.

oddMLan:
> bandaid patches for bugs is not good

*More* bandaid patches for bugs is not good, cuz we already have some of those in the project. I'd rather not have any of that in the first place. But Project64 is not aimed to be 1:1 a N64, but also providing added functionality & enhancements. I'd rather have more, than less, that's what I'm saying. And it's an optional feature that has to be manually enabled, so accurate emulation is the default behavior.

LegendOfDragoon:
If the dev is recommending 1.6, that likely means there's multiple issues and that they probably should be using Project64 1.6 for these rom hacks. My problem with this patch is that it enables complacency. With these workarounds, there's less less incentive to properly fix it.

theboy181:
Thats not it at all. That dev was working with bugged PJ64, and has changed his methods since. Its not meant to encourage the use of bad alignment, its meant to maintain compatibility for an old bug, while encouraging people to use latest PJ64 builds.

cxd4:
I'm not entirely sure what exceeding or not exceeding polygon limits has to do with keeping the PI DMA address 16-bit-aligned.  If other factors besides the RCP registers are causing hacks to work in emulators but not the actual hardware that's a separate issue altogether from creating a (potentially GUI-option-triggered) regression in how the RCP memory map bit fields are decoded.

>1.  We have 32-bit mode and Fast SP which are already hacks, not to mention RDB hacks, patching annoyances such as Subscreen Delay in Zelda, Animal Crossing, etc. Which, needless to say, bypass accurate emulation

32-bit mode has to be the stupidest thing literally still alive in Project64.  Not the coding of it, not the fact it's a hack...but because 64-bit systems have been universally widespread for over a decade now.  There's literally no reason to have a "assume 32-bit" option for optimization purposes, when Project64 ran faster in the 64-bit port I helped create anyway.  It's masochism.  zilmar should be encouraging other developers like @LegendOfDragoon to be adding a 64-bit dynarec to go along with the 64-bit interpreter if that's the only thing that's holding him back from moving on past the 32-bit world any sooner than he moved on from wanting to code for DOS.

Let's not forget the option to use a dynamic recompiler in the first place:  That is also an option present in Project64 which as you say "enables bypassing accurate emulation".  However, that and the 32-bit-mode setting are on a different scale from the hack option in this pull request, which outright offers an intentionally wrong decoding of one or more RCP registers.

>2.  People are staying with outdated versions of PJ64 because they can't play these romhacks anymore, missing many improvements from the newer versions. You could say that's on them, but I myself don't like switching between an older and a newer version, and rather have it all in one place.

They already had a choice...if Project64 1.0 through 1.6 had unaligned DMA addresses supported back then the users wouldn't have had to choose whether to bug the ROM hack developers or use an older version of each emulator, because the ROM hack developers wouldn't have had that escape window.  Users do have some influence over how the software they run is updated, but caving in on Project64's side instead of the buggy ROM hack's side flees from that social power.

cxd4:
By the way, why are several people in on this campaign to "stop" users from using 1.6 or older versions?  Not sure I see the point to fighting users using which version they prefer...I actually prefer 1.6 myself.  Though the PR being merged is a nice touch to include part of the old 1.6 back in anyway. :)

Fanatic-64:
Just an observation, but in order to make this useful for SM64 hacks out of the box, one of the following things need to be done:
1. Make Unaligned DMA on by default (this would probably give cxd4 an aneurysm).
2. Make Unaligned DMA on by default for ROMs not in the RDB (like the RDRAM Size setting).
3. Add all SM64 hacks to the RDB with this setting enabled.

oddMLan:
> 
> 
> Just an observation, but in order to make this useful for SM64 hacks out of the box, one of the following things need to be done:
> 
>     1. Make Unaligned DMA on by default (this would probably give cxd4 an aneurysm).
> 
>     2. Make Unaligned DMA on by default for ROMs not in the RDB (like the RDRAM Size setting).
> 
>     3. Add all SM64 hacks to the RDB with this setting enabled.

I don't see either of those options happening to be honest. But telling a user to tick "Unaligned DMA" under "Defaults" it's easier than making them download an outdated version of PJ64 and having to set it up

cxd4:
Precisely what oddMLan except you can also tick "Unaligned DMA" for per-ROM settings whenever you install a hack you know is not designed to be a genuine N64 executable or things like that.

theboy181:
got 404

Fanatic-64:
>got 404

Fixed link.

theboy181:
It might make more sense to have a setting for how much memory can be used, and then an enable disable mempak option.  This would be like having one installed or not installed.

Fanatic-64:
> 
> 
>It might make more sense to have a setting for how much memory can be used, and then an enable disable mempak option. This would be like having one installed or not installed.

Wait, are you talking about RDRAM size (which is what the commit I mentioned is about) or save chip size?

theboy181:
I’m talking about always 8mb on and a toggle to turn it off. 

In the game that gets destroyed with it on there would be an RDB override.  

I’m also talking about allowing the user to allocate more that 8mb if they choose as well. This would allow some of the modding community take advantage of the extra RDram.

PR #1547, opened 2019-01-13
Debugger ux improvements

theboy181:
This is going to make debugging less of a chore!  I hope that other elements like window position memory can be added soon! Thanks @oddMLan 

It would be nice if (debug ON) would remember what the last window positions were, and open them up. 

Might need a reset window position, to counter resolution, or secondary monitor/screen changes.


oddMLan:
lol the commits weren't ready for merge yet... writing the values manually in the cfg to retrieve position and size works.
But it needs to update those values when dragging or resizing the window... which I didn't implement successfully yet. I would appreciate some guidance with that, I'm not too familiar with WTL.

PR #1548, opened 2019-01-14
Use Interpreter's for x64.

Frank-74:
Recompiler doesn't work in x64.


oddMLan:
You should use this option instead.

imagen

Enable it by default for x64 builds.

Frank-74:
@oddMLan That only works for CPU not RSP.  Made the change for that.
if (g_Settings->LoadBool(Setting_ForceInterpreterCPU)) doesn't work from RSP Main.cpp.
Was easier to just force it on #ifndef _M_X64.

PR #1554, opened 2019-01-15
Misc bugfixes

oddMLan:


Source/Project64/UserInterface/MainMenu.cpp
```diff
@@ -449,17 +449,17 @@ bool CMainMenu::ProcessMessage(HWND hWnd, DWORD /*FromAccelerator*/, DWORD MenuI
         g_Settings->SaveBool(Debugger_ShowPifErrors, !g_Settings->LoadBool(Debugger_ShowPifErrors));
         break;
     case ID_DEBUG_SHOW_DLIST_COUNT:
-        g_Notify->DisplayMessage(0, "");
         g_Settings->SaveBool(Debugger_ShowDListAListCount, !g_Settings->LoadBool(Debugger_ShowDListAListCount));
+        g_Notify->DisplayMessage(0, "");
```
project64: what is the effect of moving the display message

oddMLan: Try toggling "DList/Alist" on the current master. You'll see that it won't properly clear the message when disabled. Moving the message fixes it. The order also matches with "Show CPU Usage".

PR #1556, opened 2019-01-16
[Debugger] Commands: Update address textbox when address is changed

shygoo:


theboy181:
Looking forward to this fix, does it also address the scrolling issues with when you double click on a instruction in CPU log?

PR #1564, opened 2019-01-19
[Debugger] CPU Log: More UI behavior improvements

shygoo:


theboy181:
Was this stuff we duscused earlier or something new?

PR #1565, opened 2019-01-20
[Debugger] Add GPR read/write breakpoints

shygoo:
screenshot

Click/right click labels to break on read/write to register


Source/Project64/UserInterface/resource.h
```diff
@@ -79,6 +79,7 @@
 #define IDD_Debugger_CPULog             204
 #define IDD_Enhancement_GS              205
 #define IDD_Debugger_ExceptionBP        207
+#define VERSION_BUILD                   835
```
project64: Version_build should not be added

PR #1572, opened 2019-01-23
Fix lockups caused by randomize dma?

Frank-74:
Fixes lockups in Conkers Bad Fur Day and random DMA.

### DIFF
--- a/Source/Project64-core/N64System/Mips/Dma.cpp
+++ b/Source/Project64-core/N64System/Mips/Dma.cpp
@@ -459,7 +459,7 @@ void CDMA::PI_DMA_WRITE()
         {
             g_Recompiler->ClearRecompCode_Phys(g_Reg->PI_DRAM_ADDR_REG, g_Reg->PI_WR_LEN_REG, CRecompiler::Remove_DMA);
         }
-        g_SystemTimer->SetTimer(g_SystemTimer->PiTimer, PI_WR_LEN_REG/8 + (g_Random->next() % 0x40), false);
+        g_SystemTimer->SetTimer(g_SystemTimer->PiTimer, PI_WR_LEN_REG/24 + (g_Random->next() % 0x40), false);
 
         //g_Reg->PI_STATUS_REG &= ~PI_STATUS_DMA_BUSY;
         //g_Reg->MI_INTR_REG |= MI_INTR_PI;


Frank-74:
Been debugging the PI_WR_LEN_REG, smallest value I've seen it is 16.  So probably shouldn't /24...

PI_WR_LEN_REG/16 + (g_Random->next() % 0x20 = no lock up in Conker.
PI_WR_LEN_REG/8 + (g_Random->next() % 0x10 = lock up.

Frank-74:
Grrr.

Conker lock up with /16 + rand->next() 0x20 here instead of intro logo:
gliden64_conker_bfd_000

Frank-74:
PI_WR_LEN_REG/16 + (g_Random->next() % 0x10 is the largest max size without any lock up.  I played up until here without lock up:
gliden64_conker_bfd_001

loganmc10:
mupen64plus ran into similar issues with Conkers when trying to randomize the PI DMAs, you can see the discussion here:

https://github.com/mupen64plus/mupen64plus-core/pull/415

We are doing:

PI_WR_LEN_REG/8 + Rand() %0x40

That seemed to keep Conkers stable

Frank-74:
@loganmc10 That's how it is already.  It will eventually lockup with default settings.  With overclock x2 it locks up sooner.  Been testing with overclock and turning off frame limiter.

/16 + rand() %16 is largest I can make it without an eventual lockup.

loganmc10:
Have you tried uncommenting this line?

https://github.com/Frank-74/project64/blob/master/Source/Project64-core/N64System/Mips/Dma.cpp#L55

mupen64plus sets that bit for reads and writes. PJ64 only does it for writes. I believe that could also affect games locking up

Frank-74:
Just tried it, locks up at N64 logo when booting.

Frank-74:
With FAT and /16 + rand() %16, only 2 characters ever come out of the box on SSB.  Pikachu and Samus.
So it's not random enough.  Without FAT they all appear eventually.

Frank-74:
@project64 How about just making PI random if Fixed Audio is ON?
```cpp
		if (g_Settings->LoadBool(Game_FixedAudio) == false)
        {
            g_Reg->PI_STATUS_REG &= ~PI_STATUS_DMA_BUSY;
            g_Reg->MI_INTR_REG |= MI_INTR_PI;
            g_Reg->CheckInterrupts();
        }
```
It's only needed if FAT is on anyway.

Azimer:
I think that's the easy way out Frank.  Ultimately if you want remote play FAT is a requirement as is this not-so-random randomization.  I think more research is required.

Frank-74:
I've managed to fix it I think.

Seeing as the lock up on Conker only occurs when Overclocking, I added this to Dma.cpp:
```cpp
        g_SystemTimer->SetTimer(g_SystemTimer->PiTimer, PI_WR_LEN_REG/8 + (g_Random->next() % 0x40), false);

		if (CGameSettings::OverClockModifier() > 1)
        {
            g_Reg->PI_STATUS_REG &= ~PI_STATUS_DMA_BUSY;
            g_Reg->MI_INTR_REG |= MI_INTR_PI;
            g_Reg->CheckInterrupts();
        }
```
And to fix randomization in Super Smash Bros, I added this to SystemTiming.cpp in UpdateTimers():
```cpp
random = m_Reg.RANDOM_REGISTER - (TimeTaken / g_System->CountPerOp() + (g_Random->next() % 0x20);
```
I set overclock to x 4 in SSB, so dma is done instant.  Everytime I reset SSB, a different character came out of the box.

Frank-74:
Just reverted the optimize update timers to how it was before roughly.  It's cleaner.
```cpp
void CSystemTimer::UpdateTimers()
{
    int TimeTaken = m_LastUpdate - m_NextTimer;
    if (TimeTaken != 0)
    {
        m_LastUpdate = m_NextTimer;
        m_Reg.COUNT_REGISTER += (TimeTaken / CGameSettings::OverClockModifier());
        m_Reg.RANDOM_REGISTER -= TimeTaken / g_System->CountPerOp() + (g_Random->next() % 0x20);
        while ((int)m_Reg.RANDOM_REGISTER < (int)m_Reg.WIRED_REGISTER)
        {
            m_Reg.RANDOM_REGISTER += 32 - m_Reg.WIRED_REGISTER;
        }
    }
}
```

gamemasterplc:
Instant DMA causes lockups on death in the OOT debug ROM too due to the game not handling the display list data being overwritten too quickly properly.

Frank-74:
> Instant DMA causes lockups on death in the OOT debug ROM too due to the game not handling the display list data being overwritten too quickly properly.

Is that the Master Quest version?  that's the only debug OOT ROM I have.

gamemasterplc:
Yes. Also happens on decompressed retail ROMs. Decompressing the ROMs makes it load files in 1 big uncompressed chunk instead of several smaller compressed chunks.

gamemasterplc:
It only affects OOT debug ROM and it may not happen with shitty graphics plugins. Try dying in hyrule field too. Some places do not trigger the crash because of different contents landing there.

Frank-74:
Just tested dying in OOT debug ROM at Hyrule Field.  No lock up.
My change only does instant dma when overclocking.  I overclocked x 4.  Maybe overclocking fixes the lock ups in OOT debug?

gamemasterplc:
Can you save and respawn elsewhere

oddMLan:
You're both on Discord64, why not communicating over there?

PR #1576, opened 2019-01-26
64DD support overhaul

LuigiBlood:
Currently overhauling 64DD support almost entirely.

### Features:
- 64DD Disk Images now appear in the ROM Browser, Recent Games as well.
- Seperate settings for each 64DD disk (RDB and User Config)
- Refactored game loading into core instead of UI.
- Unload Disk and 64DD IPL when playing a regular N64 game (frees memory when you don't need it, reminder that a 64DD Disk is almost 64MB)
  - This also fixes that exception error when it's about to save a disk file when it was not used.
- Forge Header of a 64DD game for plugin usage (or not? a discussion could be done about that)
  - Done but I'm not sure if I caused a bug with GUI Messages now or before then...
- More explicit error messages.
- Handle 64DD IPL loading errors better.
- **RDB to contain 64DD game configurations**
- Translation in French

### TODO:
- Proper Rom Information window for 64DD games?
- ***FIX BUGS***

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -7186,14 +7186,14 @@ Status=Compatible
 [E848D316-444B444A-C:4A]
 32bit=No
 Fixed Audio=0
-Good Name=Doshin no Kyojin 1 (J)
+Good Name=Kyojin no Doshin 1 (J)
 RDRAM Size=8
 Status=Compatible
 
 [E848D316-444B494A-C:4A]
 32bit=No
 Fixed Audio=0
-Good Name=Doshin no Kyojin 1 (J) (Store Demo)
+Good Name=Kyojin no Doshin 1 (J) (Store Demo)
 RDRAM Size=8
 Status=Compatible
 
@@ -7206,9 +7206,9 @@ Status=Only intro/part OK
 
 [E848D316-444B4B4A-C:4A]
 32bit=No
-Core Note=Expansion Disk for Doshin no Kyojin 1. Use the Swap feature.
+Core Note=Expansion Disk for Kyojin no Doshin 1. Use the Disk Swap feature to play.
 Fixed Audio=0
-Good Name=Doshin no Kyojin Kaihou Sensen Chibikkochikko Dai Shuugou (J)
+Good Name=Kyojin no Doshin - Kaihou Sensen Chibikkochikko Dai Shuugou (J)
 Status=Compatible
 
 [E848D316-45465A4A-C:4A]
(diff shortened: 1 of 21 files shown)


LuigiBlood:
To be honest I think the real meat of this overhaul is done, it's all polish at this point (still don't merge yet).
So that's nice :)

LuigiBlood:
image
No one should complain about the lack of notification anymore :)

LuigiBlood:
While I think this could be merged now, maybe some people might want to test it first, I also have to update the RDB as well for future work.

LuigiBlood:
image
This is good enough for me now.

oddMLan:
Super Mario 64 (Disk version) is missing in the Rdb. Other than that, excellent work! 😄

LuigiBlood:
I somehow forgot that. But then that means I have to do actual CRC because Mario 64 Disk does not have any Disk ID in it.

oddMLan:
imagen

.ndd roms aren't reading the "Fixed Audio=0" setting from their RDB entry.

imagen
The IPL rom reads it successfully however. Which makes me think something is wrong with the .ndd setting handling.

Also, if you right click a N64DD game from the ROM browser, then try to edit its settings, 1 out of 3 times it crashes the emulator.

LuigiBlood:
I thought there was an issue with something else, actually, but I guess not.

oddMLan:
Well it's not 1 out of 3 times after all. More like 1 out of 5 times. To reproduce the 'Edit Settings' crash more easily, close the emulator every time afterwards. Then open again, select game -> Edit settings. If it did not crash close and repeat, eventually it does crash.

>You sure you didn't disable Fixed Audio Timing in Defaults? Because honestly it does not read it when I do it on 64DD IPL either.

I didn't.

LuigiBlood:
Okay I get what you mean: I had already found this bug but I'm really not sure where the cause is, it's a hard to debug thing.

oddMLan:
imagen
SM64Disk showing these warnings, but otherwise game loads fine.

Frank-74:
Fixed Audio setting is read from Audio.rdb.  I've changed it to be read from Project64.rdb in this branch.
https://github.com/Frank-74/project64/tree/FAT-SUA-defaults

LuigiBlood:
...Okay I know why Mario 64 Disk is doing this I think: It has no Game ID or anything whatsoever. So it's a mess there. I think that's the plugin that's confused.

Frank-74:
Ok, I've done a small PR that moves Fixed Audio setting back to Project64.rdb.
https://github.com/project64/project64/pull/1580

PR #1585, opened 2019-01-27
[Project64-Audio] Remove FAT from audio plugin.

Frank-74:
Fixes main.cpp error.  Caused by FAT being read from Project64.rdb instead.


Frank-74:
Shall I close this PR and just comment out the offending line that reads FAT from Audio.rdb?

At moment without this PR, Project64 doesn't run because of audio plugin looking for FAT in Audio.rdb.
There's error in main.cpp on starting Project64.exe.

Maybe there should be extra error checking inside plugin, as error in main.cpp makes you think it's problem with project64.exe, not a plugin.

LuigiBlood:
This needs to be taken care of.

project64:
I am fine to merge it you want it in

oddMLan:
Yes, Project64 fails to start otherwise.

PR #1589, opened 2019-02-08
[RDB] Set 32bit=No and Counter Factor=1 for FIFA 99

Fanatic-64:
32-bit engine causes the geometry to be messed up and the game to quickly freeze when starting a match. Setting Counter Factor to 1 also makes the game a lot less choppy.


project64:
Can one of the admins verify this patch?

LegendOfDragoon:
A lot of sports games need this setting too. Some user reported issues with 32bit enabled for some NHL game iirc.

PR #1600, opened 2019-03-27
Fix undefined behavior related to sprintf

francois-berder:
sprintf source and destination should not overlap otherwise the behavior is undefined (as stated in C99 standard, section 7.19.6.6)


theboy181:
Thanks for the contributions! 

What is the best example of this fix?

PR #1607, opened 2019-04-15
Minor UI improvements

oddMLan:


oddMLan:
Do not merge still. Need to figure out a way so PJ64 scales up without issues with higher DPI displays.

oddMLan:
Ready for review/merging.

project64:
I had put a note down for my self to look at https://github.com/litehtml/litehtml to see if I could use that for the about/Support window to be able to make it a bit more dynamic .. but that is something different for the future

PR #1613, opened 2019-04-20
Restore Overclock modifier in Game Settings panel

oddMLan:
As requested by @theboy181 and @crashoveride

More convenient this way to test OC quiickly. Significantly less clicks.
Enhancement OC Modifier should still take priority over the OC Game Setting (tested and confirmed).
This way one can change OC from the Game Settings first and then optionally turn it into an Enhancement, if it proves effective.


theboy181:
Thanks @oddMLan !

If Zilmar doesn't want this enabled by default, maybe an option in the cfg file to enable it would make him accept the pull request?

project64:
I will put this in for now, not sure it will stay

theboy181:
Thank you!

PR #1617, opened 2019-04-25
[Debugger] Fix #1521 breakpoints bug

### LINKED ISSUE #1521: Breakpoints cause exception on DK64 if core mode is interpreter
Note: This only happens if the game config also has the CPU mode set to interpreter. If you just have "Always Use Interpreter Core" checked, this doesn't happen. **This is probably a separate bug but I'm not quite sure if I'm misunderstanding what that setting is for.**

Set "Always Use Interpreter Core"
Set the game-config to also use the interpreter
Launch DK64
Open debugger
Add a breakpoint to 803DE220
Exception is thrown

image

Latest dev build (after the previous fix). @shygoo

shygoo:


project64:
what is the bug that this is meant to fix?

shygoo:
Issue #1521 - setting a breakpoint causes an unknown memory action exception if the per-game core style is set to interpreter.

PostUpdateBP uses SysEvent_ResetRecompilerCode, and the handler for that event uses g_Recompiler, which is null if the per-game core style is set to interpreter.

project64:
https://github.com/project64/project64/commit/7a964b058e1f569e55c820caa0700d8abdc60298 I think fixes it

shygoo:
Looks good

PR #1620, opened 2019-04-29
Fixed some debugger script bugs

joeedh:
Fixed a few bugs in the debugger script system.  The most important was removing two calls that script threads were making to the debugger UI thread, which was causing major race conditions (m_Debugger->Debug_RefreshScriptsWindow() and m_ScriptSystem->DeleteStoppedInstances()).


project64:
I know you have done the same as the rest of the debugger code for critical sections, which is the right thing to do. The debugger it self should use CGuard instead of enter/leave, mostly cause if there is an exception then it will automatically close leave the critical section. Maybe when I have some time I can go though this and change it.

PR #1641, opened 2019-08-12
[64DD] Updated 64DD emulation

LuigiBlood:
Basically, the main changes are:
- Removes the dependence to MAME format internally (no more conversion)
  - Now supports SDK & D64 format natively via LBAToPhys & LBAToByte (rewritten libleo functions)
- CN64Disk handles Disk access calculations, now performs heuristics to find proper System Data and Disk ID in case of errors.
- Handles all 64DD IPLs seperately (JP disk with JP IPL, US disk with US IPL, Dev disk with Dev IPL).
- Performs custom CRC calculations for disks, to differenciate them properly.
  - RDB is updated as a result.
- Add "64DD" tab in Settings.
  - Handles IPL file paths and save type.
- Supports alternate save format that only saves RAM Area (smaller, now default).
  - Shadow files are still supported and taken as priority if Shadow File is found.
  - D64 Shadow File will have fully expanded RAM Area if smaller or not present.

**It's not very tested. I gotta say. I may push random bugfixes from time to time.**


project64:
is this fine to merge?

LuigiBlood:
huge bug found

LuigiBlood:
Question, when loading a ROM + disk combo (such as F-Zero X + Expansion Kit disk), do you want the settings from a base ROM to be loaded, or the settings from the disk to be loaded?

Fanatic-64:
Chiming in on the above, perhaps a setting could be included under a disk's entry to specify whether it's settings should override those of a base ROM? Something like OverrideBaseROMSettings=1/Yes or 0/No.

LuigiBlood:
For now the default is load from disk. I was suggested the idea of taking the base ROM settings and then have the settings of the Disk on top, but it's pretty difficult to implement properly.

LuigiBlood:
Putting aside the problem of configuration for combo ROM + Disk games, what it looks like to me is that it seems to work just fine for the most part.

LuigiBlood:
I think this is mergeable. I didn't find big issues when I messed around.

oddMLan:
@project64

PR #1642, opened 2019-08-17
Make code base compatible with Visual Studio 2019.

JunielKatarn:
- Allow building with the MSVC 14.2 toolset.
- Removed explicit reference to deprecated typedef std::map::_Pairib.

### DIFF
--- a/Source/Project64/UserInterface/Debugger/Breakpoints.cpp
+++ b/Source/Project64/UserInterface/Debugger/Breakpoints.cpp
@@ -64,7 +64,12 @@ bool CBreakpoints::WBPAdd(uint32_t address)
 bool CBreakpoints::AddExecution(uint32_t address, bool bTemporary)
 {
     PreUpdateBP();
+#if _MSC_VER >= 1920 // Visual Studio 2019 deprecates _Pairib
     auto res = m_Execution.insert(breakpoint_t::value_type(address, bTemporary));
+#else
+    breakpoints_t::_Pairib res = m_Execution.insert(breakpoint_t::value_type(address, bTemporary));
+#endif // _MSC_VER
+
     if (!res.second && !bTemporary)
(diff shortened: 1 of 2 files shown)


Source/Project64/UserInterface/Debugger/Breakpoints.cpp
```diff
@@ -64,7 +64,7 @@ bool CBreakpoints::WBPAdd(uint32_t address)
 bool CBreakpoints::AddExecution(uint32_t address, bool bTemporary)
 {
     PreUpdateBP();
-    breakpoints_t::_Pairib res = m_Execution.insert(breakpoint_t::value_type(address, bTemporary));
+    auto res = m_Execution.insert(breakpoint_t::value_type(address, bTemporary));
```
project64: this will likely break older versions of VS

JunielKatarn: The "official" building version is VS2015, right?

These C++2011 keywords are supported at least since VS2013. The change is safe :)

JunielKatarn: @project64 Regarding `_Pairib`, it should not be used directly, anyways.
Here is a reply by the main maintainer of the MSVS standard library:

https://twitter.com/StephanTLavavej/status/1162569765774168065

project64: it is used that way for android

JunielKatarn: OK, I'll if-def it to not affect versions prior to VS2029/VC14.2.

JunielKatarn: Updated.

PR #1644, opened 2019-08-21
[Disk] Add Disk Drive Settings file

LuigiBlood:
Forgot to add those, sorry about that.

### DIFF
--- a/Source/Project64/UserInterface/Settings/SettingsPage-DiskDrive.cpp
+++ b/Source/Project64/UserInterface/Settings/SettingsPage-DiskDrive.cpp
@@ -1,0 +1,162 @@
+/****************************************************************************
+*                                                                           *
+* Project64 - A Nintendo 64 emulator.                                      *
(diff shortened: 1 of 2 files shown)


project64:
you might have to re enable the code I disabled to get it to compile

LuigiBlood:
I've done it in ddrevamp_fix but I'll compile it first on my end, then do the PR

project64:
jni\Project64-core/N64System/N64DiskClass.h:129:40: warning: in-class initialization of non-static data member is a C++11 extension [-Wc++11-extensions]
    const uint16_t SCYL_ZONE_TBL[2][8] = {

LuigiBlood:
Man I tried to do const to optimize things but I didn't think C++ sucked that much...

That said I do not have that warning on my end?

LuigiBlood:
Should I add them to the constructor instead then? Or is there a better thing possible in C++? That really seems messy to me.

PR #1646, opened 2019-08-22
Add Combo Loading in Command Line & Main Menu

LuigiBlood:
Implemented N64 ROM + 64DD Disk Combo Loading via Command Line `--combo <Disk File>` and via Load Combo in the Main Menu (shortcut is Ctrl+Shift+O and is most likely breaking the *.sc3 file).

I see something about a `--help` argument in AppInit but is completely unused, how did you intend to use it @project64 ?


project64:
was planning on doing a lot more with command line options, and started to do help, but I never finished it

PR #1660, opened 2019-10-02
Implement Randomize SI/PI interrupts setting

### LINKED ISSUE #1656: Make SI/PI randomization a toggleable setting
SI/PI randomization was implemented a while back to make FAT random to fix #1290 & #1397. This is good to have as a default setting, however some users have reported that it breaks some enhancements (overclock, 60FPS) and specifically Conker's Bad Fur Day while using one of retroben's GS codes to boost the framerate.
Since Mupen64plus made it optional, I think it should be optional here as well.

mrexodia:
closes #1656

dialog

I tried to follow the existing code as best as possible. I only created a `Game_` setting, because according to #1656 it's a very fridge use case. Please let me know if you would like to see it implemented differently.

As for testing, I manually added a `__debugbreak()` in both code paths and checked if the setting was working as expected and everything seems fine.


mrexodia:
@oddMLan it looks like there is something misconfigured in the appveyor file, please take a look at it.

theboy181:
Great to see PR from new faces. :)

oddMLan:
@mrexodia looks like when you first pushed your commit, AppVeyor generated a commit specifically for the merge. https://github.com/project64/project64/commit/f8b18a81ecd6d421f35dd3462c5129705c21808a
This is when it compiled directly.
However, you didn't wait for the x64 build to build (about 2 minutes in addition to the 2 minutes it takes for the Win32 build to finish). How do I know this... well... the first version of your PR passed... but not as v2.4.0-1052-gf8b18a8. Check the history. It passed as the future merge the force push generated. Since the x64 build pre-FP was yet to be built, it queried the repository, which generated a new number. It updated what should've been v2.4.0-1052-gf8b18a8 to v2.3.2-1052-*g34f0172a*. UpdateVersion updated the name accordingly for that job. Then it tried to build the actual v2.3.2-1052-g34f0172a, it stopped and said it "already existed". Your additional commit did nothing but generate yet again another commit for the merge, avoiding the duplication in the job name, so it succeeded.

mrexodia:
Hm that makes sense. I changed the fix to now suffix the AppVeyor version with `build-%APPVEYOR_BUILD_NUMBER%`, which should fix the issue for good. Should I do that in a separate PR?

oddMLan:
That's one way to solve it.
I'll submit a PR with the other approach I had in mind

oddMLan:
@MrExodia would you like to join our N64 emulation server? Invite link: https://discord.gg/7n2Ks3

Lots of development talk is happening there right now. We would love to have you there 😁 All the most well-known N64 emudevs are a part of this server as well

oddMLan:
Also I would suggest adding this new setting to the "Defaults" page as well, that way it can be enabled/disabled globally if desired.

mrexodia:
Alright, I’ll add the default button later today. How do I open this menu?

oddMLan:
Perfect!

PR #1662, opened 2019-10-03
[appveyor] Fix package zip filename

oddMLan:
instead of using `current_tag` for the zip, use `APPVEYOR_BUILD_VERSION`. this fixes an empty zip filename


oddMLan:
wait a sec, dont merge yet.

oddMLan:
Just checked and now everything's working as intended.

oddMLan:
Hold on, I just noticed the zips are generating a v2.3.2-xxxx filename
meanwhile https://www.pj64-emu.com/nightly-builds uses v2.4.0-xxxx
which is incorrect because we're using `git describe --tags --long --dirty` to generate a filename
which means that the official buildbot hardcoded the v2.4.0 suffix.

Should this discrepancy be addressed, by either making a v2.4.0 tag in this repo (recommended) or by hardcoding the v2.4.0 part in the appveyor.yml?

oddMLan:
Just checked and now everything's working as intended.

PR #1665, opened 2019-10-23
Discord RPC

CrankySupertoon:
This is a modification of some.of @shygoo's Work. Everything is not loaded from the core from the DiscorrRPC class. It checks if the RDB matches this game and ifit does display it. Ifit doesn't it passes until it gets to the bottom. Client id is in DiscordRPC near the first few lines. I would recommend zilmar to add his own then a stranger like me


oddMLan:
Looks great!

Source/Project64/UserInterface/DiscordRPC.cpp
```diff
@@ -0,0 +1,332 @@
...
+		char szState[256];
+		sprintf(szState, "Playing Mario Party 2");
+		discordPresence.state = szState;
+		discordPresence.startTimestamp = Timestamp();
+		discordPresence.largeImageKey = "mp2";
+		discordPresence.largeImageText = "Mario Party 2";
+		discordPresence.smallImageKey = "icon";
+		discordPresence.smallImageText = "Project64";
+		discordPresence.instance = 1;
+	}
```
shygoo: To cut down on repeated code, I suggest adding the `largeImageKey` strings to the RDB file and use Rdb_GoodName for `largeImageText` and `state`.

CrankySupertoon: Funny enough I have the RDB good name like that originally for the text and the state: how do I add the strings to the RDB file. Assuming somewhat with. https://github.com/project64/project64/blob/4e2c5949dd6a38eb76c56a09807e8925b604b97b/Source/Project64-core/Settings/SettingsID Also since the key is just the file extension in discord, should I add them to discord via just the RDB name like fully writing out Super Mario 64, or what do you prefer

CrankySupertoon:
Im picking this up now

CrankySupertoon:
Compared to the last PR the largeImageText and state keys are using the RDB_GoodName. I tried using the ROM Header for the key to upload but the file cant have spaces. It worked for some games that headers didnt have spaces. I need help finding how to add a new line to the Database file and also i cleaned up the code a lot.

CrankySupertoon:
I also need @project64 to make a app on developers.discordapp.com and upload the pictures for the game box art.

CrankySupertoon:
This is basically done. @project64 has to go onto the discord developers site and make a PJ64 app and add Rich Presence to it. The Client ID Gets put here https://github.com/Project64Netplay/project64/blob/801aa13623f033452c10b52f095599a86ca696b3/Source/Project64/UserInterface/DiscordRPC.cpp#L20. You can add images for the games in the app. For example Super Mario 64 should be **sm64-u** or something like that. You have to add that name in the RDB here https://github.com/project64/project64/blob/68e59320d94f2e4eb3e8c196c436c0e3149aa1a4/Config/Project64.rdb#L3358

oddMLan:
Code look much cleaner now. LGTM

PR #1674, opened 2019-11-27
Turkish language file character encoding changed to UTF-8 BOM.

gurkangullu:


oddMLan:
Looks good to me. All the other lang files are encoded as UTF-8 BOM

PR #1677, opened 2019-11-30
[Debugger] Rewrite memory viewer

shygoo:
Changes:
 * replace the list control with a custom hex editor control
 * add ability to select/copy/paste/delete multiple bytes
 * add byte group size option
 * make window resizable
 * fix DPI-related issues
 * speed up auto-refresh
 * remove option to disable auto-refresh (new control's cpu usage is insignificant)
 * move the dump and search buttons into the context menu
 * change colors of breakpoints to match the command window's
 * highlight CPU read and write targets while stepping
 * move address info into a status bar
 * allow window to open before MMU is initialized
 * add shortcut keys for various actions
 * add "follow pointer" feature
 * add "safe mode" feature
 * add copy gameshark code feature
 * add copy with row/group addresses feature
 * add tabs feature
 * add jump menu for hardware regions
 * fix cartridge ROM addresses not being viewable
 * allow writes to cartridge ROM
 * use thread-safe functions for reading/writing memory

Manual:
 * https://hack64.net/docs/pj64-memview-reference.htm


theboy181:
you are the MAN!!

oddMLan:
This is down to preference but instead of Consolas for the font I would suggest the same one the Breakpoints window has for the registers, Lucida Console if I'm not mistaken. The slashed zeroes makes the information in the window look a bit busy IMO. Having the slashed zero is only crucial if you have information other than numbers (like the letter 'O')

PR #1684, opened 2019-12-16
Trap Exceptions

krimtonz:
Add remaining trap exceptions to the interpreter, proper handling of trap exceptions.


oddMLan:
Awesome! Out of curiosity, in what scenarios will this help?

krimtonz:
The Ocarina of Time community is planning on attempting to use traps for Arbitrary Code Execution.
Also they should be included for completion anyways.

MrCheeze:
To be exact, there is an ACE-related blueprint that unavoidably passes through some "trap if condition" instructions as a side effect, where the condition in question is false.

PR #1686, opened 2019-12-17
Cartdom2addr2

krimtonz:
Properly emulate the way n64 handles Cartridge Domain 2 Address 2 reads/writes when they're outside of savefile scope.  Writes are quietly ignored while reads generate random noise.


krimtonz:
This is primarily being added for when romhacks use things such as the everdrive 64 registers at 0xA8040000, it does not generate an exception as it shouldn't.

Source/Project64-core/N64System/Mips/MemoryVirtualMem.cpp
```diff
@@ -1507,6 +1507,12 @@ void CMipsMemoryVM::Load32CartridgeDomain2Address1(void)
 
 void CMipsMemoryVM::Load32CartridgeDomain2Address2(void)
 {
+    uint32_t PAddr = m_MemLookupAddress & 0x1FFFFFFF;
+    if (PAddr >= 0x10000)
+    {
+        m_MemLookupValue.UW[0] = rand();
+        return;
+    }
```
shygoo: PAddr will always be greater than 0x10000 here, I think you meant:
```
uint32_t saveOffset = (m_MemLookupAddress & 0x1FFFFFFF) - 0x08000000;
```

Also this technically depends on the cartridge, but the value you get from an open bus read shouldn't be a random number. On official cartridges it's something like this:
```
m_MemLookupValue.UW[0] = m_MemLookupAddress & 0xFFFF;
m_MemLookupValue.UW[0] = (m_MemLookupValue.UW[0] << 16) | m_MemLookupValue.UW[0];
```

krimtonz: Thanks for catching that!  And the open bus read makes sense.

PR #1692, opened 2019-12-29
Real Subscreen Delay Fix

AriaHiro64:
This fixes the subscreen delay without gameshark codes. This doesn't fix crashes on uncompressed roms but that happens regardless.


oddMLan:
Compared against angrylion and against console footage. Subscreen menu now opens at the same time. Nice catch!

AriaHiro64:
Added crash fixes for commonly used base roms, Specifically 1.0 and Debug when uncompressed.

Config/Project64.rdb
```diff
@@ -5825,8 +5825,8 @@ Self Texture=1
 Good Name=The Legend of Zelda - Ocarina of Time (E) (GC)
 Internal Name=THE LEGEND OF ZELDA
 Status=Compatible
-Cheat0=801D8F8B 0002 //Subscreen Delay Fix
-Cheat1=D109A8E4 0320,8109A8E4 0000,D109A8E6 F809,8109A8E6 0000 //End Credits Fix
+Counter Factor=1 //Actual Subscreen Delay Fix
```
project64: setting cf to 1 is not a real fix, it is to do with clearing the background in the gfx plugin, as jabo's does not modify the rdram this does not fix things. So a true fix is in the gfx plugin

oddMLan: It does help with Glide64 and GLideN64. CF=1 matches console (subscreen appears right before jingle finishes playing). I verified this with console footage. CF = 2, subscreen appears moments _after_ jingle has played out.

So CF=1 does help with accurate gfx plugins. Do you just want @AriaHiro64 to leave the Cheat0 intact for Jabo (which is not even the default anymore and Project64 Video seems to work fine without the fix). The fix is not even getting applied to Project64 Video because the plugin is not called "Glide64 for Project64" as it is hardcoded in the RDB for that fix. So it only helps Jabo.

AriaHiro64: Jabo is literally the only plugin that's unaffected

AriaHiro64: Jabo has its own bug and shouldn't be a basis on every other plugin

AriaHiro64: I've tested rice angrylion gliden64 and glide and all benefit

AriaHiro64: Yes jabo needs improvement and I'm not trying shit on jabo when I say this but you would be hard-pressed to find a worse option for oot in it's current state

AriaHiro64:
I added back the jabo related fixes and added new crcs for decompressed roms 1.0 u an j

project64:
this might be fine, but I am not sure and closing old pull requests, so closing now

PR #1705, opened 2020-02-11
 Added type, name, value, desc editing in the symbols window; 'new tab' shortcut to memory viewer

Drahsid:
I'm not sure if this is written how it should be, but the lack of this feature had been bugging me for a bit and I would like to see it to be standard.

Edit: I added a 'new tab' shortcut in the memory viewer with ctrl+n. This is different from the duplicate tab feature since it opens a tab at the start of rdram. This may be preferred in some cases over ctrl+t, or for some who don't know all of the features, they might find this to be more intuitive.

### DIFF
--- a/Source/Project64/UserInterface/Debugger/Debugger-Symbols.cpp
+++ b/Source/Project64/UserInterface/Debugger/Debugger-Symbols.cpp
@@ -17,6 +17,43 @@
 
 #include "Symbols.h"
 
+const CSetValueDlg::ComboItem CDebugSymbols::ModalChangeTypeItems[] = {
+	{ "code",   SYM_CODE},
+	{ "uint8",  SYM_U8 },
+	{ "int8",   SYM_S8 },
+	{ "uint16", SYM_U16 },
+	{ "int16",  SYM_S16 },
+	{ "uint32", SYM_U32 },
+	{ "int32",  SYM_S32 },
+	{ "uint64", SYM_U64 },
+	{ "int64",  SYM_S64 },
+	{ "float",  SYM_FLOAT },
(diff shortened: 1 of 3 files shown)


Source/Project64/UserInterface/Debugger/Debugger-Symbols.cpp
```diff
@@ -17,6 +17,43 @@
...
+	case SYM_S16:  return "int16";
+	case SYM_S32:  return "int32";
+	case SYM_S64: return "int64";
+	case SYM_FLOAT: return "float";
+	case SYM_DOUBLE: return "double";
+	case SYM_INVALID: return "invalid";
+	}
+
+	return NULL;
+}
```
shygoo: There is already a `CSymbolTable::GetTypeName(int typeId)` function for this

Source/Project64/UserInterface/Debugger/Debugger-Symbols.h
```diff
@@ -20,7 +20,22 @@ class CDebugSymbols :
...
 {
 private:
+	enum {
+		m_SymbolsListView_Col_Address,
+		m_SymbolsListView_Col_Type,
+		m_SymbolsListView_Col_Name,
+		m_SymbolsListView_Col_Value,
+		m_SymbolsListView_Col_Description,
+		m_SymbolsListView_Num_Columns
+	};
```
shygoo: The `m_` prefix is not needed here. Also please use spaces instead of tabs. Everything looks good otherwise 👍

oddMLan:
Is this done?

Drahsid:
If there are no issues with this pr in its current state, I do not intend to update it.

PR #1736, opened 2020-05-05
Discord-rpc: Code cleanup

oddMLan:
This is inspired by the Dolphin-emu approach

The internal name is not as informative so now it'll try to use GoodName first then FileName (same as RomBrowser)
Checked and it also works good with zip/7z files


oddMLan:
Merge #1735 first

PR #1769, opened 2020-05-28
[RDB] Add Zelda 64 Dawn & Dusk & English F-Zero X Expansion Disks

LuigiBlood:


oddMLan:
Why not naming the expansion disks Dawn & Dusk?

LuigiBlood:
In case someone else makes another expansion with the same codebase, since it's open source

PR #1789, opened 2020-06-09
[Save State] Add Disk Interface to Save State file

LuigiBlood:
I noticed 64DD games don't really play well with save states. I noticed that I forgot to add the registers to the save states.

I don't know if you're satisfied with this, as this breaks the Save State format, maybe you'd prefer putting the Disk Registers at the end of the file?
@project64


LuigiBlood:
Actually I might need to do a bit more for the save states.

LuigiBlood:
Just saying here for sake of documentation:
- We need to look up how to not break save state backward compatibility.
- Save states broke between 2.2, 2.3 (64DD support added) & 2.4 (Disk Seek support)

LuigiBlood:
I would like input on the code, I may want to add a way to make sure it recognizes combo loading as well by putting 0x20 bytes of the base N64 ROM and 0x20 bytes of the Disk ID (which is the entirety of it).

PR #1809, opened 2020-06-15
Added Angrylion RDP Plus to graphics card section in README

RokkumanX:


project64:
Angrylion RDP Plus is not relevant to the read me since it is not included with pj64

RokkumanX:
It's a shame man, maybe you should re-think and start include Angrylion-RDP-Plus because it's one of the best and most accurate N64 GFX plugins to date and the performance is great even.

Runs full-speed on my system with cxd4 RSP.

project64:
that might be true and the emulator is designed to allow this, but it will not likely to be included in the distribution of Project64. People are free to add if they want to.

PR #1844, opened 2020-08-07
Discord link added for Notepad users

Alen-Alic:


oddMLan:
This is not necessary, you can already see the link in parentheses.

PR #1862, opened 2020-10-26
[Debugger] Add missing cop1 cvt ops to assembler, remove constexprs

shygoo:


JunielKatarn:
Wat's wrong with `constexpr`?

shygoo:
Nothing really. Just wanted to make the code more consistent with the rest of the project since it doesn't use C++11 anywhere else as far as I know. I recall someone in the n64 emulation discord having trouble building with an older version of VS because of this file in particular.

JunielKatarn:
> Nothing really. Just wanted to make the code more consistent with the rest of the project since it doesn't use C++11 anywhere else as far as I know. I recall someone in the n64 emulation discord having trouble building with an older version of VS because of this file in particular.

Makes sense.
IMHO, the source base should be pushed to be C++17-compatible.
I don't see any benefit of using older and buggier MSVC compilers.

project64:
> IMHO, the source base should be pushed to be C++17-compatible.

what is the benefit, does it make the code more consistent, easier to understand, faster?

or is it using the latest for the sake of using the latest?

JunielKatarn:
> what is the benefit, does it make the code more consistent, easier to understand, faster?
> 
> or is it using the latest for the sake of using the latest?

It depends on specific features, but `constexpr`, for instance, is *not* syntactic sugar.
It solves problems related to constant values in memory.

The bulk of the benefits in recent C++ standards relies not only in language features, but vastly improved APIs from the Standard Library.

Ultimately, it's not about the language standard.
Recent versions of MSVC do have documented performance improvements. This is also true for GCC and Clang.

I'd pose the counter-question:
What is the benefit of sticking with older, buggier and barely supported compilers, other than clinging to Windows XP?

mlopezrnet:
> What is the benefit of sticking with older, buggier and barely supported compilers, other than clinging to Windows XP?

In my personal experience, while the newer compilers do apply more optimizations they tend to be buggier too. GLideN64 had a severe gfx bug in the 2019 compiler but not 2017, which only was able to be solved by disabling Link Time Optimization.

oddMLan:
> What is the benefit of sticking with older, buggier and barely supported compilers, other than clinging to Windows XP?

In my personal experience, while the newer compilers apply more optimization this tends to create more bugs too. GLideN64 had a severe blending issue that only happened with MSVC 2019 but not 2017. Only by disabling Link Time Optimization the bug was avoided.

JunielKatarn:
> In my personal experience, while the newer compilers apply more optimization this tends to create more bugs too. GLideN64 had a severe blending issue that only happened with MSVC 2019 but not 2017. Only by disabling Link Time Optimization the bug was avoided.

Interesting. My experience had been quite the opposite (granted, with other projects like FFmpeg).
Would you please share more about the issue you found with GLideN64? I'd like to reproduce it and tinker with it.

oddMLan:
> @oddMLan Interesting. My experience had been quite the opposite (granted, with other projects like FFmpeg).
> Would you please share more about the issue you found with GLideN64? I'd like to reproduce it and tinker with it.

https://github.com/gonetz/GLideN64/commit/e1acebd84acfe1d7bdc905ecf10051af0c0f23a8

PR #1881, opened 2020-12-08
Update About.cpp

Alen-Alic:
(grammar fixes and more 'thanks' credits added in about window)


CrashOveride95:
yes very useful, pj64 will never be the same

Alen-Alic:
> yes very useful, pj64 will never be the same

Every little bit helps my sarcastic friend.

CrashOveride95:
> > yes very useful, pj64 will never be the same
> 
> Every little bit helps my sarcastic friend.

Alright, so you think you can add yourself to the list of people who have made actual contributions to the codebase because you fixed *A grammatical error*
OK buddy

CrashOveride95:
Also, you did the same thing almost a month ago #1870 but closed it shortly before making this new one with the exact same useless changes

Is this some new form of attention-seeking I am unaware of?

PR #1896, opened 2021-01-11
Don't merge yet, im just totally making use of the functional appveyor on pull requests

Squall-Leonhart:


Squall-Leonhart:
won't work, the 64DD disk header isn't really implemented at all.

Issue #1835, opened 2020-07-12
Error when manually starting emulation after loading ROM


Fanatic-64:
On the latest build. if the option "Start emulation when ROM is opened" is disabled, a ROM is opened, and then File>Start Emulation is selected or F11 is pressed, emulation will fail to start with the following message:
>Break point found at
UserInterface\MainMenu.cpp
300

This also happens on release version 2.3.2.202 (albeit with a Break point found error pointing to line 262 of MainMenuClass.cpp), but not on 2.3.0.210 or earlier.

vgturtle127:
Is this fixed with the merged PR?

Fanatic-64:
Fixed.

PR #1910, opened 2021-02-02
Fix Start Emulation by only skipping StartEmulation when AutoStart is disabled

flagrama:
RunLoadedImage is skipped in the Run*Image methods in CN64System if AutoStart is disabled. If the check for if there is a current g_BaseSystem fails, these changes run RunLoadedImage to create one and StartEmulation.

I'm not sure the extra NULL guard is necessary, but I don't want to skip the first g_BaseSystem check altogether. I added it as a "just in case."

Related: #1835


flagrama:
Taking another look at it today, an alternative could be to always call `RunLoadedImage` and do the Setting_AutoStart check in there to decide to StartEmulation. That's probably a cleaner solution.

PR #1994, opened 2021-02-28
Update N64System.h copyright header and comments

vgturtle127:
Switched from block comments to single line comments, updated copyright year, reduced verbosity of license lines, and switched to HTTPS for license link.

This copyright header will also serve as the new copyright header for the rest of the files in the repo, each file will be a separate pull request so changes can be reviewed on a per file basis.

### DIFF
--- a/Source/Project64/N64System.h
+++ b/Source/Project64/N64System.h
@@ -1,13 +1,8 @@
-/****************************************************************************
-*                                                                           *
-* Project64 - A Nintendo 64 emulator.                                      *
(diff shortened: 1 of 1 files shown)


vgturtle127:
Let me know if I should do comments in a separate PR from copyright headers from now on.

flagrama:
It's recommended to display the year of first publication of the work. There is no reason to update it because only the publication date matters, but if you do, it should be in addition to it, not replacing it.

vgturtle127:
I went ahead and corrected it and then did both

PR #1995, opened 2021-02-28
Linux fixes

flagrama:
Trying to fix the build issues with the Unix build script. The issues solved so far:
* Backslashes in include directives are incompatible with `g++`. 
  * They have all been replaced with slashes.
* Build script is out-of-date.
  * New files have been added, old files have been removed.
  * I am unsure if the build order is correct or if I've added all of the relevant files.
    * I believe I've removed the files that need to be removed however, as I am no longer getting missing file errors from the compiler.
* Platform-dependent function definitions are being used without being properly defined for unix-based platforms.
  * This is handled in most places already, but some new ones seems to have appeared.
  * I've included `stdafx.h` or `Common/stdafx.h` to correct those new errors.

Still remaining are the following:
<details><summary>Possible mix up between SearchPath and SearchDir?</summary>

```
In file included from ./../../Project64-core/RomList/../stdafx.h:6,
                 from ./../../Project64-core/RomList/stdafx.h:1,
                 from ./../../Project64-core/RomList/RomList.cpp:11:
./../../Project64-core/RomList/RomList.cpp: In member function ‘void CRomList::FillRomList(strlist&, const char*)’:
./../../Project64-core/RomList/RomList.cpp:158:76: error: ‘SearchPath’ was not declared in this scope
  158 |     WriteTrace(TraceRomList, TraceVerbose, "SearchPath: %s", (const char *)SearchPath);
      |                                                                            ^~~~~~~~~~

./../../Project64-core/RomList/RomList.cpp:168:98: error: ‘SearchPath’ was not declared in this scope
  168 |         WriteTrace(TraceRomList, TraceVerbose, "Found: \"%s\" m_StopRefresh = %s", (const char *)SearchPath, m_StopRefresh ? "true" : "false");
```

This appears to be a Windows SDK method that is not defined for unix-based platforms. A cross-platform solution or a unix-based platform definition should probably be added to imitate it.
</details>

* Corrected variable being used.

<details><summary>Strict Narrowing Conversion Warning Being an Error</summary>

```
./../../Project64-core/Settings/SettingType/SettingsType-RDBSaveChip.cpp: In member function ‘virtual void CSettingTypeRDBSaveChip::Save(uint32_t, uint32_t)’:
./../../Project64-core/Settings/SettingType/SettingsType-RDBSaveChip.cpp:103:10: error: narrowing conversion of ‘SaveChip_Auto’ from ‘int’ to ‘unsigned int’ [-Wnarrowing]
  103 |     case SaveChip_Auto:       m_SettingsIniFile->SaveString(m_SectionIdent->c_str(),m_KeyName.c_str(),"First Save Type"); break;
      |          ^~~~~~~~~~~~~
```

g++ really doesn't like the "-1" value for `SaveChip_Auto` in it's enum. This can be ignored by adding the `-Wno-narrowing` compiler flag, but seems extremely overkill to me.

</details>

* Ignore the error in that area (for now)

### DIFF
--- a/Source/Common/IniFileClass.cpp
+++ b/Source/Common/IniFileClass.cpp
@@ -1,4 +1,3 @@
-#include "stdafx.h"
 #include <stdlib.h>
 #include <stdarg.h>
 
--- a/Source/Common/IniFileClass.cpp
+++ b/Source/Common/IniFileClass.cpp
@@ -1,3 +1,4 @@
+#include "stdafx.h"
 #include <stdlib.h>
 #include <stdarg.h>
 
--- a/Source/Common/IniFileClass.h
+++ b/Source/Common/IniFileClass.h
@@ -4,6 +4,7 @@
 #include <strings.h>
 #endif
 
+#include "stdafx.h"
 #include "FileClass.h"
 #include "CriticalSection.h"
 #include <string>
--- a/Source/Common/Platform.h
+++ b/Source/Common/Platform.h
@@ -4,6 +4,7 @@
 #include <alloca.h>
 #include <stdarg.h>
 
+#define stricmp strcasecmp
 #define _stricmp strcasecmp
 #define _strnicmp strncasecmp
 #define _snprintf snprintf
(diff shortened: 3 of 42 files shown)


cxd4:
Props for maintaining those.

PR #1999, opened 2021-03-03
Code and comment beautification (starting with main.cpp)

vgturtle127:
Eliminate title case in some instances, improve some wording, and add a space after the single line comment block to improve readability on some editors and systems.

Also did this to the trace logging as well.

### DIFF
--- a/Source/Project64/main.cpp
+++ b/Source/Project64/main.cpp
@@ -14,8 +14,9 @@ int WINAPI WinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, LPSTR /
             WelcomeScreen().DoModal();
         }
 
-        //Create the main window with Menu
-        WriteTrace(TraceUserInterface, TraceDebug, "Create Main Window");
+        // Create the main window with menu
+		
+        WriteTrace(TraceUserInterface, TraceDebug, "Create main window");
         CMainGui MainWindow(true, stdstr_f("Project64 %s", VER_FILE_VERSION_STR).c_str()), HiddenWindow(false);
         CMainMenu MainMenu(&MainWindow);
         CDebuggerUI Debugger;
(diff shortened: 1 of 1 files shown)


Source/Project64/main.cpp
```diff
@@ -100,4 +101,4 @@ int WINAPI WinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, LPSTR /
     AppCleanup();
     CoUninitialize();
     return true;
-}
\ No newline at end of file
+}
```
flagrama, line 104: While we are here, can we make sure the files have an empty new line at the end so every change to the file doesn't have this "change"?

flagrama:
If we're doing some cleanup stuff, perhaps we could implement an [EditorConfig](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019)? It unfortunately requires [an add-on in Visual Studio 2015](https://marketplace.visualstudio.com/items?itemName=EditorConfigTeam.EditorConfig) but VS2017 and 2019 should support it natively. It should help ensure some sort of consistency for those who it works for automatically, those who choose to install the extension, and for cleanup projects like these in the future.

vgturtle127:
I definitely agree about the EditorConfig, yeah. I haven't looked into how that works yet, but I want to implement it into this project and my own at some point.

As for the empty new line thing, I was told by Zilmar to only do one file per PR to make it easier to see what's being changed and what isn't, but if he says it's OK to do all the new lines in this PR I'd be OK with that.

flagrama:
I didn't necessarily mean in this PR. Just as you're going through them. I pointed it out only because this one didn't have one showing that "fake change" git always shows in that instance.

vgturtle127:
Ohhhhh, I see. Got ya, I will add that then. 😀

vgturtle127:
@project64 Lemme know if you have any suggestions about this PR

PR #2006, opened 2021-03-11
Add support for ARM64

JunielKatarn:
Allows building for the Windows ARM64 platform.

~~- Video and Input plugins not supported yet.~~
- Everything builds as with `x64`. Ready for testing.
- The app actually boots.


Project64.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio Version 16
```
JunielKatarn, line 2: Note, this change is introduced by the IDE and is inconsequential.
The solution will work the same on any supported VS version.

JunielKatarn:
Hey @project64, this PR did not show up in the Discord channel for some reason.
Care to take a look?
It increases platform compatibility.

flagrama:
It's in the channel https://discord.com/channels/542911640127930381/770544305042751528/819361750066397225 just buried under it's own synchronize events that Yappy doesn't support yet for some reason.

JunielKatarn:
> It's in the channel https://discord.com/channels/542911640127930381/770544305042751528/819361750066397225 just buried under it's own synchronize events that Yappy doesn't support yet for some reason.

True! Somehow I missed it. I'll edit my comment.

project64:
I am just being cautious with merging any project settings before I finalize 3.0, which is why I am closing with out merging.

I am also not sure when pj64 would need to be built for arm64

JunielKatarn:
> I am just being cautious with merging any project settings before I finalize 3.0, which is why I am closing with out merging.
> 
> I am also not sure when pj64 would need to be built for arm64

@project64  Because Windows on ARM64 is a reality and there is increasing adoption of devices using Windows/ARM64.
Same as with x64, it's best to have a fully native experience instead of just relying on emulators.

Can we revisit this PR after 3.0 is finalized?
Do you have an estimate target date?

project64:
Talk to me it on discord and I can go through my plans

PR #2007, opened 2021-03-11
Add final new lines to 2 files and beautify 2 files in the settings d…

vgturtle127:
…irectory


vgturtle127:
TL;DR this is the first PR in a series of PR's to beautify the code and comments

oddMLan:
I don't see why is necessary adding a space after the // comment . it looks wrong imo

vgturtle127:
The only time I do it for comments which aren't commenting out code, in which case it increases readability and is more proper

PR #2008, opened 2021-03-11
Beautification part 2

vgturtle127:
More things! I'm trying to hit things as I see them, so I can tackle the big directories later.


JunielKatarn:
At this point, it could be worth it to consider clang format.

That thing can keep the sources formatted automatically and abide by whatever convention is chosen.

PR #2012, opened 2021-03-12
Chose correct default Windows SDK

JunielKatarn:
When using an older MSVC compiler toolset than the default for the Visual Studio version, build fails to select or find the correct Windows SDK installation.

Example:
Visual Studio 2019 can install the VS 2015 (MSVC  `v140`) compilers, but that compiler version assumes the installation of Windows SDK version 8.1.

## Proposed Changes

  - Explicitly set the minimum target Windows SDK version (8.1)
  - Allow MSBuild to select the  most likely Windows SDK installation, based on the Visual Studio version:
    - When using MSVC `v142`, using `10.0` will select the highest available `10.0.*` instance.
    - When using VS2019 but an older MSVC (i.e. `v140`), which can not use the `10.0` metavalue, use the default SDK version provided by Visual Studio as of the current release (`10.0.19041.0`).
    - Any other scenario will default the SDK version to `8.1`, which is available in VS 2017 and VS 2015.

### DIFF
--- a/PropertySheets/Platform.props
+++ b/PropertySheets/Platform.props
@@ -2,8 +2,8 @@
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup Label="Globals">
-    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)' == '' AND $(PlatformToolsetVersion) >= 142">10.0</WindowsTargetPlatformVersion>
-    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)' == '' AND '$(VisualStudioVersion)' >= 16.0">10.0.19041.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)' == '' AND '$(PlatformToolsetVersion)' != '' AND $(PlatformToolsetVersion) >= 142">10.0</WindowsTargetPlatformVersion>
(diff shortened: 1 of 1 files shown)


PropertySheets/Platform.props
```diff
@@ -1,228 +1,241 @@
...
-</Project>
\ No newline at end of file
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup Label="Globals">
+    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)' == '' AND '$(PlatformToolset)' == 'v142'">10.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion Condition="'$(WindowsTargetPlatformVersion)' == '' AND '$(VisualStudioVersion)' >= 16.0">10.0.19041.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformMinVersion>8.1</WindowsTargetPlatformMinVersion>
+  </PropertyGroup>
```
Jj0YzL5nvJ, line 8: VS2015 can also use SDK 10.0, but it is not mandatory to set up an SDK to start with. It's only mandatory to establish the SDK from VS2017 onwards.

These implementations are more robust:
https://github.com/chakra-core/ChakraCore/blob/master/Build/Chakra.Build.Default.props#L5-L14
https://github.com/gonetz/GLideN64/blob/master/projects/msvc/GLideN64.vcxproj#L57-L60

Fixed versions like `10.0.19041.0` are not guaranteed in Enterprise editions for stability, `10.0` is safe.

Jj0YzL5nvJ (changes requested):

PropertySheets/Platform.props
JunielKatarn: Thanks for the reference.
I'm aware I was sort of hard coding the specific 10.0.X version, which should last for a while, but Gonetz's approach looks better.

I'll update this.

project64:
I am just being cautious with merging any project settings before I finalize 3.0, which is why I am closing with out merging

PR #2016, opened 2021-03-14
Fix not being able to create issues the old way

vgturtle127:
Sorry, that's my bad. But in all fairness, GitHub basically told me that this would "just work". :(

Fixes not being able to create issues


vgturtle127:
@project64 Please merge ASAP since this is pretty important. 😊

PR #2058, opened 2021-05-29
Updated translation

Felipefpl:
Updated the brazilian portuguese translation.

I'd like to report that the GUI of the settings (including drop-down boxes) is too small to fully display the content of several translated languages. When you got the time/will do to this could you please make the GUI/drop-down boxes bigger and check them to see if they properly display the translated languages?


oddMLan:
Can you show some pictures where the text cuts off so I know where to make the space wider?

Felipefpl:
We cant see all words of the last phrase.



2 phrases inside the drop-down box cant be seen as they are.



The 4 phrases below outside the drop-down box cant be seen with all their words.

oddMLan:
Thanks for reporting. I'll see what I can do.

Felipefpl:
Thanks for caring, i unfortunately didnt look at the problems in the game settings editing GUI (sorry for that )so i'll show the images below:

Aside the 1st phrase in the drop-down box the 2 phrases below that start with "Atrasar a interrupção do" arent fully displayed.

The recompilador "recompiler" section below has a phrase that cant be fully seen too.



My suggestion is to test every translation to see if they fit, it's the only way to be sure all is fine. ;)

PR #2076, opened 2021-07-08
Add Donald Duck and Star Fox 16:9 codes

icup321:
Donald Duck 16:9 code was made from Rayman 2's 16:9 code and Star Fox is just copied from v1.0 but works with v1.1


oddMLan:
Please try adding the author information whenever available.

PR #2081, opened 2021-07-12
Enhancements: Add placeholder support for new enhancement settings (Author, WIP...)

oddMLan:
- Implementation pending.
- Required to merge before #2080


DerekTurtleRoe:
@oddMLan Is multiple 'else if' statements the best way to handle this, or could it be done with a switch statement for all the strings for ROM info? Just curious.

Also it says "implementation pending" does that mean it still needs to be added elsewhere?

flagrama:
Switch statement cases in C++ have to be integers. These would all be case 0.

DerekTurtleRoe:
> Switch statement cases in C++ have to be integers. These would all be case 0.

Ah, you're right. Maybe I was thinking of Python or C#. Thanks for reminding me. Heh.

PR #2125, opened 2021-10-08
[Debugger] JS API: Improve callback checks (fix #2123)

### LINKED ISSUE #2123: [JS Script API] Using events is a lot slower with the v2 API
For example with Ocarina of Time 64, the title screen runs fine at 20 fps without any script running, then just having this one-line script running:

```js
events.onread(0, function(){});
```

brings the fps down to 3.

I do not even have a bad laptop, it looks like the API is so slow it will be unusable in practice unfortunately.

In a Pj64 3.0.1 version (with the "old" api) running OoT64 title screen with the same script, I can't see any slow down.

shygoo:
Still room for improvement but should be significantly faster now


shygoo:
Should be faster than the old api as of that last commit. The onread/onwrite checks could probably be improved further by moving them into the interpreter ops (i.e. like the breakpoints checks), but I think this is good enough for now.

PR #2145, opened 2021-12-05
Linux build and chrome os build via linux terminal

PaperVb64:
Developing the Project64 for another platform could be hard (and most likely a pain in the a**), but could you at least try for linux? Also, could you develop for chrome OS? It doesn't have to be for the MAIN operating system as an app, but as a .deb for use when Linux apps is installed? If you can, thanks a lot!


CrashOveride95:
why did you make what looks like an issue, a pull request

ariahiro64:
this pr is so messed up its actually impressive in the wrong ways

oddMLan:
Average linux user

ghost:
i love this pr

PaperVb64:
this is a suggestion, not an issue.

ghost:
> this is a suggestion, not an issue.

okay, but this isn't a code contribution either. suggestions are usually issues on github. stop being an eta kid. if you'd really like to emulate the n64 on linux use mupen.

project64:
this is not a pull request that can be merged

PaperVb64:
does mupen have the same features tho

LegendOfDragoon:
> does mupen have the same features tho

Nope.

PR #2160, opened 2022-01-06
[Project64-input] N64 Mouse Support

LuigiBlood:
This pull request adds N64 Mouse Support to Project64-input, by making use of a premade configuration (I see no reason to use this with something else than an actual mouse). It reuses the Present value as a means to recognize the device.
I'm aiming to make this easy to use, but it isn't quite working the way I want it.

I also want to add a mouse sensitivity option (not added yet), similar to what the Deadzone and Range options are doing for the N64 Controller. I added a Mouse graphic by myself when the N64 Mouse is selected in the plugin menu.

It also needs a way to lock / unlock the mouse pointer into Project64, I personally don't believe it should be something that the input plugin should do.

Otherwise to what's explained previously in terms of needed features, the N64 Mouse works as intended.

### Proposed changes
  - Add N64 Mouse Support in Project64-input.
    - Mouse Sensitivity option.
  - Add N64 Mouse Support in PIF RAM management in Project64-core.
    - Lock / Unlock the mouse in Project64 window.

~~Currently, when you start up a game with a N64 Mouse present in any controller port, it replaces that controller port configuration with the N64 Mouse configuration. I want it to still have the main controller configuration.
It also cannot recognize controller type changes during gameplay (Mouse -> Controller or viceversa), it still uses the configuration it started with.~~ (Fixed)

Yes.


LuigiBlood:
Currently the Lock Mouse Shortcut isn't really working and I don't know why.

LuigiBlood:
(The fix was dumb)

LuigiBlood:
Currently the implementation works but it relies on GetKeys for shortcut polling... I feel like we could do something better.

EDIT: The thing is like, if the game freezes (without ending emulation), and stops polling the controllers, then you're unable to unlock the mouse. And if you somehow access other menus like settings without unlocking the mouse, the mouse would be still hidden and stuck as long as the emulator is in focus.

LuigiBlood:
ClipCursor allows the mouse to get out when it's not actively polled.

LuigiBlood:
I opted for SetCursorPos instead. That said like, I couldn't figure out how to do the cursor locking properly. This is technically fully featured now, but the locking is kinda meh because there's no real control to figure out when the CPU would be paused or something...

LuigiBlood:
Adding WM_KillFocus and EmulationPaused to the Input Plugin API, while a bit hacky imo, pretty much does the best solution in terms of experience. It allows me to unlock the mouse in pretty much all cases.

PR #2179, opened 2022-01-27
Graceful label truncation

sredna:
The bold text causes text clipping, truncate the label instead.

Changed UI

No.

Don't know. I did not actually compile this. If the resource compiler is not happy with the extra style then the control type needs to be changed from LTEXT to generic CONTROL.


Source/Project64/UserInterface/UIResources.rc
```diff
@@ -485,8 +485,8 @@ BEGIN
     RTEXT           "Game Directory:",IDC_GAME_DIR_STATIC,6,99,60,12
     EDITTEXT        IDC_GAME_DIR,76,97,130,12,ES_AUTOHSCROLL
     PUSHBUTTON      "...",IDC_SELECT_GAME_DIR,209,97,17,12
-    CONTROL         "� GLideN64 �\nFor modern PC",IDC_RADIO_GLIDEN64,"Button",BS_AUTORADIOBUTTON | BS_CENTER | BS_PUSHLIKE | BS_MULTILINE,75,115,70,24
```
project64: were these changes meant to be in the pull request?

sredna: No, the GitHub online editor probably converted to UTF-8. Only the two LTEXT lines are supposed to change.

sredna: Fixed it by using a local editor.

PR #2181, opened 2022-01-27
Don't use PROCESS_ALL_ACCESS when it is not required

sredna:
The PROCESS_ALL_ACCESS define changes at compile time based on the SDK version and the NTDDI_VERSION define. I don't know if you care about Windows XP anymore but `OpenProcess` fails with invalid parameter if you pass it bits it does not know about. It is better to just ask for the required access.

No.

Don't know.

### DIFF
--- a/Source/Common/Util.cpp
+++ b/Source/Common/Util.cpp
@@ -11,7 +11,9 @@
 static bool IsWinVistaOrLater()
 {
 #ifdef _WIN32
-    return GetProcAddress(LoadLibrary(TEXT("KERNEL32")), "CancelIoEx") != nullptr;
+    OSVERSIONINFO vi;
+    vi.dwOSVersionInfoSize = sizeof(vi);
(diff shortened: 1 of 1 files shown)


project64:
was code I wrote a long time ago to deal with zombie versions of project64 that would not allow writing to saves if there was a a zombie process. It was written in windows 2000/xp time, do not remember.  While I no longer test on XP, it should not be hard to keep support of it. and I try to keep things as compatible in general.

sredna:
> was code I wrote a long time ago to deal with zombie versions of project64 that would not allow writing to saves if there was a a zombie process.

Ah, so that is why it acts this way even when you want single instance mode.

I thought about making it possible for `Project64.exe "foo.n64"` to take over the existing instance (if already running). This would be handy when just double-clicking ROM files in Explorer.

Is this zombie hack still required?

PR #2182, opened 2022-01-28
Reuse existing window in single instance mode

sredna:
This converts the "check if already running" setting into a more usable single instance mode. If P64 is already running, a ROM specified on the command line of a new P64 process will run in the existing window instead.

 - It could abort a already playing game if P64 is asked to play another game with a command line argument.
 - It does make the "Terminate existing instance" dialog less likely to appear. It will still appear if the existing window is not responding to messages however.

Don't know

---

 - You might want to rename the "check if already running" setting to something like "Only use one instance".
 - Calling functions from MainWindow in AppInit.cpp is perhaps not the cleanest solution but at some point we have to interact with the existing window before opening a new window much later in WinMain. This is not helped by the fact that the return value of `AppInit()` is ignored in main.cpp and I'm not familiar enough with your codebase to do large refactors.
 - I created `WM_EXTERNALAPI` in a way that could allow for future expansion. Other 3rd-party applications could gain the ability to pause/resume emulation, enter/exit fullscreen and query about the emulation/fullscreen state if more `wParam` handlers are added. I don't know if this is something you think is relevant or not?

### DIFF
--- a/Source/Project64/UserInterface/MainWindow.cpp
+++ b/Source/Project64/UserInterface/MainWindow.cpp
@@ -973,7 +973,7 @@ LRESULT CALLBACK CMainGui::MainGui_Proc(HWND hWnd, UINT uMsg, WPARAM wParam, LPA
             case JSAPI_ACT_RESET:
                 if (g_BaseSystem)
                 {
-                    g_BaseSystem->ExternalEvent((bool)lParam ? SysEvent_ResetCPU_Soft : SysEvent_ResetCPU_Hard);
+                    g_BaseSystem->ExternalEvent(lParam ? SysEvent_ResetCPU_Soft : SysEvent_ResetCPU_Hard);
                 }
                break;
             case JSAPI_ACT_PAUSE:
@@ -1247,7 +1247,7 @@ HWND WINAPI FindOtherInstanceWindow()
 {
     std::wstring classNameStr = CMainGui::GetWindowClassName();
     LPCWSTR className = classNameStr.c_str();
-    SIZE_T data[2] = { nullptr, (SIZE_T) className };
+    SIZE_T data[2] = { 0, (SIZE_T) className };
     EnumWindows(FindOtherInstanceWindowEnumProc, (LPARAM) data);
     return (HWND) data[0];
 }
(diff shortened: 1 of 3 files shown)


Source/Project64/UserInterface/MainWindow.cpp
```diff
@@ -973,7 +973,7 @@ LRESULT CALLBACK CMainGui::MainGui_Proc(HWND hWnd, UINT uMsg, WPARAM wParam, LPA
             case JSAPI_ACT_RESET:
                 if (g_BaseSystem)
                 {
-                    g_BaseSystem->ExternalEvent((bool)lParam ? SysEvent_ResetCPU_Soft : SysEvent_ResetCPU_Hard);
```
project64, line 953: I assume adding the (bool) was to fix a warning, probably better is "lParam != 0"

sredna: > I assume adding the (bool) 

I removed the (bool).

> probably better is "lParam != 0"

Even that is overkill? LPARAM is LONG_PTR which is long or int64.

project64: I would think when lParam is 0 then SysEvent_ResetCPU_Soft , otherwise SysEvent_ResetCPU_Hard .. to do x ? y : z, x has to be a bool so it is not really overkill it just making sure the types are correct

sredna: > to do x ? y : z, x has to be a bool

No? From any arbitrary C standard:

 > The first operand shall have scalar type.
 > The second operand is evaluated only if the first compares unequal to 0; the third operand is evaluated
only if the first compares equal to 0 ...

    int x = ...;
    return x ? a : b; // This is perfectly legal and the same as writing (x != 0).

When using `(bool)x` then it might even depend on your C++ language version? bool is more strict now than it used to be. The compiler might do `x & 1` instead of `x != 0`. Manually doing `x != 0` is of course allowed but not required.

> I would think when lParam is 0 then SysEvent_ResetCPU_Soft , otherwise SysEvent_ResetCPU_Hard

The only caller is `bool bSoftReset = ...; ...WM_JSAPI_ACTION, JSAPI_ACT_RESET, (WPARAM)bSoftReset);`. Why soft is "the true" state, I don't know.

project64:
cleaning up old merge requests. while it seems like it will do the job, my guess 99% of the time someone has two instances is that you have a zobmie instance, which you do not want to re-use. This is why the check for an existing version was put in because if the zombie version has locks on files then the new instance will not run properly

PR #2183, opened 2022-01-30
Fix many warnings

sredna:
This tries to fix a bunch of compiler warnings, many of them related to 64-bit. I tried to stay away from the CPU/emulation and video stuff but there are many warnings left there that need attention.

Hopefully not but I can't rule it out with this many changes. There are a couple of cases where local variables shadowed some global variables and I renamed the local. It would perhaps be better if some of the globals could be renamed with a `g_` prefix, especially the ones with short names like r, g, b and buf but I really don't want to risk breaking any emulation related code. I realize the `safe_truncate_cast_assign` function is very ugly but I don't really have a better idea of how to truncate without a warning but not potentially break code if the destination becomes larger in the future. I only used it in one place for now in case you don't like it. The point of it is to retain the speed of `dest = (smaller int cast) source` without just a cast to silence the compiler. If dest becomes larger than the cast the warning should come back automatically or `static_assert` if available.

Don't know.

### DIFF
--- a/Source/Common/IniFile.cpp
+++ b/Source/Common/IniFile.cpp
@@ -504,7 +504,9 @@ bool CIniFileBase::DeleteSection(const char * lpSectionName)
         return false;
     }
     m_File.Seek(m_CurrentSectionFilePos, CFileBase::begin);
-    long DeleteSectionStart = m_CurrentSectionFilePos - (strlen(lpSectionName) + strlen(m_LineFeed) + 2);
+    size_t tmp = (strlen(lpSectionName) + strlen(m_LineFeed) + 2);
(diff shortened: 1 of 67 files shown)


project64:
it might all be fine, reviewing it is difficult cause there is so many changes, it might be better to do each component separately.

sredna:
> it might all be fine

I do hope so. I only touched things I was confident about.

 >  reviewing it is difficult cause there is so many changes

I suppose but splitting it up does not reduce the amount of work. At least half of the changes should be easy to review. The Win32 stuff I am comfortable with and most of those are just changing uint32_t to size_t/LPARAM.

project64:
cleaning up, closing old pull request

PR #2209, opened 2022-03-10
Update the current RDB with some minor fixes and improvements

DerekTurtleRoe:
### Proposed changes

This updates the current RDB with some small label fixes. A majority of the games marked as having issues or not supported actually work, but were never changed.

I am also redefining what "compatible" means.  This will just mean "can it work in Project64" as opposed to "can it work with default settings, default plugins, and/or with specific settings?" I'm doing this in preparation for larger additions to the RDB, which will for the most part be similar versions of already working games or alternate builds.

Some games actually did have issues, but if you eliminate the plugins causing issues or other factors, many have core issues (such as virtual memory issues, CPU bugs, etc).

I also added another dump of NBC-LFC Kings of Porn Vol 1, specifically the alpha 2 dump. This dump works on real hardware, just needed to be added to the RDB for it to work in Project64.

I removed several bad 'Good Name' titles and replaced them with their proper titles as well.

I also fixed the Y2K Demo by WT_Riker by forcing 32-bit engine off and Delay DP off. This makes it run like it should.

Besides the games I specifically mentioned, the only other things that were changed were the status.

TL;DR updated RDB with more accurate titles, fixed some incorrect status names, and fixed a couple of games while I was at it.

No, the opposite in fact. ;)

Yes.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -204,7 +204,7 @@ Status=Compatible
 [B98BA456-5B2B76AF-C:4A]
 Good Name=64 de Hakken!! Tamagotchi - Minna de Tamagotchi World (J)
 Internal Name=ÐÝÅÃÞÀÏºÞ¯ÁÜ°ÙÄÞ
-Status=Issues (plugin)
+Status=Compatible
 32bit=Yes
 Counter Factor=1
 RDRAM Size=4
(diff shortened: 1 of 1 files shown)


DerekTurtleRoe:
One thing I forgot to mention is that some games do have core issues, and I marked those games as such. 😄

PR #2220, opened 2022-03-26
Fix TLB map bug involving 0 sentinel value

### LINKED ISSUE #2218: [Bug]: Frequent crashes in Paper Mario on Windows 11 due to TLB entries not working
### Expected behavior?

Jumping on enemies should not show the in-game crash screen.

More specifically: When Mario lands on the ground after hitting an enemy with a jump attack, a dust particle loads, This creates a TLB mapping for address E000C000-E000CFFF, and then jumps to some TLB mapped code at E000C160, which executes without any problems.

### Actual behavior?

In-game, the crash screen appears:

image

The behaviour in the game is exactly what it would be if the TLB mapping had not been created at all - it behaves as if E000C000 isn't mapped to anything. I also see the same thing in the memory viewer window, it acts like there is no mapping.

Strangely, the TLB debugger *does* show the mapping that was supposed to be created, even though the ingame behaviour and memory viewer behave as though it doesn't exist.

### Steps to reproduce the problem

~~Play the game normally through the prologue, using Mario's jump attack and Goombario's bonk to attack enemies. Some of the time (not too rarely) the crash screen will show.~~ EDIT: Just wait on the title screen instead, it will always crash during the demo battle with the Monty Moles.

It does at least seem to be deterministic - I've made a savestate right before the crash happens, and loading the state always leads to the crash.

There are reportedly many other ways that the crash can happen, though the visible result will always be a crash screen with "PC: E0XXXXXX" in the top left.

### Error messages?

Only the ingame crash.

### What game were you playing?

Paper Mario (USA)

The issue also happens frequently in rom hacks for the game.

### Specifications and build details

Project64: 3.0.1.5664-2df3434 32 bit

OS: Windows 11 21H2 (OS Build 22000.556)

### Using included plugins

- [X] I can confirm I am using the included RSP, audio, input, and graphics plugins

### Plugins used while the issue occurred

(default)

### Copyright Compliance

- [X] There are no infringing files attached to this bug report.

ethteck:
Fixes #2218

### Proposed changes
  - Makes -1 the indicator value for invalid tlb mapping, rather than 0

Not as far as I'm aware

Yes

### DIFF
--- a/Source/Project64-core/N64System/Mips/MemoryVirtualMem.cpp
+++ b/Source/Project64-core/N64System/Mips/MemoryVirtualMem.cpp
@@ -83,8 +83,8 @@ void CMipsMemoryVM::Reset(bool /*EraseMemory*/)
     {
         size_t address;
 
-        memset(m_TLB_ReadMap, 0, 0xFFFFF * sizeof(size_t));
-        memset(m_TLB_WriteMap, 0, 0xFFFFF * sizeof(size_t));
+        memset(m_TLB_ReadMap, -1, 0xFFFFF * sizeof(size_t));
+        memset(m_TLB_WriteMap, -1, 0xFFFFF * sizeof(size_t));
         for (address = 0x80000000; address < 0xC0000000; address += 0x1000)
         {
             m_TLB_ReadMap[address >> 12] = ((size_t)m_RDRAM + (address & 0x1FFFFFFF)) - address;
@@ -318,7 +318,7 @@ CFlashram* CMipsMemoryVM::GetFlashram()
 
 bool CMipsMemoryVM::LB_VAddr(uint32_t VAddr, uint8_t& Value)
 {
-    if (m_TLB_ReadMap[VAddr >> 12] == 0)
+    if (m_TLB_ReadMap[VAddr >> 12] == -1)
     {
         return false;
     }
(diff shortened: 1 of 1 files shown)


project64:
I can not merge this cause more work than just this need to be done for this to work, cause the recompiler has not been touched .. it is checking 0 still on all reads/writes.

But I take your point 0 might be a valid value not just not set.

I am looking at changing how some of this does work in the future cause I am trying to remove the large amount of memory I use and the use of structured exception handling

ethteck:
If I were to change the behavior in the recompiler, would this be mergeable? Or should I just close this? I mainly opened it to demonstrate a possible fix, and at least in my testing, everything already works fine. Maybe I didn't test the recompiler..

DerekTurtleRoe:
I would say leave it open until another solution is implemented at the very least, so other people can find it.

project64:
does https://github.com/project64/project64/commit/fbf65bce12fd7288f938aac27c88e25d54cdbd89 fix your issues?

flagrama:
User Timberwolf on the Discord server has fewer crashes after that fix, though still gets a "TLB exception on load" in a particular instance:

> If star storm is used on dark koopas in the sewers, this happens on build Dev-4.0.0.5778-80c21b4. That said, many of the other crashes are gone now from this particular build.

project64:
That is good to hear.. There is more work to be done to remove exceptions, but at least as I work towards it hopefully things work better

project64:
can I close this now?

ethteck:
Yeah, I haven't tried the latest build but any further issues can be addressed in another pr. Thanks!

PR #2223, opened 2022-04-03
Added a "do not show nag" checkbox

AchilleBailly:
I added a checkbox in the general settings tab to disable the automatic display of the nag window on start-up

Fixes #

### Proposed changes
  - the corresponding setting is set to false by default, the user will have to enable it himself to disable functionality
  - also added the corresponding line in the English language file and its translation in the French one

I don't think so.

Yes, at least to what I could test.


flagrama:
It used to be possible to disable it with a line of text in the ini. There is a reason that was changed to not be so easy to bypass. Support the software you use.

project64:
I can appreciate doing the work for the pull request and wanting to see this happen, but it is not likely to happen. If you want to talk to me about it you can always find me on discord (zilmar) and I am happy to chat about it.

PR #2234, opened 2022-05-16
Code refactor and fixed bugs

GermanAizek:
Minor optimization, code refactor, fixed memleak and bug with rgb struct in different sections of project.

### Proposed changes
  - Slight code simplification and bug fix

  - No, just do a code review

  - Yes, I checked test game

### DIFF
--- a/Source/Common/File.cpp
+++ b/Source/Common/File.cpp
@@ -66,7 +66,7 @@ bool CFile::Open(const char * lpszFileName, uint32_t nOpenFlags)
         return false;
     }
 
-    if (lpszFileName == nullptr || strlen(lpszFileName) == 0)
+    if (lpszFileName == nullptr || lpszFileName[0] == '\0')
     {
         return false;
     }
--- a/Source/Common/IniFile.cpp
+++ b/Source/Common/IniFile.cpp
@@ -180,7 +180,7 @@ void CIniFileBase::SaveCurrentSection(void)
         }
         m_File.Write(SectionName.get(), (int)strlen(SectionName.get()));
         m_CurrentSectionFilePos = m_File.GetPosition();
-        m_SectionsPos.insert(FILELOC::value_type(m_CurrentSection, m_CurrentSectionFilePos));
+        m_SectionsPos.emplace(m_CurrentSection, m_CurrentSectionFilePos);
     }
     else
     {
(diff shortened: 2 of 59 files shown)


project64:
Thank you for making the pull request and trying to improve the code base.

there is some good changes in here, nothing that is wrong that I can see at quick inspection.

There are things like changing the desonstructor from {} to = default, that might be correct and just cause I have done it that way for a long time does not mean it is right, there is changes like that mixed in that I can not just merge it straight away. 

Smaller changes make it easier to merge and get in.

GermanAizek:
@project64, I hope that many changes you will merge, I also updated PR with commit that affects cleaning of sensitive data in MD5 functions.
https://github.com/project64/project64/pull/2234/commits/e616ab9c8db930ac12cefaf534e3e2167d58d9d8

project64:
there is no sensitive data in the application. Why do you think that md5 needs to be changed to use RtlSecureZeroMemory?

if you want to talk to me about this or anything related to project64 source code you can always talk to me on discord

GermanAizek:
@project64, comment says.
https://github.com/project64/project64/blob/e616ab9c8db930ac12cefaf534e3e2167d58d9d8/Source/Common/md5.cpp#L155-L165

DerekTurtleRoe:
@GermanAizek I think what Zilmar is saying is that this MD5 hash isn't used for anything in regards to security and doesn't involve verification of anything transmitted over a network, it is just the MD5 hashing algorithm used for hashing files. I don't necessarily oppose this change, but it would be interesting to see if this affects performance of the emulator in terms of latency and calculation times on the same system.

project64:
I have two main uses of MD5 in Project64, After the rom is loaded a md5 of the loaded content is made. The second is when a block of code it generated, I will md5 it for use in future comparisons to make sure the code has not changed. This could be some type of crc calculation instead to try to improve the speed. 

If I was hashing passowrds then this  could matter a lot more. For Project64 there is nothing sensitive being hashed.

GermanAizek:
@project64, Do you accept my pr after removing `RtlSecureZeroMemory`?

project64:
the use of RtlSecureZeroMemory is not wrong, mostly cause it is 3rd party software I try to leave it untouched.

I appreciate that your trying to improve the code base so want to accept the pull request

there might be nothing wrong in any of it and I could.

but I like to try make sure I understand all the changes and what the impact is going to be, with so many changes in the pull request that makes it hard to say not a single change is not going to have a negative impact.

more likely it is take a copy of this code and the current code base and manually copy the easy to determine changes one by one. 

Once that is finished, I might have nothing left, or maybe a whole for me to think about and work out if it has merit to use or why I have issues with it.

GermanAizek:
@project64,
OK, I'll split different changes in functionality and make separate PR.

project64:
cleaning up, closeing old pull request

project64:
cleaning up pull requests, closing old pull requests

PR #2251, opened 2022-07-18
Move the API and example scripts to their own directory

DerekTurtleRoe:
I also changed the names in the example script to match.

### Proposed changes
  - Move documentation HTML and JavaScript example scripts to their own directory
  - Rename "JS-API-Documentation.html" to just "Documentation.html"

No.

Yes.


project64:
Error on line 42 in C:\projects\project64\Source\Installer\Installer.iss: Source file "C:\projects\project64\Scripts\example.js" does not exist.
  Compile aborted.

shygoo:
Without source changes, the scripts must remain in the /Scripts directory for them to show up in the script console. And I named the documentation file the way it is so it wouldn't be confused with the plugin api.

Anyway I had planned to move the scripts directory, api doc, and other debugger data to a /Debugger folder and have a warning shown in the script console if the old scripts directory is detected.

DerekTurtleRoe:
Hmmm. I assumed these weren't directly linked to anything so changing them would be straightforward and not break anything. 😩

Do I just need to change the Inno setup script only, or is there something I need to change in the actual core source as well for this specific change? or should I just close this and we can deal with it at a later date?

shygoo:
I can deal with it later. Moving the scripts folder is going to break existing setups so there will need to be a warning

DerekTurtleRoe:
Alright, sounds good. Appreciate it!

PR #2292, opened 2022-10-24
Fix MBC30 transferpak support in nrage input plugin

### LINKED ISSUE #1771: Transfer Pak MBC3 emulation assumes 4 RAM banks
The function WriteCartMBC3 in nragev20/GBCart.cpp, when writing to addresses 0x4000–0x5FFF to select a RAM bank, masks the bank number with 0x03 (unless it’s one of the special real-time clock banks 0x08–0x0C). While this works for most MBC3 Game Boy cartridges, which have 4 RAM banks, it is not correct for Pocket Monsters - Crystal Version (Japan), which has 8 RAM banks. Because of this issue, Pokemon Stadium Kin Gin (Japan) cannot read data from Crystal’s banks 0x04–0x07, which are used for the Mobile Stadium feature.

I think WriteCartMBC3 can use the same logic as WriteCartMBC5, which just masks the bank number with 0x0F. It doesn’t look like any specific logic is needed for selecting the RTC banks.

MatandoYT:
Changes maximum number of possible RAM banks in mbc3 carts to 8 instead of 4 to support the mbc30 variant. This allows for proper handling of the Pokemon Crystal Japanese saves in Pokemon Stadium 2.

Fixes #1771 

### Proposed changes
  - Change maximum number of RAM banks in WriteCartMBC3 from 0x03 to 0x07.

No. I've tested regular mbc3 saves with Pokemon Stadium to ensure they didn't break in the process.

Yes.

### DIFF
--- a/Source/nragev20/GBCart.cpp
+++ b/Source/nragev20/GBCart.cpp
@@ -889,7 +889,7 @@ bool WriteCartMBC3(LPGBCART Cart, WORD dwAddress, BYTE *Data)
     {
         if (Cart->bHasRam)
         {
-            Cart->iCurrentRamBankNo = Data[0] & 0x03;
+            Cart->iCurrentRamBankNo = Data[0] & 0x07;
             DebugWriteA("Set RAM Bank: %02X\n", Cart->iCurrentRamBankNo);
             if (Cart->bHasTimer && (Data[0] >= 0x08 && Data[0] <= 0x0c))
             {


Squall-Leonhart:
be better to bring nrage up to date with latest vba-m code,  its currently using ported code from 1.4

how vbam handles bank select also mitigates corrupting the rtc block.
https://github.com/visualboyadvance-m/visualboyadvance-m/blob/3f571fb930384476e13ed3669df6bc9bedc067a7/src/gb/gbMemory.cpp#L386

DerekTurtleRoe:
If you are going to do a whole bunch of work to port something over, wouldn't it make sense to use something like SameBoy instead of VBA-M for more accurate behavior?

Not sure how much code is needed, and what exactly is emulated. But SameBoy or Gambatte or something might be better for that.

Squall-Leonhart:
cart read/writes aren't really subject to accuracy, they either work or they don't.

project64:
I am hesitant to add code in related to functionality with out a test rom

DerekTurtleRoe:
The Game Boy MBC3 is pretty well documented, and there are test ROMs for it as well, including testing the real time clock functionality.

https://gbdev.io/pandocs/MBC3.html
https://gbdev.io/resources.html
https://github.com/aaaaaa123456789/rtc3test
https://github.com/c-sp/gameboy-test-roms

If you mean the transfer pak, there is documentation here:

https://sites.google.com/site/consoleprotocols/home/nintendo-joy-bus-documentation/n64-specific/controller-accessories/transfer-pak

DarkshadeDX:
Hey, just putting out a reminder on this - as Matando has mentioned, Japanese Pokémon Crystal can be used as a test ROM.

Squall-Leonhart:
This change did nothing because it was based on faulty assumptions.

MatandoYT:
> This change did nothing because it was based on faulty assumptions.

It does in fact work and has been tested. If you wish to try it yourself use either the Japanese version of Pokemon Crystal or the pokecrystal-mobile-en in the gb-mobile repo. 

Either way it works and there are no issues with the code. It was merged for a reason.

*Edit*
Although, I will note that the proper way to implement it would have been to port the newer code over from vba-m like you had previously mentioned. It also would have been beneficial for you to state what "faulty assumptions" you are talking about instead of blindly throwing accusations.

flagrama:
I'm not sure exactly what Squall is going for here, but the regular MBC3 does support 32K in 4 RAM banks in most situations. Crystal is the only one that uses 64K in 8 RAM banks. For the former, mirroring the RAM if a higher bank is probably correct. (I can't find documentation that it works this way, but it does commonly happen that way for RAM/ROM hardware) There should be some way to detect that it uses the larger RAM and ROM size and only allow access to those banks in that case, but probably otherwise return the old behaviour.

MatandoYT:
> I'm not sure exactly what Squall is going for here, but the regular MBC3 does support 32K in 4 RAM banks in most situations. Crystal is the only one that uses 64K in 8 RAM banks. For the former, mirroring the RAM if a higher bank is probably correct. (I can't find documentation that it works this way, but it does commonly happen that way for RAM/ROM hardware) There should be some way to detect that it uses the larger RAM and ROM size and only allow access to those banks in that case, but probably otherwise return the old behaviour.

It's actually described in the ROM header itself. Although, I don't have the information at hand.

Squall-Leonhart:
It is not, MBC30 and MBC3 are both MBC3 identifier 

The Rombank code needs to be modified to support up to 0x0FF, as technically MBC30 can go up to 4MB when all the address pins are used,  Crystal JP leaves one unused.

MatandoYT:
Yes it does. 0x148 for the ROM size and 0x0149 for the RAM size. So you'd check if the cartridge type is MBC3 and then check the RAM and ROM size after that.

https://gbdev.io/pandocs/The_Cartridge_Header.html

Issue #2303, opened 2022-11-14
[Bug]: Cruis'n USA v1.0 not working correctly


xiartic:
### Expected behavior?

This game seems to have problems running with both stable & nightly releases.

### Actual behavior?

Cruis'n USA hangs after the controller screen in both stable 3.0.1 and dev 4.0.0-6016-989827c.

Video of Cruis'n USA:

https://user-images.githubusercontent.com/78692485/201761745-1cace8a7-82a9-413c-8c76-b1c6c2a91ca0.mp4

### Steps to reproduce the problem

1. Try to play Cruis'n USA
2. 
3.
4.
5.
...

### What game were you playing?

Title: Cruis'n USA

Nintendo 64 game or Nintendo 64 Disk Drive: Nintendo 64

### Specifications and build details

Project64 3.0.1-5664-2df3434

OS: Windows 11 Home 21H2

CPU: Intel Core i5-3350P

RAM: 8 GB DDR3

Graphics: Nvidia Geforce GTX 1060

### Using included plugins

- [X] I can confirm I am using the included RSP, audio, input, and graphics plugins

### Plugins used while the issue occurred

Graphics plugin: GLideN64 rev. f431466 HLE
Audio plugin: Project64 audio plugin 3.0.1-5664-2df3434
Input plugin: Project64 input plugin 3.0.1-5664-2df3434
RSP plugin: RSP plugin 3.0.1-5664-2df3434

### Copyright Compliance

- [X] There are no infringing files attached to this bug report.

Squall-Leonhart:
Cruis'n USA has no issues getting in game with U 1.2, you need to specify the region and rom version as some issues are specific to a specific version of the game.

Exotica is another SP Dram not in RDRAM issue.

xiartic:
I tried (U) v1.0 and it would not get further than the controller screen. Looks like (U) v1.1 and v1.2 work fine.

Squall-Leonhart:
i don't get a freeze, i get a pak access error instead on 1.0

xiartic:
Interesting. Is the hash of your ROM: A4E241B2535D6A54B62D08AF53CA83B0?

Squall-Leonhart:
Freeze is due to PJ64 Inputs memory pak emulation.

~~My suspicion is the pak size is emulated to be too large for the 1.0 variant, 1.1 and 1.2 have support for larger pak sizes and rumble pak respectively~~

Squall-Leonhart:
[...]

PR #2304, opened 2022-11-15
Extend mempak Index Table to the intended 256 bytes, so that the default checksum is actually correct, and include the backup of that data.

Squall-Leonhart:
The Table was only 224 bytes instead of 256, this was invalidating the checksum causing certain games to fail on first time Mempak access and creation.

Also adds the backup region just in case theres a rare game that breaks without it.

Fixes #
At this point, only Cruis'n'USA 1.0, its all i know affected

### Proposed changes
Extends the affected area to 256 bytes.

No

Yes, so far.

Fixes part of #2303.


bryc (approved):
This should be good now.

This came as a result of 94cdc8d, where I accidentally missed the final 32 bytes, thus invalidating a checksum. Most games are able to silently handle this without alerting the user, so it was not caught until now.

This proposed solution goes beyond just adding the missing 32 bytes, but also the missing backup data. This is more robust and doesn't rely on games' ability to repair the data at all, which as we've seen, is not always possible.

Cruis'n USA (1.0) has buggy Controller Pak support; it is unable to format or repair data in many cases, resulting in the game crashing, which is likely why 1.1 and 1.2 were made, which don't suffer from this issue.

PR #2328, opened 2023-01-02
RDB fixes and optimizations

DerekTurtleRoe:
Adds some internal names and compatibility status entries to those that were missing and removed all entries relating to saves, since we autodetect the save type.

This is meant to be a replacement of the other PR, since it removed the 32-bit engine hack. The 32-bit engine hack isn't present in the CPU interpreter, and it is optional in the recompiler, so it will be kept for now.

The improvements to categorization and such can also be handled in a separate pull request.

### Proposed changes
  - Add some miscellaneous missing entries
  - Remove entries forcing save type, since we can autodetect it

No.

Yes.

### DIFF
--- a/Config/Project64.rdb
+++ b/Config/Project64.rdb
@@ -180,7 +180,6 @@ Status=Compatible
 Culling=1
 Emulate Clear=1
 Linking=Off
-RDRAM Size=4
 
 [1FBAF161-2C1C54F1-C:41]
 Good Name=1080 Snowboarding (JU) (M2)
@@ -190,7 +189,6 @@ Status=Compatible
 Culling=1
 Emulate Clear=1
 Linking=Off
-RDRAM Size=4
 Self Texture=1
 SMM-Cache=0
 SMM-PI DMA=0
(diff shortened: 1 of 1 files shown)


project64:
you going to have an easy time getting this commit as more targeted fixes then huge changes to everything. Any disagreement on a single game is going to block all fixes.

Did you remove RDRAM Size=4 from every game? Was this to actually fix something or just cause you think 8mb is better and should not cause a problem?

DerekTurtleRoe:
@project64 No, I didn't change every game. The games I didn't change were Iggy's Reckin' Balls and Space Station Silicon Valley, as those crash on real hardware with 8 MB of RDRAM.

As for all other titles, this doesn't "fix" any games, but it doesn't break any either, at least in my testing. I tested every game with a blank RDB, as well as forcing 4 and 8 MB of RDRAM, and none of them broke.

On real hardware, all the games work with 8 MB of RDRAM other than the 2 mentioned above, and there is a fixed revision of Space Station Silicon Valley that works with 8 MB of RDRAM, which I added an exception for.

I also didn't remove every game that had a forced save type, only games that worked fine without the option in the RDB. I tested all games with and without forcing the correct save type, and none of them were broken with my changes. Most of the games that I kept the save type were ROM hacks, N64DD games, or iQue games.

Everything else should be fairly trivial changes, like adding some missing internal name entries and adding some status entries to games that didn't have it before. I tested those games as well and they worked fine.

PR #2333, opened 2023-01-17
Add second checkbox for users to confirm they are using latest nightly before reporting bug.

Squall-Leonhart:
Wording might be up for changing, it occurred to me after that some users might still want to report bugs for 3.0, despite it no longer being the active focus for dev (perhaps 3.0 should be grandfathered at this point and issues for it no longer accepted?)


project64:
maybe also need to check/report if it works in interpret too

project64:
closing old pull requests

PR #2334, opened 2023-01-21
Added playtime counter on a per ROM basis.

DidgeridooMH:
This PR is to add a timer for total playtime. It simply takes the time from when the user starts a game to when they end it. This time is then added to a playtime counter in a config document. It currently uses the name of the ROM from the database to keep track of which game has what time.

### Proposed changes
  - Add playtime as field on the ROM browser.
  - Add of a routine that starts a "timer" and writes its time to a document on closing the emulator.

No.

Yes, it runs games and when you exit the game the ROM browser shows an updated playtime.

### DIFF
--- a/Source/Project64-core/Multilanguage.h
+++ b/Source/Project64-core/Multilanguage.h
@@ -195,6 +195,7 @@ enum LanguageStringID
     RB_FORCE_FEEDBACK = 318,
     RB_FILE_FORMAT = 319,
     RB_NAME = 321,
+    RB_PLAYTIME = 3114,
 
     // Select ROM
     SELECT_ROM_DIR = 320,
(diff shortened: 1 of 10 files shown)


Squall-Leonhart:
Please don't be storing the value for this inside the cfg file.

DidgeridooMH:
@Squall-Leonhart I'm not sure I get what you mean. This is getting stored in a separate file much like how the notes are stored. Is there a different directory that data is usually stored in?

Squall-Leonhart:
> This is getting stored in a separate file much like how the notes are stored. Is there a different directory that data is usually stored in?

Great, but  you appear to be storing based on GoodName, this might cause a problem where romhacks using the same goodname will update the playtime on all variants of it (or even the the original)

Shouldn't be a frequent issue,  but the "Super Luigi 64 - Definitive Edition" hacks are an example of it.

Instead, using the Romhash, and print the Goodname just like is done with the Cheat and User RDB would be better.

PR #2337, opened 2023-02-01
Add generic "Ocarina of Time + Majora's Mask Randomizer" entry

ghost:
A new randomizer combining two games, it requires 8MB RDRAM Size. I copied what was done for DK64 Randomizer.


Squall-Leonhart:
not good enough, OOT needs protect memory.

Nax:
OoTMM has specific fixes to work around the memory protection requirement, so it's not strictly required.
However it would be nice to include it anyway, so we can remove these hacks on our side going forward.

PR #2339, opened 2023-02-09
[SortRDB] Use auto for _Pairib starting MSVC 14.2

JunielKatarn:
Use `auto` declaration to avoid accessing internal member `_Pairib` on MSVC 14.2 and later.

See https://github.com/project64/project64/commit/05c2b59c3a5f130cf5eccc37a18cdd0a2cb4c105

### Proposed changes
  - Avoid directly declaring internal member `_Pairib` on MSVC 14.2 and later.

No.


project64:
I removed all usage of pairb

https://github.com/project64/project64/commit/4390a0926c8d1aca0460afe167afca53e997b109

PR #2391, opened 2023-10-18
Update Project64.sln

ghost:
Fixes # Update Visual Studio to the actually latest version.


Squall-Leonhart:
if Zilmar was using the latest visual studio, this file would have already updated itself.

YigaUpdate:
that's to say ?

ghost:
But is it good for the update ?

PR #2417, opened 2024-03-18
Adding functionality to copy a specified number of commands in the Commands window

Summate:
It's personally very difficult for me to work with the assembly directly without being able to put it in a text file and play around with it. I didn't see anything that would allow me to copy sections of code that I'm looking at specifically. There is some precedent, I think NO$PSX allows this sort of thing although it's specified in bytes for some reason as opposed to commands and it only dumps to file.

There may be edge cases and other miscellanea of concern, this is really simple and straightforward. The dialog defaults to copying a single command and allows the user to specify a quantity of commands to copy. Attached are screenshots showing the feature flow.

It works with the symbols and other annotation features, those end up rendered out as well with a spacial buffer to mimic their look in the Commands window.

No, it strictly widens the functionality of the application.

Seemingly.


Summate:
Not sure what to do about the clang-format error. I guess uppercase probably sorts before lowercase, then? Regardless, `Debugger-CopyCommands.h` has to be present before the inclusion for `Commands.h` because the symbol(s) won't be defined otherwise. I used `Debugger-AddSymbol` as my template; it would be more correct for `Commands.h` itself to include the file. Is this the preferred path?

PR #2419, opened 2024-05-10
Set file versions dynamically

JunielKatarn:
Leverages MSBuild targets to gather the Git info and pass it to the source and resource compilers at runtime.

This removes the need to call a separate executable for each project to be versioned, as well as parsing and copying header files over.

### Proposed changes
  - Define global target `GetGitVersion`.
  - Pass version preprocessor definitions from MSBuild at runtime.

No

Yes

Validated updated versions using the following PoerShell command:
```pwsh
(Get-Item $fileToTest).VersionInfo | select Filename, FileVersion, FileVersionRaw, LegalCopyright
```

Sample result:
```
FileName                        FileVersion             FileVersionRaw LegalCopyright
--------                        -----------             -------------- --------------
C:\pj64\Bin\Debug\Project64.exe Dev-4.0.0.6408-914c95d2 4.0.0.6408     Copyright (C) 2024
```


project64:
I had a bit of a look at this, I got your code and set it to the autoversion branch. 

It builds fine in visual studio 2015, but the GIT_VERSION and VERSION_BUILD do not seem to change in the main exe, I did not look at other binaries.

Did I do something wrong, I know the current system you can get code from github, open in visual studio and build and it will show infomration about version of the code is being to used to make that build

JunielKatarn:
To be honest, I don't recall checking the final EXE, but only the plugins.
(See the PowerShell snippet in the description).

I'll give this a thorough look tonight.

JunielKatarn:
> I had a bit of a look at this, I got your code and set it to the autoversion branch.
> 
> It builds fine in visual studio 2015, but the GIT_VERSION and VERSION_BUILD do not seem to change in the main exe, I did not look at other binaries.
> 
> Did I do something wrong, I know the current system you can get code from github, open in visual studio and build and it will show infomration about version of the code is being to used to make that build

@project64 `Project64.exe` properties update fine on my local builds.\
**Note**, I am using Visual Studio 2022.

image

Can you please check the results of this PowerShell command on you local build?
```pwsh
gi Bin\Debug\Project64.exe | select VersionInfo | fl
```

JunielKatarn:
I was able to reproduce the issue with Visual Studio 2015.
I believe the culprit is the older MSBuild version.

Perhaps it's a good opportunity to move to a more recent version of Visual Studio?
Note: Newer VS versions allow installing older compiler versions.
image

JunielKatarn:
I have successfully reproduced the issue using Visual Studio 2015.
I suspect patching doesn't work because of the older MSBuild version being used.

Perhaps it's a good excuse to upgrade to newer VS versions?
Note, if the compiler version is an issue, VS 2022 allows installing older compiler (MSVC) versions side by side, while using the modern MSBuild engine which correctly executes the needed tasks.

image

JunielKatarn:
As suspected, the issue is the older version of MSBuild.
Installing the MSVC 14.0 tools into Visual Studio 2022 yields the right results.
image

project64:
closing old pull rquests, as i just do not see this happening at this time

PR #2426, opened 2024-05-29
Update vcruntime140_1.dll

Xphalnos:
### Proposed changes
  - I updated vcruntime140_1.dll from 14.29.30135.0 to 14.40.33810.0

No, just updating a binary file.

Yes, this compiles perfectly !


Squall-Leonhart:
This version of the vcruntime is not windows xp compatible, and at some point zilmar wants to address the regressions that broke XP.

Xphalnos:
However it is marked that the minimum OS is Windows 10.

Squall-Leonhart:
You confuse support with function.

PR #2427, opened 2024-06-01
Fix implementation of quad3d for f3dex 0.95 on pj64 video plugin

MegaMech:
This change shouldn't break starfox because I believe only mk64 uses the quadrangle command.

mk64 does actually have a debug wire frame mode for some courses. But as far as I can tell, it's not using line3d (Now that I'm thinking about it, this wire frame might actually be another renderer bug).

Fixes #
 - Rendering of startup logo for mk64
 - Fixes bug with `m_force_quad3d` where the value changed inconsistently.

### Proposed changes
  - Removes `m_force_quad3d` bool
  - Removes line3D implementation for f3dex because that command requires the line3d micro-code to work.
  - Rename line3D to quad3D


MegaMech:
Confirmed patch does not break sf64. And I confirmed this is using G_TRI2 implementation on the renderer side.
image

MegaMech:
Yoshi's story uses g_line3d implementation according to the code-base. So I've split g_quad and line3d into separate functions.

PR #2432, opened 2024-07-28
Remove Discord-RPC from the project

DerekTurtleRoe:
This pull request removes Discord-RPC code as well as all references and documentation of the Discord-RPC features from Project64.

The reason this was done is because Discord-RPC has been deprecated for a long time now, and the new "Discord Game SDK" API hasn't been implemented by us yet, but is cross-platform and much more portable. This also eliminates some technical debt and makes the code base smaller and a bit easier to manage until while we prepare Project64 to be much more portable and hopefully get an Android and Linux port up and running at some point.

Discord-RPC was disabled by default, so there isn't really a privacy benefit to most users. However, this does also prevent RPC applications from accessing the game name and/or window title in theory, but they would have to be using the old methods still. But you could argue that removing this is a small benefit to some people.

### Proposed changes
  - Remove all Discord-RPC files
  - Remove all references to Discord-RPC
  - Remove all Discord-RPC code and associated documentation
  - All references to the Project64 Discord were kept intact

Yes.

This needs to be tested.


DerekTurtleRoe:
All checks failed, so that's a good sign! 😅 

Please check my work on this PR, I did it mostly very late at night.

PR #2464, opened 2025-02-14
Fix: NRage plugin gets stuck in initialization in certain cases

pcca-matrix:
The function wscanf was incorrectly used to parse VID_ and PID_ values from a string, causing failures in device detection. wscanf expects input from stdin, whereas in this case, the device ID string is stored in memory.

Using swscanf ensures correct parsing from memory instead of waiting for stdin

This fixes the issue where XInput devices not being detected properly at plugin init.

### Proposed changes
Replace wscanf with swscanf to correctly extract the vendor and product IDs from the device ID string.

No

Yes

### DIFF
--- a/Source/nragev20/XInputController.cpp
+++ b/Source/nragev20/XInputController.cpp
@@ -105,10 +105,10 @@ BOOL IsXInputDevice( const GUID* pGuidProductFromDirectInput )
                         // If it does, then get the VID/PID from var.bstrVal
                         DWORD dwPid = 0, dwVid = 0;
                         WCHAR* strVid = wcsstr( var.bstrVal, L"VID_" );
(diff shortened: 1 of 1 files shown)


Squall-Leonhart:
good catch, this must have been fucked in the nrage for pj64 refactor because i remember using this as is right from the DX SDK sample.

edit: or not, i see its been wrong since i merged the code into the original project

what control were you seeing this with, i never have personally.

pcca-matrix:
issue occurs when the plugin attempts to initialize in streaming mode (e.g., Moonlight) or when the emulator is launched as a subprocess of a launcher using an Xbox360 or Xbox1 controller.
 
haven’t tested with other controllers but the problem should occur with any XInput device in these scenarios.

Squall-Leonhart:
I personally sign off on this change @project64

PR #2465, opened 2025-02-18
Clean up Windows Start Menu entry

RedBlackAka:
Cleans up the Windows Start Menu entry for the Windows installer.

[Add description of the PR here]

Having an uninstall shortcut is obsolete and an unclean practice in Windows 10+, as one simply can right-click and uninstall. Mostly the same for support, Start Menu is not the right place for it and it can still be added under Apps and Features if needed. This change makes Project64 directly appear on the Start Menu, not requiring further clicks to expand a folder and making the appearance clean.

### Proposed changes
  - Removes uninstall and support shortcut
  - Creates Project64.lnk directly in the Start Menu, without any folders

None, other than leaving the old shortcuts behind when installing over the old version.

Not compiled at this point, will be done if requested.

Thanks!

### DIFF
--- a/Source/Installer/Installer.iss
+++ b/Source/Installer/Installer.iss
@@ -54,9 +54,7 @@ Name: "{app}\Plugin\GFX\GLideN64"; Permissions: everyone-full
 
 [Icons]
 Name: "{commondesktop}\Project64"; Filename: "{app}\Project64.exe"; Tasks: desktopicon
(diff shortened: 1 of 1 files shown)


DerekTurtleRoe:
I have a branch where I am rebuilding the installer, but until that is done, this seems like a good change. 👍

Fanatic-64:
This has actually been the recommended practice by Microsoft since at least the Windows XP days.

Squall-Leonhart:
leave it alone.

Issue #2461, opened 2025-02-08
[Bug]: Brazilian language pt-br does not appear in the language selector


williansti:
### Actual behavior?

When going to the language option, the Brazilian Portuguese language pt-br is not displayed and cannot be selected, only Portuguese from Portugal, pt-pt, appears in the list.

### Steps to reproduce the problem

1.Open the emulator, any development version and go to the language menu, it will not display the Brazilian Portuguese language.

### Specifications and build details

Project64: (include commit version number and whether you are using the 32-bit or 64-bit version) 

x64 version - DEV-4.0.0.6530-fd062a28 

OS: (include version and OS build number using WinVer): 

Windows 11, 26100.2894

CPU: Core i5 12400

RAM: 32GB 3200mhz

Graphics: GTX 1660 Super

williansti:
![Image](https://github.com/user-attachments/assets/62da8ff2-ed18-4f98-bd3d-85c441791808)

williansti:
I made the solution for the bug in #2467

PR #2467, opened 2025-02-21
Fix bug portuguese brazilian language

williansti:
This fixes the bug that prevents this language from being displayed for selection in the emulator.

This fixes this issue https://github.com/project64/project64/issues/2461


williansti:
Before


With this fix


Working

Squall-Leonhart:
the spacing is not the cause of the issue, these lines are commented so they are never parsed, what was the problem was likely the line encodings, which have been an issue in the distant past for other language files

edit
it was, brazilian is set to utf8 instead of utf8 bom

PR #2469, opened 2025-03-08
NRage plugin enhancements

pcca-matrix:
This PR adds features for analog stick management

capt

### Proposed changes
- sensitivity adjustment
- stick Visualizer
- respect for the N64 octagonal shape
- adjustable stick range and N64 range
- XInput axis inversion support
- Virtual corners in octagon
- Small UI fixes
No

Yes


Squall-Leonhart:
Screenshots of the changes would be great, but i'll build it myself and have a look.

I have a desire to scrap the legacy api's in nrage and implement a common manager built upon sdl, which will select the best api for the control provided and all the other features just sit on the input manager rather then reimplementing things for each, the end result will be like an enhanced version of RMG-Input.

 I may later PR transferpak support into the pj64 core and only expose the emulators support to the plugin.

pcca-matrix:
Screenshot added !
 
Porting the plugin to SDL would definitely make things much simpler, as the two APIs are handled too differently in the code, making it easy to get lost.
I made this small modification for my own needs because i really like how Nrage allows switching between the Mem Pack and Rumble Pack using hotkeys.
 Adding a analog calibration function would also be a great idea, as stick range shapes vary depending on the controllers used.

Per-ROM configuration would also be a great feature. My launcher (currently in dev),  allows loading an INI file based on the launched ROM, but having native support in the plugin would be even better.

E.g. Wave Race is easier to play with a range of 100 for tricks, while Turok controls better with a range of 70.

for the sensitivity adjustment slider , it's more of a curve setting ! it allows shifting the sensitivity to different levels within the range, which is useful but could be improved further.
virtual corners work like in Dolphin, locking the stick into the octagon's corners within a certain range. I found this feature interesting in Dolphin, especially for fighting games.

Squall-Leonhart:
> E.g. Wave Race is easier to play with a range of 100 for tricks, while Turok controls better with a range of 70.

And then theres buck bumble which starts to ramp negatively above a certain range until he's no longer moving.

oddMLan:
@pcca-matrix Are you on Discord? We have some questions about the implementation that would be more convenient to clarify over there (don't want to spam the issue here).
[Invite link to Discord64 - Emulation Haven](https://discord.gg/sSkQTBpFhv)

theboy181:
This is looking great!

I have a couple of suggestions:

Button Mapping Flexibility
It would be really helpful to have the ability to map any button to any XInput control. This level of customization would allow users to fine-tune the control scheme to their liking.

Controller Port Mapping
I'd also like to see an option to assign a single controller to any XInput port (e.g., Controller 1, 2, 3, or 4). This would be especially useful for games like GoldenEye 007, which can benefit from specific controller assignments.

Is this kind of functionality possible?

Squall-Leonhart:
>  It would be really helpful to have the ability to map any button to any XInput control. This level of customization would allow users to fine-tune the control scheme to their liking.

You can already set the n64 controller button to xbox control mapping.

>  I'd also like to see an option to assign a single controller to any XInput port (e.g., Controller 1, 2, 3, or 4). This would be especially useful for games like GoldenEye 007, which can benefit from specific controller assignments.

Not happening, the api determines the order (when its not putting every control on 2+)

PR #2508, opened 2025-12-06
Fix some UI elements not being highlighted/greyed out when settings are changed

Fanatic-64:
Fixes some text labels not being highlighted when their respective setting is changed from its default value. Also fixes a checkbox in the ROM Browser tab not being greyed out when "Use ROM browser" is disabled.

No.

Yes.


Fanatic-64:
Something seems to have gotten screwed up. Lemme try again.

